PR #17, opened 2026-04-15
[hotfix] Add inadvertently dropped climate heuristics inputs

patrickbrown4:
## Summary

The `.gitignore` was a little off, so the `inputs/climate` folder wasn't copied over, which causes the run to fail during input processing. This PR adds it back.

## Charge code for review

GDOP > - > 12495 - 30.01.02

The Pacific test case failed in `copy_files.py` before the change; now it works.


atpham88 (approved):
confirmed working with the USA_defaults case locally and on HPC. thanks Patrick.

do we need to update `upload-artifact` to later version to pass runner tests @kennedy-mindermann?

Issue #22, opened 2026-04-16
Programmatically create hourly profiles for github tests


patrickbrown4:
## Summary

We end up downloading the Zenodo files fairly frequently on the github runner. That stresses the Zenodo servers, slows down the runner, and increases the per-run storage requirements for the runner.

## Proposed solution

Write some functions to programmatically generate fake CF/load profiles, for use only when a `GSw_UseFakeData` (or something) flag is turned on. This approach would only be used for the github tests.

- CF profiles
  - [x] It could read the site class and average CF from `inputs/supply_curve/supplycurve_{tech}-{siting}.csv` and then use a suitable distribution from `np.random` to create a noisy profile with the same mean for each site, matching the structure of the [real site-level files](https://doi.org/10.5281/zenodo.18407659) on Zenodo. (Or they could be completely random.)
  - [x] I think it would go here in `recf.py` as a new if/then: https://github.com/ReEDS-Model/ReEDS/blob/4bf761bcb06ad88fff738405d9f100fcaedd13d2/input_processing/recf.py#L132-L139
- Demand
  - [ ] Same idea in `hourly_load.py`: https://github.com/ReEDS-Model/ReEDS/blob/4bf761bcb06ad88fff738405d9f100fcaedd13d2/input_processing/hourly_load.py#L618-L623
- Infrastructure in `runbatch.py`
  - [x] Add a new test to `check_compatibility()` to enforce that all parallel cases must use the same `GSw_UseFakeData` setting for simplicity
  - [x] Bypass `reeds.remote.download_remote_files()` if `GSw_UseFakeData` is turned on

## Alternative solution

We could instead store a subset of the actual data on Zenodo. But we'd need to update it whenever we update any of the other Zenodo files, or the regions used in the runner tests, which is not ideal. It would also keep us from doing larger-area simplified tests.

(copied from https://github.nrel.gov/ReEDS/ReEDS-2.0/issues/2036)

patrickbrown4:
Addressed by #26

Issue #19, opened 2026-04-16
Clean up logs


patrickbrown4:
## Summary

1. Freshly built environments are printing debug-level log messages from imported packages (matplotlib, fiona for geopandas, maybe others), which makes the terminal log very hard to read. Some examples are below [1].
  - [x] Disable logs from imported packages in `reeds/log.py`. Starting points are [here](https://stackoverflow.com/questions/35325042/python-logging-disable-logging-from-imported-modules) and [here](https://stackoverflow.com/questions/11029717/how-do-i-disable-log-messages-from-the-requests-library).
2. On windows and linux, some users get tons of "Generic family 'sans-serif' not found" messages from matplotlib.
  - [x] The fix on linux/hpc, described [here](https://stackoverflow.com/a/37921026/21987784), is to delete `~/.cache/matplotlib` or `~/.matplotlib`; then the next time you run a script from the `reeds2` environment (as long as you have `mscorefonts` in the environment, which is included by default), it should regenerate the cache file and work. Add that procedure to a new FAQ question on "Help, I'm getting lots of log messages about missing fonts!"
  - [x] It looks like the related file on Windows is at `%HOMEPATH%\.matplotlib` but someone on Windows should try and confirm

[1] Example of log messages from imported packages
```text
2026-04-15 17:41:03,185 - GDAL data found in package: path='/Users/pbrown/anaconda3/envs/reeds2_20251020/lib/python3.11/site-packages/fiona/gdal_data'.
2026-04-15 17:41:03,185 - PROJ data found in package: path='/Users/pbrown/anaconda3/envs/reeds2_20251020/lib/python3.11/site-packages/fiona/proj_data'.
2026-04-15 17:41:03,191 - GDAL data found in package: path='/Users/pbrown/anaconda3/envs/reeds2_20251020/lib/python3.11/site-packages/fiona/gdal_data'.
2026-04-15 17:41:03,191 - PROJ data found in package: path='/Users/pbrown/anaconda3/envs/reeds2_20251020/lib/python3.11/site-packages/fiona/proj_data'.
2026-04-15 17:41:03,191 - Got coordinate system
2026-04-15 17:41:03,191 - Got coordinate system
2026-04-15 17:41:03,191 - Matched. confidence=100, c_code=b'102008', c_name=b'ESRI'
2026-04-15 17:41:03,191 - OLC_FASTSETNEXTBYINDEX: 0
[...]

PR #26, opened 2026-04-17
Create fake hourly profiles for github tests; fix logs

patrickbrown4:
## Summary

This PR addresses #22 and #19:

- A function is added to `recf.py` to create fake CF profiles with the new `GSw_FakeData` switch is active, allowing us to avoid ~12 GB of data download and caching for the runner tests
- The logs had inadvertently been reformatted by bokeh; now they're back to normal

#### Fake profiles

- A new `make_fake_profiles()` function is added to `recf.py`. To make the profiles somewhat realistic in terms of min/mean/max (but not temporal shape), we randomly sample from the two-state [0,1] array, where probability(1) = CF [fraction].
  - These profiles should NOT be used for any kind of analysis; they are just for testing when you want to avoid downloading the actual data via `reeds/remote.py`.
  - Some example histograms are below in the validation section.
- In the current implementation, we only use it for the site-level upv, wind-ons, and wind-ofs profiles.
  - Those are by far the biggest files we download from Zenodo, so avoiding their download speeds up the test and reduces the peak storage needs of the runner
  - It's probably good to keep automatically testing the other Zenodo downloads anyway
    - [ ] Though we might want to extend the new functionality to distPV, since that file is still kind of large at ~440 MB
- In `runbatch.py`, we check to make sure all the cases in a batch use the same `GSw_FakeData` setting (since we download the files shared by all runs)

#### Logs

The verbose logging came from this line: https://github.nrel.gov/ReEDS/ReEDS-2.0/pull/1927/files#diff-a48c953794de025c1a499dbd77153a1da54ebb6bcd68499d4509b32bc4b8477fR21. Because a bokehpivot import was added to `reeds.output_calc`, and because bokeh uses its own debug-level logging, all the python scripts started using debug-level logging. This PR reverts that change, so now the logs are back to normal.

### Additional changes 

- Fixed the git metadata at the top of `meta.csv` (it wasn't working since we don't have any tags on this repo yet)

- `GSw_FakeData` (default 0): Use fake capacity factor profiles (for TESTING ONLY when you do not want to download the real CF data)

### Issues resolved 

- #19
- #22

Zero change for the Pacific test case: [results-v20260416_mainM0_Pacific,v20260416_fakeM0_Pacific.pptx](https://github.com/user-attachments/files/26804587/results-v20260416_mainM0_Pacific.v20260416_fakeM0_Pacific.pptx)

The `github_Pacific` case finished for me locally.

Here are histograms of the CF for a couple random samples:

image

image

### Details to double-check

No


ahamilton5 (approved):
Verbose logs fix works great. Thanks, Patrick.

docs/source/faq.md
```diff
@@ -95,6 +95,16 @@ Here is an example of what this would look like:
 git checkout tags/v2024.0.0
 ```
 
+### Help, I'm getting lots of log messages about missing fonts
+
+We use the `mscorefonts` package to get nicer-looking fonts in plots.
+If you had `matplotlib` installed before running a script from the `reeds2` environment,
+you might need to clear your fonts cache (you can back it up first if you like).
+
+- On Mac/Linux, try deleting `~/.cahce/matplotlib` or `~/.matlotlib`
```
kodiobika:
```suggestion
- On Mac/Linux, try deleting `~/.cache/matplotlib` or `~/.matplotlib`
```

kodiobika (approved):
Thanks for doing this! LGTM

PR #47, opened 2026-04-20
Update PRAS version used in ReEDS to 0.8.0 

sriharisundar:
## Summary
Upgrading PRAS and Julia dependencies, moving Julia from conda environnment to juliaup installation based installation on Linux and MacOS, and updating to Julia 1.12+. With this change PRAS is installed from the Julia general repository and this should reduce installation issues with SSL etc.

## Technical details <!-- if any, otherwise delete -->

### Implementation notes <!-- if any, otherwise delete -->
We restrict Julia to a very specific version because of more flexibility just throwing a warning and not an error - https://github.com/JuliaLang/Pkg.jl/pull/2296
Something to keep an eye out for.

### Additional changes <!-- if any, otherwise delete -->

### Switches added/removed/changed <!-- if any, otherwise delete -->

### Issues resolved <!-- if any, otherwise delete -->
#32 

### Known incompatibilities <!-- if any, otherwise delete -->

### Relevant sources or documentation <!-- if any, otherwise delete -->


## Charge code for review


## Validation, testing, and comparison report(s)
<img width="2328" height="1362" alt="53a60f80-fe3c-43ff-b078-031ccb81b558" src="https://github.com/user-attachments/assets/f510957f-bdcf-40d1-9e89-e56e3fb0ef9e" />

New repo and installation procedures listed work on 
- [x] Kestrel
- [x] Windows (@wesleyjcole)
- [x] MacOS (@kennedy-mindermann can you please test?)
- [x] Yampa (@louisaserpe)
- [x] GH runners

<!--
How did you verify that the changes...
- [ ] have the intended effects?
- [ ] have ONLY the intended effects?
-->

<!--
Include comparison reports (e.g., .pptx files generated by compare_cases.py) for cases commensurate with the level of code/data changes:
- No changes to model or data: Pacific test case only
- Changes to model or data: Full US reference case and full US decarb case
- Changes to model structure, spatial/temporal resolution, or other large changes: All cases in cases_test.csv
-->

<!--
Include additional illustrative plots describing input data, methods, testing, and/or key differences from the comparison report(s)
-->


## Checklist for author
<!-- Fill out before requesting review -->

### Details to double-check
<!-- Delete or ~~strikethrough~~ if not relevant for this PR -->
- [x] Charge code provided for review
- [x] Included comparison reports for appropriate test cases
- [x] Documentation updated if necessary
  <!--
  - model_documentation.md: High-level description of default model behavior
  - user_guide.md: User/developer-facing description of model switches and input files
[...]

PR #36, opened 2026-04-19
Reorganize root of repo

patrickbrown4:
## Summary

Addresses issues [1713](https://github.nrel.gov/ReEDS/ReEDS-2.0/issues/1713) / [2017](https://github.nrel.gov/ReEDS/ReEDS-2.0/issues/2017) (on the old repo) by reorganizing the root of the repo.

**Feedback welcome!**

### Proposed structure

The general idea is to move scripts/functions that are used during the run to subfolders of `reeds/`:

- The GAMS scripts move from the root of the repo to `reeds/core`
- The `input_processing` scripts move to `reeds/input_processing`
- Job-submission scripts go to `reeds/hpc`
- Solver files go to `reeds/solver`
- `ReEDS_Augur` goes to `reeds/resource_adequacy`
  - `reeds2pras` goes to `reeds/resource_adequacy/reeds2pras`
  - Mentions of "Augur" are changed to "resource adequacy calculations"
  - The handoff data between ReEDS and PRAS are now in `{casename}/handoff`
- Auto-generated code (`b_sets.gms`, `report_params.gms`) goes to `{casename}/autocode`
- Root-level scripts mostly go to `helpers`

Here's the high-level view of the proposed structure (generated by running `tree . -I '__pycache__|preprocessing|postprocessing|runs|inputs|hourlize|docs|tests|*.jl' --dirsfirst`; leaves out .jl scripts since the `ReEDS2PRAS` structure is unchanged (the `run_pras.jl` and `instantiate.jl` scripts still exist but are not shown)).

```text
OLD                                  →    NEW
---------------------------------    →    --------------------------------------------
├── input_processing                 →    ├── helpers
│   ├── __init__.py                  →    │   ├── interim_report_batch.py
│   ├── aggregate_regions.py         →    │   ├── interim_report.py
│   ├── calc_financial_inputs.py     →    │   ├── restart_runs.py
│   ├── check_inputs.py              →    │   └── runstatus.py
│   ├── climateprep.py               →    ├── reeds
│   ├── copy_files.py                →    │   ├── core
│   ├── forecast.py                  →    │   │   ├── setup
│   ├── fuelcostprep.py              →    │   │   │   ├── a_createmodel.gms
│   ├── h2_storage.py                →    │   │   │   ├── b_inputs.gms
│   ├── hourly_load.py               →    │   │   │   ├── c_model.gms
│   ├── hourly_plots.py              →    │   │   │   ├── d_mga.gms
[...]

PR #56, opened 2026-04-23
ci: simplify workflow setup and labels

pesap:
## Summary
- add shared `.github/actions/setup-reeds-env` composite action for Python + Julia setup
- refactor `.github/workflows/python-app.yaml` to reuse that action and remove duplicated setup blocks
- remove `.github/workflows/cache.yaml` (cache warming now handled by CI setup job)
- add/update PR automation config:
  - `.github/dependabot.yml`
  - `.github/labeler.yaml`
  - `.github/workflows/labeler.yaml`
- add `.github/workflows/workflow-quality.yaml` with `zizmor` and `actionlint`
- harden workflow tooling by pinning the `actionlint` installer to a commit + SHA256 verification
- normalize dependency label naming to existing `dependencies` label
- why labeler + dependabot:
  - automate PR triage/routing with consistent labels from paths/branch patterns
  - keep `conda`/`pip`/`julia`/`github-actions` dependencies current with low-touch maintenance
  - improve CI reliability and security posture over time
  - use cooldown/PR limits to prevent update noise

## Follow-up fixes in this PR
- switch labeler trigger from `pull_request_target` to `pull_request`
- pin `actions/cache` references to full SHA in CI workflow
- harden Julia PATH export step to avoid template-injection lint warning
- set `persist-credentials: false` in docs checkout step
- remove obsolete conda Julia cleanup path from setup action after #47
- update labeler globs to include root level `*.gms`, remove redundant `createmodel.gms`, trim unused suffix globs, and classify `cases.csv`

## Developer experience impact
- less duplicated CI setup logic and easier maintenance through one shared setup action
- automatic labels on PRs to speed triage and review routing
- automatic dependency update PRs with cooldown and PR limits to keep noise manageable
- workflow lint and security checks catch GitHub action issues earlier

## Watchouts
- label coverage may need tuning as paths evolve, especially after #36
- dependabot updates may still arrive in small bursts during upstream release weeks
- new labels should be monitored for signal versus noise in the first few cycles

## Validation
- `actionlint`
- `uvx zizmor .github`


.github/workflows/python-app.yaml
```diff
@@ -5,11 +5,22 @@ on:
   pull_request:
     branches:
       - main
-    paths-ignore:
-      - 'docs/**'
-      - '*.md'
-      - 'LICENSE'
-      - '.gitignore'
+    paths:
```
kennedy-mindermann: I feel like it might be better to leave this as paths-ignore. What if something changes in the input processing or output processing scripts that causes a run to fail. If that's the only change this wouldn't catch that

pesap: Good catch — agreed. I reverted this back to `paths-ignore` so CI still runs when input/output processing code changes, instead of relying on a brittle allowlist.

pesap: @kennedy-mindermann addressed in commit 046a73f: restored `pull_request.paths-ignore` so CI still runs for input/output processing changes. Resolving this thread now — thanks for catching it.

kennedy-mindermann (approved):

.github/actions/setup-reeds-env/action.yml
```diff
@@ -0,0 +1,70 @@
...
+    required: false
+    default: "1.12.1"
+  instantiate-julia:
+    description: Run julia --project=. instantiate.jl.
+    required: false
+    default: "true"
+  remove-conda-julia:
+    description: Remove julia from micromamba env before setup-julia.
+    required: false
+    default: "true"
```
patrickbrown4: Don't need this part anymore after #47

.github/actions/setup-reeds-env/action.yml
```diff
@@ -0,0 +1,70 @@
...
+        environment-file: environment.yml
+        cache-environment: true
+        post-cleanup: "all"
+        cache-environment-key: ${{ runner.os }}-conda-${{ hashFiles('environment.yml') }}
+
+    - name: Remove conda Julia package
+      if: ${{ fromJSON(inputs.setup-python) && fromJSON(inputs.setup-julia) && fromJSON(inputs.remove-conda-julia) }}
+      shell: bash -le {0}
+      run: micromamba remove -y julia || true
+
```
patrickbrown4: Can remove

.github/workflows/workflow-quality.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref_name }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  zizmor:
```
patrickbrown4, line 21: What do zizmor and actionlint do?

.github/dependabot.yml
patrickbrown4, line 1: What does this workflow do?

.github/labeler.yaml
```diff
@@ -0,0 +1,90 @@
...
+          - "tests/**"
+          - "hourlize/tests/**"
+          - "reeds2pras/test/**"
+          - "**/test_*.py"
+          - "**/*_test.py"
+
+model_changes:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "**/*.gms"
```
patrickbrown4, line 46: If I'm understanding this right, it seems like it would miss the .gms files in the root of the repo. Should it just be "*.gms" (and then you could drop the explicit createmodel.gms)?

.github/labeler.yaml
```diff
@@ -0,0 +1,90 @@
...
+      - any-glob-to-any-file:
+          - "inputs/**"
+          - "hourlize/inputs/**"
+          - "postprocessing/**/inputs/**"
+          - "**/*.csv"
+          - "**/*.h5"
+          - "**/*.hdf5"
+          - "**/*.parquet"
+          - "**/*.xlsx"
+          - "**/*.xls"
```
patrickbrown4: We don't have any of these file suffixes in the repo - ok to remove?

.github/labeler.yaml
patrickbrown4, line 1: So is the idea that all of these labels get applied automatically in github? It's kind of a lot of labels (and we'll need to rework it after #36) but it seems worth trying.

The one other category I can think of that might be useful would be changes to `cases.csv`, as they affect the user interface / default assumptions.

patrickbrown4 (commented):
Thanks Pedro. I'm not familiar with a lot of the details, so will defer to you and Kennedy for signoff.

Would you mind adding to the PR text a short description of how this PR will change the developer experience? Are there new things we need to watch out for?

.github/actions/setup-reeds-env/action.yml
pesap: Cool was not aware of that!

.github/workflows/workflow-quality.yaml
pesap: [zizmor](https://docs.zizmor.sh) is the cybersecurity audit for each action to avoid any risks (obvious ones). [actionlint](https://github.com/rhysd/actionlint)  is just a static GitHub action linter (like ruff).

.github/dependabot.yml
pesap: Dependabot helps keep our dependencies up to date by automatically suggesting bumps for any hot-fixes by dependencies.

.github/actions/setup-reeds-env/action.yml
pesap: Agreed. I’ll remove this step as well (same cleanup from #47 context).

.github/labeler.yaml
pesap: Yes, the intent is automatic path based labels for faster triage. Great suggestion on `cases.csv`. I will add it so UI and default assumption changes are surfaced. We can tune label granularity again after 36.

.github/labeler.yaml
pesap: Great catch. You are right. I will update this to include root level `*.gms` coverage and remove the redundant `createmodel.gms` entry.

.github/labeler.yaml
pesap: Agreed. I checked the repository and we do not currently have those file types, so I will remove those globs to keep label matching tighter.

pesap:
@patrickbrown4 requested your re review after latest updates. Addressed your feedback on setup action cleanup, labeler rules, and added dev experience plus watchouts to the PR description.

.github/labeler.yaml
```diff
@@ -0,0 +1,88 @@
...
+      - any-glob-to-any-file: "postprocessing/**"
+
+reeds2pras:
+  - changed-files:
+      - any-glob-to-any-file: "reeds2pras/**"
+
+inputs:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "inputs/**"
```
patrickbrown4: `"inputs/**"` shows up twice; here and in `data_changes`. Is that intended?

patrickbrown4 (approved):
Thanks! One more minor comment in labeler.yaml but otherwise looks good.

.github/labeler.yaml
pesap: I think so. They are in two different labels. Shall we just capture it on data_changes?

patrickbrown4: Yeah I think just including it in `data_changes` is enough. Then this one could be either `defaults` or `switches` since it only contains `cases.csv`.

pesap: Sounds good. it is included here: https://github.com/ReEDS-Model/ReEDS/pull/56/commits/7c1963178bd447600cba506a6438da65f1e595a0

Issue #41, opened 2026-04-20
Fix Monte Carlo sampling for demand and VRE availability


patrickbrown4:
## Summary

Monte Carlo sampling for electricity demand and VRE availability started failing after a recent PR; this issue is a reminder to fix them.

## TODO

- [ ] Demand
  - [ ] Only allow two spatial options for load (`weight_r`): `country` or `st` (no sampling at BA or other resolutions)
- [ ] VRE availability
  - [ ] Wait to address this functionality until #16 is addressed, since that will change the way the supply curves and CF profiles are processed.
  - [ ] One option that would simplify things would be to only use a single profile (regardless of siting access scenario) for each `sc_point_gid`. Then we wouldn't have to redetermine the resource classes or site CF profiles when using Monte Carlo sampling (and would also cut down the amount of data we'd need to download from Zenodo by ~2/3).
    - It shouldn't make a meaningful difference for PV, but the site-level CF for wind is noticeably higher for some sites with limited siting than open-access siting, because the sparser turbine arrangement reduces wake losses. So, assuming we'd use the open-access CF everywhere, we'd need to decide if we're ok with underestimating the wind CF in the limited-access scenario.
      - <img width="1900" height="609" alt="Image" src="https://github.com/user-attachments/assets/a08c5f54-e5f6-479f-9c7c-b6473f2f8654" />
    - [ ] Some of these sites might have very little capacity; it would help to look at the difference in annual energy production by site instead of CF
    - [ ]  Consider using reference-access by default and only fill in with open-access if reference-access is missing a profile for the site

(moved from https://github.nrel.gov/ReEDS/ReEDS-2.0/issues/1981)

patrickbrown4:
Now that #12 has been merged, the rest of this issue should be addressable.

PR #12, opened 2026-04-15
Adding technology classes

atpham88:
## Summary
This PR:
- Removes `startyear` switch and instead establishes `startyear` based on `yearset`
- Adds algorithm to assign pv, wind, and geothermal tech classes to NEMS units
- Fixes a few issues that cause missing existing capacities in main
- Adds an option to specify a flexible start year, instead of 2010
   - Although this option is added in this PR, enabling and testing for start year later than 2010 is in a separate PR.

## Technical details 
These changes are made in this PR:
- `process_unitdata.py`: Read `unitdata.csv` from `inputs_case` and process it to assign sc_point_pids, capacity factors/mean temps to pv, wind, and geothermal units. This file is run immediately after `copy_files.py`.  This allows us to remove all the mapping of NEMS units to `sc_point_gids` in the [ReEDS_input_processing repo](https://github.com/ReEDS-Model/ReEDS_Input_Processing).
- `WriteHintage.py`: Allow start year to be flexible instead of 2010.
- `writecapdat.py`: Add mappings to solar, wind, geothermal resource class are based on the resource quality (capacity factors for wind and solar and temperature for geothermal) of the technology as it comes from reV.
- `c_supplymodel.gms`
   - `pcat` set is removed from the model: The equations `eq_force_prescription_power` and `eq_force_prescription_energy`, along with the variables `EXTRA_PRESCRIP` and `EXTRA_PRESCRIP_ENERGY`, are currently indexed by `pcat`. Since the prescribed generator classes and vintages are now defined in `writecap.py`, the `pcat` index is no longer needed in these equations or variables. Accordingly, the equations are revised so that the investment (`INV`) of prescribed generators—indexed by (i, newv, r, t)—is equal to the sum of the prescribed build values and the additional prescribed investments (`EXTRA_PRESCRIP`) for those same (i, newv, r, t) combinations.
  - Integrate prescribed retirements when the prescribed facilities retires before their life time: In the NEMS database, the retirement year of some facilities occurs before the end of their expected lifetimes. To accurately capture these early retirements for prescribed facilities, the prescribed retirement values are subtracted from total investments when calculating facility capacities in `eq_cap_new_noret` and `eq_cap_energy_new_noret`. Although this implementation follows the same logic as `m_capacity_exog`, it explicitly captures prescribed generator builds in the INV variables in the current version of the model.
- `hourlize`: 
  - Remove scripts that generate `exog_cap_{tech}.csv` and those are associated with `existing_sites` files as they are no longer needed.
[...]

PR #57, opened 2026-04-23
Add Latin Hypercube Sampling for Monte Carlo Analysis

bsergi:
## Summary

This PR adds the option to sample using a Latin Hypercube sampling (LHS) method when running a Monte Carlo analysis. It also fixes the sampling approach for load. 

## Technical details <!-- if any, otherwise delete -->

LHS is a quasi-random sampling method that partitions each input distribution being sample into N bins of equal probability and then draws one sample from each bin. The advantage of this method is that it enables convergence to the "true" input distribution with a lower number of samples. The figure below illustrates this with a comparison of random and LHS methods for uniform and triangular costs using the 2050 nuclear ATB costs: 

<img width="486" height="291" alt="image" src="https://github.com/user-attachments/assets/fd926eca-d808-4c39-be4f-6b22ab99ee69" />

More details on the LHS approach can be found in [Sheikholeslami, R., & Razavi, S. (2017)](https://doi.org/10.1016/j.envsoft.2017.03.010 ).

### Implementation notes <!-- if any, otherwise delete -->

The LHS method is implemented in the `mcs_sampler.py` during input processing. When activated, the LHS method samples an n x d matrix where n=number of samples (specified by `MCS_runs`) and d=dimensions, or the number of variables being sample (specified by `MCS_dist_groups`). 

This matrix is generated upfront and provides the place on the cumulative distribution function for each sample. These are subsequently used by a set of lhs functions to derive the realized values from the respective distributions. The original LHS sample matrix is saved to `inputs_case/mcs_latin_hypercube_samples`. Note that this approach is distinct from the implementation of the pure random approach, which samples weights and the applies them to the relevant files and switches.

### Additional changes <!-- if any, otherwise delete -->

### Switches added/removed/changed <!-- if any, otherwise delete -->

Adds `MCS_lhs`: 0 to use random sampling, 1 to use LHS (default: 1)
Modifies `input_processing_only`: adds an option 2 that stops input processing right after Monte Carlo sampling (useful for testing the input distributions before running).

### Issues resolved <!-- if any, otherwise delete -->

Partially addresses #41   by fixing load.

### Known incompatibilities <!-- if any, otherwise delete -->

- The LHS method does not currently support regional sampling.
- Sampling with VRE availability is still not functioning. The plan is to address this after the changes in #12 .

### Relevant sources or documentation <!-- if any, otherwise delete -->

Monte Carlo sampling is off by default so there is no change to the default case (see compare below). 

[results-main,update.pptx](https://github.com/user-attachments/files/28154418/results-main.update.pptx)

The next slide deck summarizes input distribution and results from a set of 54-region ReEDS runs using both the random and LHS approaches for a different number of samples. In general the LHS converges faster to the expected input distributions than the random approach. Both approaches yield reasonably comparable results on aggregate metrics such as mean and 90% coverage for installed capacity, annual generation, and system costs.

[20260430_latin_hypercube_sampling.pdf](https://github.com/user-attachments/files/27374268/20260430_latin_hypercube_sampling.pdf)

<!--
How did you verify that the changes...
- [ ] have the intended effects?
- [ ] have ONLY the intended effects?
-->

<!--
Include comparison reports (e.g., .pptx files generated by compare_cases.py) for cases commensurate with the level of code/data changes:
- No changes to model or data: Pacific test case only
- Changes to model or data: Full US reference case and full US decarb case
- Changes to model structure, spatial/temporal resolution, or other large changes: All cases in cases_test.csv
-->

<!--
Include additional illustrative plots describing input data, methods, testing, and/or key differences from the comparison report(s)
-->


## Checklist for author
<!-- Fill out before requesting review -->

### Details to double-check
<!-- Delete or ~~strikethrough~~ if not relevant for this PR -->
  <!--
  - model_documentation.md: High-level description of default model behavior
  - user_guide.md: User/developer-facing description of model switches and input files
  - faq.md: Limitations, caveats, and known issues
  - postprocessing_tools.md: User/developer-facing description of scripts in postprocessing folder
  - README.md files: User/developer facing description of individual input folders
  -->
  - [ ] Units are specified
- [x] Code formatting standardized <!-- Coding conventions: https://reeds-model.github.io/ReEDS/developer_best_practices.html#coding-standards-and-conventions -->

<!-- These do not need to be checked to merge, but if unchecked, a deeper level of review may be required -->

I used Claude to generate the function docstrings

<!-- Model dimensions -->
<!-- - [ ] County resolution: @lserpe -->
<!-- - [ ] Demand data: @ahamilto -->
<!-- - [ ] Emissions: @apham -->
<!-- - [ ] Financial calculations: @wcole -->
<!-- - [ ] FINITO: @molmezt or @cavraam -->
<!-- - [ ] Hybrids: @aschleif -->
<!-- - [ ] Monte Carlo: @bsergi -->
<!-- - [ ] Sparse chronology or interday diurnal storage: @ychen10 -->
<!-- - [ ] State policies: @wcole or @aschleif -->
<!-- - [ ] Stress periods, resource adequacy, or ReEDS2PRAS/Julia: @pbrown -->
<!-- - [ ] Supply curves, hourlize, reeds_to_rev: @bsergi -->
<!-- - [ ] Time resolution: @pbrown -->
<!-- - [ ] Water cooling: @jcarag or @scohen -->

<!-- Pre/Postprocessing -->
<!-- - [ ] Dispatch mode (run_pcm.py): @pbrown -->
<!-- - [ ] Github actions: @kminderm -->
<!-- - [ ] Health impacts: @bsergi -->
<!-- - [ ] Input data structure and processing: @cobika -->
<!-- - [ ] Interactive plots (bokeh, vizit): @mmowers -->
<!-- - [ ] Land use: @bsergi -->
<!-- - [ ] R2X: @psanchez -->
<!-- - [ ] Retail rates: @wcole -->
<!-- - [ ] Static plots (single_case_plots.py, compare_cases.py): @pbrown -->
<!-- - [ ] Tableau: @ahamilto -->
<!-- - [ ] Technology value (reValue, valuestreams.py): @mmowers -->

<!-- Technologies -->
<!-- - [ ] Batteries: @wcole -->
<!-- - [ ] EV managed charging (EVMC): @mvanatta -->
<!-- - [ ] Fossil, CCS, or DAC: @mbrown1 -->
<!-- - [ ] Geothermal: @ssharma2 -->
<!-- - [ ] Hydropower or PSH: @scohen -->
<!-- - [ ] Hydrogen: @bsergi -->
<!-- - [ ] Nuclear: @wcole -->
<!-- - [ ] Solar: @wcole -->
<!-- - [ ] Transmission: @pbrown -->
<!-- - [ ] Wind: @mmowers -->


docs/source/user_guide.md
```diff
@@ -567,6 +589,7 @@ These parameters can be adjusted to emphasize or de-emphasize specific scenarios
 Since these weights are used as multiplicative factors on reference files,
 Dirichlet sampling tends to favor combinations that lean toward central scenarios when the inputs represent "low",
 "medium", and "high" pathways.
+-
```
patrickbrown4:
```suggestion
```

input_processing/mcs_sampler.py
```diff
@@ -16,6 +18,7 @@
 import datetime
 from typing import Tuple, List
 from collections import defaultdict
+from scipy.stats import qmc, uniform, triang
```
patrickbrown4: (nitpick / optional) For imported functions I usually do `import scipy.stats` and then call these as `scipy.stats.qmc`, just because I don't have these interfaces memorized and it avoids needing to go look up where `qmc`, `uniform`, etc. came from when reading the code

input_processing/mcs_sampler.py
```diff
@@ -226,6 +241,52 @@ def get_hierarchy_file(inputs_case: str, ReEDS_resolution: str) -> pd.DataFrame:
 
     return filtered_hierarchy
 
+def check_lhs_settings(dist_params, sample_group):
+    """Check whether a dirichlet distribution can be mapped to a supported LHS distribution.
+
+    Dirichlet distributions with identical parameters of length 2 or 3 are
+    equivalent to uniform or triangular distributions, respectively, which
```
patrickbrown4: Could you say more about how the usage with 3 parameters relates to the Dirichlet distribution? The "peak" of the 3-equal-parameter Dirichlet could range between a mix of low/high and pure mid, and since the high/mid and mid/low ratios change over time, they're not exactly equivalent. Does LHS use a pure triangular distribution or is it more like Dirichlet?

runbatch.py
```diff
@@ -531,6 +531,14 @@ def check_compatibility(sw):
             )
             raise ModuleNotFoundError(err)
 
+# function to stop the model after input processing
+def stop_after_input_processing(OPATH, reeds_path, casedir, caseSwitches):
+    comment(f'Exit after input_processing', OPATH)
```
patrickbrown4: (nitpick) ruff complains about unused f-strings
```suggestion
    comment('Exit after input_processing', OPATH)
```

cases.csv
patrickbrown4, line 1: Could you add demand back to the `MonteCarlo` case in `cases_test.csv` so we can keep an eye on it? I think it would just be adding `.load_country` to its `MCS_dist_groups` setting.

input_processing/mcs_sampler.py
```diff
@@ -736,56 +844,65 @@ def _get_weights_general(
...
-        #Examples: p1, p2, New_England, ...(covers NG and LOAD)
+        #Examples: p1, p2, New_England, ME ...(covers NG and LOAD)
         columns_in_hierarchy = [col for col in dist_files[0].keys() if col in set(self.r_weights.keys())]
 
+        # load files are subsetted to the relevant regions later in hourly_load.py, 
+        # so here we add placeholder weights for any states not being modeled
+        if file_name == 'load.h5':
+            columns_other_states = [col for col in dist_files[0].keys() if col not in columns_in_hierarchy]
+            if len(columns_other_states) > 0:
+                generic_weight_matrix = self.r_weights[next(iter(self.r_weights))]
```
patrickbrown4: (nitpick) I don't see `next(iter())` often - it looks like it grabs the first element of `r_weights`, is that right? If you don't need the values, maybe it'd be clearer to just use `np.ones()` instead?

input_processing/mcs_sampler.py
```diff
@@ -736,56 +844,65 @@ def _get_weights_general(
...
-        # (sample, file) -> pd.DataFrame
+        # file -> pd.DataFrame
         dict_df_weights = {}
 
         # Handle the simple case where there is only one weight for all regions 
         # (or no regions). 
         if single_r_weight:
             # Get the first region key
-            first_region = next(iter(self.r_weights)) 
+            first_region = next(iter(self.r_weights))
```
patrickbrown4: (nitpick) It might be marginally more self-explanatory to just do `self.r_weights[0]` (or `list(self.r_weights)[0]` if it's a dictionary) instead of `next(iter())` (which I originally thought was doing some complicated generator stuff) but up to you (not necessary to change now since it's also used elsewhere in this script)

reeds/input_processing/mcs_sampler.py
patrickbrown4, line 1: I tried running the `MonteCarlo` case from `cases_test.csv` and got this error:

```text
-----------------------
starting mcs_sampler.py
-----------------------
mcs_sampler.py | 2026-05-11 15:48:03 | INFO | Starting mcs_sampler.py
mcs_sampler.py | 2026-05-11 15:48:03 | ERROR | Traceback (most recent call last):
mcs_sampler.py | 2026-05-11 15:48:03 | ERROR |   File "/Users/pbrown/github/ReEDS/runs/v20260611_lhsM0_MonteCarlo/input_processing/mcs_sampler.py", line 1973, in <module>
mcs_sampler.py | 2026-05-11 15:48:03 | ERROR | main(reeds_path, inputs_case, n_samples=MCS_runs, lhs_sampling=MCS_lhs)
mcs_sampler.py | 2026-05-11 15:48:03 | ERROR |   File "/Users/pbrown/github/ReEDS/runs/v20260611_lhsM0_MonteCarlo/input_processing/mcs_sampler.py", line 1889, in main
mcs_sampler.py | 2026-05-11 15:48:03 | ERROR | mcs_run_number = int((runs_folder_name.split('_')[-1]).replace('MC', ''))
mcs_sampler.py | 2026-05-11 15:48:03 | ERROR | ValueError
mcs_sampler.py | 2026-05-11 15:48:03 | ERROR | :
mcs_sampler.py | 2026-05-11 15:48:03 | ERROR | invalid literal for int() with base 10: 'MonteCarlo'
v20260611_lhsM0_MonteCarlo has finished 
```

reeds/input_processing/mcs_sampler.py
```diff
@@ -999,11 +1115,21 @@ class MCS_Sampler:
...
 
+        # sampling method (random or latin hypercube)
+        if lhs_samples is None:
+            self.sampling_method = 'random'
+            self.sample_num = mcs_run_number
+        else:
+            self.sampling_method = 'latin hypercube'
+            self.lhs_samples = lhs_samples
+            # for each ReEDS run the sample id corresponds to the run number (-1 for 0 index)
+            self.sample_num = mcs_run_number - 1
```
patrickbrown4, line 1141: Do you think it would help to do the same thing for `sampling_method = 'random'`, just so there's one fewer point of divergence between the two? Either way is ok with me

input_processing/mcs_sampler.py
```diff
@@ -1339,121 +1456,313 @@ def _apply_weights_switches_csv(
...
-        save_path = os.path.join(log_folder, 'mcs_group_weights.csv')
+        save_path = os.path.join(inputs_case, 'mcs_group_weights.csv')
         r_weights = self.weight_calc.r_weights
         group_name = self.sample_group["name"]
         assignments_list = self.sample_group["assignments_list"]
 
         # Get shape of any region’s weight array
         any_region = next(iter(r_weights))
-        n_samples, n_assignments = r_weights[any_region].shape
+        n_assignments = r_weights[any_region].shape
```
patrickbrown4: unused

input_processing/mcs_sampler.py
```diff
@@ -1339,121 +1456,313 @@ def _apply_weights_switches_csv(
...
+        and stores the result as a string in self.samples.
+
+        Args:
+            sample_group_num (int): Index of the sample group (LHS dimension).
+            sample_idx (int): Index of the Sample_ID in sample_group.
+            dist_files (List[pd.DataFrame]): Reference DataFrames for the switch.
+            n_decimals (int): Number of decimal places for rounding the result.
+        """
+        sw_name = self.sample_group['switch_names'][sample_idx]
+        samples_sw = [None for n in range(self.n_samples)]
```
patrickbrown4: unused

input_processing/mcs_sampler.py
```diff
@@ -1496,57 +1813,53 @@ def write_samples(
...
 
     inputs_case = sample_group['inputs_case']
-
     for sample_idx, sample_ID in enumerate(samples_dict.keys()):
-        samples = samples_dict[sample_ID]
+        sample_values = samples_dict[sample_ID]
         sw_name = sample_group['switch_names'][sample_idx]
-        save_path_structure = sample_group['save_paths'][sample_idx]  # Where the samples will be copied to
+        save_path = sample_group['save_paths'][sample_idx]  # Where the samples will be copied to
+        folder_path = os.path.dirname(save_path)  # Folder path without the file name
```
patrickbrown4: unused

input_processing/mcs_sampler.py
```diff
@@ -1339,121 +1456,313 @@ def _apply_weights_switches_csv(
...
+        # iterate over columns to update with lhs 
+        for mod_col in modifiable_columns:
+            # get distribution parameters and sample
+            if self.distribution == "dirichlet":
+                # special case: dirichlet with two or three of the same parameters are 
+                # equivalent to uniform and triangular, respectively
+                self.distribution = check_lhs_settings(self.dist_params, self.sample_group)        
+
+            # note that this is deliberately not an elif as the above block adjusts select dirichlet
+            # distributions to be uniform or triangular as appropriate
```
patrickbrown4: I would generally lean toward raising an error (rather than coercing the values) if incompatible options are selected; I think it minimizes surprises and ensures that the input is always "right" no matter where you read it from. (Either that or you build the value-coercion in way upstream, in a single function that is always used to read the inputs, but it's still hard to enforce that all users use that function, and I worry that someone will look at the config and conclude "LHS still uses Dirichlet".)

I think it'd be fine to add more entries to `mcs_distributions_default.yaml` to enable the use of the explicit "uniform" input (I started doing something similar on [pb/mcplots_merge](https://github.com/ReEDS-Model/ReEDS/tree/pb/mcplots_merge) but haven't gotten around to cleaning it up yet)

reeds/input_processing/mcs_sampler.py
```diff
@@ -1566,78 +1879,100 @@ def main(
...
-        main(reeds_path, inputs_case, n_samples=1)
+    if not nolog:
+        log = reeds.log.makelog(
+            scriptname=__file__,
+            logpath=os.path.join(os.path.dirname(inputs_case), 'gamslog.txt'),
+        )
+    
+    # use sampling capability to set up random vector for modeing to generate alternatives (MGA)
+    if random_vector:
+        print("MGA random vector sampling will be implemented in the future")
```
patrickbrown4: I'd remove this for now since it's not implemented yet

runbatch.py
```diff
@@ -1393,6 +1401,12 @@ def write_batch_script(
         restartfile = batch_case
         OPATH.writelines(writeerrorcheck(os.path.join('g00files', restartfile + '.g*')))
 
+        # call to set up MGA random vector sampling if needed
+        if(float(caseSwitches['GSw_MGA_CostDelta']) > 0):
+            OPATH.writelines(
+                    f"python {os.path.join(casedir,'input_processing','mcs_sampler.py')} {reeds_path} {inputs_case} -r \n")
+        OPATH.writelines(writescripterrorcheck(s)+'\n')
```
patrickbrown4: Would you be ok removing this for now to avoid confusion? I'd save all the random-vector stuff for a dedicated PR (and it sounds like we should be able to implement it before `createmodel.gms` anyway).

reeds/input_processing/mcs_sampler.py
```diff
@@ -1566,78 +1879,100 @@ def main(
...
+        lhs_samples_out = pd.DataFrame({'run': [f"MC{i:0>4}" for i in range(1, n_samples + 1)]})
+        lhs_samples_out = pd.concat(
+            [lhs_samples_out, pd.DataFrame(lhs_samples.round(6), columns=df_input_dist_instructions.name.values)], axis=1
+        )
+        lhs_samples_out.to_csv(os.path.join(inputs_case, "mcs_latin_hypercube_samples.csv"), index=False)
+
+    # if using random sampling, set random seed using seed + MCS run number 
+    # to allow reproducibility without having the same sample for each MCS-ReEDS call
+    else:
+        np.random.seed(seed + mcs_run_number)
```
patrickbrown4, line 1910: Since there's still randomness in LHS, I think it might be best to still run `np.random.seed(seed)` for reproducibility when using LHS.

You could use the approach taken for the `pras_seed` switch, where if the switch is set to 0 we do not set the seed, and if it's a nonzero integer we set it to the integer; that way a user could still turn off the seed if desired. Personally I would set a nonzero default for reproducibility, but either works as long as it's user accessible.

cases.csv
patrickbrown4: And what do you think about adding a LHS test to `cases_test.csv` too? I could see it being used more often than some of the other capabilities we test for, and since the Monte Carlo capability is pretty sensitive to input file structure (which changes relatively often) it could be good to keep it in the test suite.

reeds/input_processing/mcs_sampler.py
patrickbrown4: Incidentally the explicit seed would also make it easier to extend a batch of random Monte Carlo runs; like if you do 300 and decide you want 200 more, you could just set the seed to 300 to get the "next" 200 samples

input_processing/mcs_sampler.py
patrickbrown4: Discussed; adding some spaghetti plots of the inputs could help explain the behavior

patrickbrown4 (commented):
Very cool! I think we should keep the old MC approach working (currently the `MonteCarlo` case in `cases_test.csv` is failing for me on this branch) but otherwise most of my comments are just stylistic. Happy to approve once the test is passing again.

reeds/input_processing/mcs_sampler.py
bsergi: This case is working for me locally. Something seems off with your run name as [this block](https://github.com/ReEDS-Model/ReEDS/blob/bdb3182b7d93186df0159e40717540b386466168/reeds/inputs.py#L317) should be leading to 4 digit padding (e.g., `MC0001`) but you are somehow ending up with just `M0`. Can you confirm you're still seeing this on the latest for this branch? If so maybe we can grab a minute to trace this down.

patrickbrown4: Ah I see, when I run it with `caffeinate -i python runbatch.py -b v20260512_lhsM0 -c test -s MonteCarlo`, it fails with the error above. But when I run it by changing the "ignore" column to zero, it passes. That same behavior also happens on main. So just user error, sorry! We can fix the `runbatch.py --single` behavior elsewhere.

patrickbrown4 (approved):
Going ahead and approving since the rest of the comments are mostly stylistic (though I do think it'd be good to add a LHS scenario to `cases_test.csv` and to be able to fix the LHS seed)

docs/source/user_guide.md
```diff
@@ -448,9 +448,11 @@ For example, `default` will use `inputs/userinput/mcs_distributions_default.yaml
 3. Set `MCS_dist_groups` to one or more YAML group names. Separate multiple groups with a dot.
    For example `tech.load_st.ng_fuel_price`.
 
-4. Run ReEDS as usual. Each Monte Carlo draw will create its own run using the sampled inputs.
+4. Choose the sampling method; `MCS_lhs=1` uses a Latin Hypercube sampling method and `MCS_lhs=0` uses random sampling.
 
-These three switches (`MCS_runs`,`MCS_dist`, and  `MCS_dist_groups`) are the only required controls.
+5. Run ReEDS as usual. Each Monte Carlo draw will create its own run using the sampled inputs.
+
+These four switches (`MCS_runs`,`MCS_dist`, `MCS_dist_groups`, and `MCS_lhs`) are the only required controls.
```
kodiobika:
```suggestion
These four switches (`MCS_runs`, `MCS_dist`, `MCS_dist_groups`, and `MCS_lhs`) are the only required controls.
```

docs/source/user_guide.md
```diff
@@ -555,6 +557,26 @@ Each state receives its own weighted combination of the two load scenarios.\
 
 This enables state level uncertainty in siting supply curves for wind and solar technologies through a random draw between `limited` and `reference` conditions.
 
+### Sampling method
+
+There are two sampling approaches available for Monte Carlo analysis: random sampling and Latin Hypercube sampling. 
+
+Random sampling uses the numpy `random` method for the relevant distribtion to sample a set of weights.
```
kodiobika:
```suggestion
Random sampling uses the numpy `random` method for the relevant distribution to sample a set of weights. 
```

docs/source/user_guide.md
```diff
@@ -555,6 +557,26 @@ Each state receives its own weighted combination of the two load scenarios.\
...
+
+The second approach, Latin Hypercube sampling (LHS), utilizes a quasi-Monte Carlo sampling method that is 
+designed to improve efficiency by reducing overlap of the sampled values. An overview of this method
+can be found in {cite}`sheikholeslamiProgressiveLatinHypercube2017`. For this method, an NxD matrix is generated upfront
+for all model runs based on the number of samples (N) and the independent dimensions being sampled (D). 
+The values in this matrix represent sampling of the cumulative probability distribution functions, 
+and are later used by the inverse CDF (percent point) functions to derive the actual sample values. 
+A single seed value is used for all runs, resulting in unique sampling matrices for a given set of values of N and D.
+
+The LHS method tends to results in sampling value that converge on the true input distributions for lower numbers of samples.
```
kodiobika:
```suggestion
The LHS method tends to result in sampling values that converge on the true input distributions for lower numbers of samples.
```

input_processing/mcs_sampler.py
```diff
@@ -330,84 +390,134 @@ def general_mcs_dist_validation(reeds_path: str, mcs_dist_path: str, sw: pd.Seri
...
         distribution = sample_group['dist']
-
         switch_names = [next(iter(s)) for s in sample_group["assignments_list"]]
         sw_assignments = [next(iter(s.values())) for s in sample_group["assignments_list"]]
+    
+        ## check that distribution specified is valid
+        if distribution not in MCSConstants.VALID_DISTRIBUTIONS:
+            raise ValueError(
+                f"The distribution {distribution} is not supported."
+                f"Please choose one of the following: {MCSConstants.VALID_DISTRIBUTIONS}")
```
kodiobika: It's probably better to check this upfront (before the for loop). In general I'd say it's ideal to check as many things upfront as possible to avoid going through the loop and then finding issues at the end, but I know some of the checks become more complicated outside of the for loop so no need to change any/everything
```suggestion
        ## check that all the distributions specified are valid
        invalid_distributions = [
            dist for dist in df_input_dist['dist'].unique()
            if dist not in MCSConstants.VALID_DISTRIBUTIONS
        ]
        if len(invalid_distributions) > 0:
            raise ValueError(
                f"The following distributions are not supported: {invalid_distributions}."
                f"Please choose one of the following: {MCSConstants.VALID_DISTRIBUTIONS}")
```

input_processing/mcs_sampler.py
```diff
@@ -575,11 +680,16 @@ def get_all_region_weights(
...
 
-        # Only save the cendiv weights if the sample_hierarchy_lvl is 'country' or 'cendiv'
+        # Natural gas fuel prices are at the census division level, so save the cendiv weights 
+        # if the sample_hierarchy_lvl is 'country' or 'cendiv'
         if sample_hierarchy_lvl in ["country", "cendiv"]:
             cendivs = hierarchy_file.loc[hierarchy_file[sample_hierarchy_lvl] == region, "cendiv"].unique()
             for cendiv in cendivs:
                 all_r_weights[cendiv] = r_weights
+        # Load data is at the state level, so save the state weights if
+        # the sample_hiearchy_lvl is 'st'
```
kodiobika:
```suggestion
        # the sample_hierarchy_lvl is 'st'
```

input_processing/mcs_sampler.py
```diff
@@ -1308,23 +1429,19 @@ def _apply_weights_recf(
...
+        for f, df in enumerate(dist_files):
+            # Get the old and new class|region combinations from weights[(s, f)]
+            for (new_c_r, old_c_r) in weights[f].index:
+                sample_sw[new_c_r] += df[old_c_r] * weights[f].loc[(new_c_r,old_c_r)].values[0]
 
-            for f, df in enumerate(dist_files):
-                # Get the old and new class|region combinations from weights[(s, f)]
-                for (new_c_r, old_c_r) in weights[(s, f)].index:
-                    sample_sw[new_c_r] += df[old_c_r] * weights[(s, f)].loc[(new_c_r,old_c_r)].values[0]
+        # Round numbers to 9 decimal places and allow a maxium values of 1
```
kodiobika:
```suggestion
        # Round numbers to 9 decimal places and allow min/max values of 0 and 1
```

input_processing/mcs_sampler.py
```diff
@@ -1467,15 +1776,24 @@ def get_samples(self, aux_files):
...
-                self._apply_weights_switches_csv(dist_files, n_decimals, dict_df_weights, sample_idx)
-            elif file_name in MCSConstants.RECF_FILES:
-                self._apply_weights_recf(dist_files, sample_idx)
+            if self.sampling_method == 'latin hypercube':
+                # apply latin hybercube sampling based on file type
+                if file_name == "switches.csv":
+                    self.apply_lhs_switches_csv(sample_group_num, sample_idx, dist_files, n_decimals)
+                elif file_name in MCSConstants.RECF_FILES:
+                    #self._apply_weights_recf(dist_files, sample_idx)
+                    pass
```
kodiobika: Should this case raise an error?

kodiobika (approved):
Very cool! LGTM

input_processing/mcs_sampler.py
bsergi: Right now we flag that the sampling with siting is disabled upstream, so I'm going to remove the `pass` and revert the comment.

cases.csv
bsergi: Good call on adding load back. I also like the idea of a separate LHS test so will add that in as well.

input_processing/mcs_sampler.py
bsergi: I thought about it a bit more--since a dirichlet(1,1,1) and a triangular are similar but not exactly the same I think it makes sense to remove this and just enforce using the corresponding distributions. Will add more options to the default yaml file and checks in the validation step up front.

input_processing/mcs_sampler.py
bsergi: I moved a few of the checks up out of the loop. Since the distributions I added complicate some of the rule checking I moved some of the rules in a new file (`mcs_distribution_rules.yaml`). There's still a loop at the end checking some things that I wasn't sure about change but those could probably be vectorized at some point too.

input_processing/mcs_sampler.py
bsergi: Attaching a few spaghetti plots comparing dirichlet(1,1,1) and triangular. For the former you notice some different angles, but I actually didn't observe any samples that crossed-over each other. I think that probably has to do with the low likelihood that you sample twice in a two places that are close enough to that they could cross, but I'm not certain.

Regardless, after some reflection I think it's probably best to just enforce using the triangular, uniform, or discrete with the latin hypercube and not try to translate a dirichlet over, so I've updated that accordingly.

[20260514_lhs_distribution_comparison.pptx](https://github.com/user-attachments/files/27971525/20260514_lhs_distribution_comparison.pptx)

reeds/input_processing/mcs_sampler.py
bsergi: Ah ok that's interesting. I took a pass at fixing for Monte Carlo runs in [this commit](https://github.com/ReEDS-Model/ReEDS/commit/eaad50cc12b5ff6443ada71e58717e99f938c9ca) but I've never used this feature before so might be good for a second set of eyes there.

runbatch.py
bsergi: Yea makes sense, seems like we're going with a different approach here anyway.

input_processing/mcs_sampler.py
bsergi: Updated

input_processing/mcs_sampler.py
bsergi: This one is easy enough to change so adjusted. I wasn't familiar with the next(iter()) syntax but I will say it is growing on me!

bsergi (commented):
Made some changes to address the comments. I still need to update with main but will aim to do that later this week. In the meantime let me know if you have additional comments, and thanks for reviewing!

reeds/input_processing/mcs_sampler.py
bsergi: We don't actually use the `sample_num` in the random approach as it's only used to reference back to the sampled Latin Hypercube matrix, so might make sense to just remove it when random is on. I'll probably do that but if you think it makes sense to keep I like the idea of having them line up so can change that.

reeds/input_processing/mcs_sampler.py
bsergi: A couple of thoughts on this one, let me know your reactions:
- The LHS sampling is done via `scipy.stats.qmc.LatinHypercube`. From what I can tell `np.random` isn't used by those methods and it isn't used elsewhere in the `mcs_sampler.py` script, so I don't think setting this would impact things. 
- The function I mentioned above does take a seed argument, and we do pass one to it (the default it zero). I just did a quick test of two different runs and got the same values for the LHS matrix, so I think we've got the reproducibility piece covered.
  - One thing to note is that the changing the number of samples or dimensions changes the matrix. That means that otherwise two identical runs with N=50 and N=100, or two runs with the same N but where one sampled UPV costs the other sampled UPV + gas prices, would have different values even for shared parameters. So, the seed only guarantees reproducibility for the same setup. 
- There isn't a straightforward way to extend the LHS matrix with more samples and still have it be a Latin Hypercube; the reason is that the sampling method bins the distribution and puts one sample in each bin, so adding new samples results in overlap. 
  - Your suggestion about setting the explicit seed would work for the random approach so I can add support for that. It's already in the python script so we just need to add to ReEDS; I put in as a scalar but let me know if you think a switch is more appropriate here. I also added some explanation on the usage in the docs/user_guide section.

runreeds.py
```diff
@@ -92,7 +92,7 @@ def create_case_lists(df_cases:pd.DataFrame, BatchName:str, single:str=''):
         # If --single/-s was passed, only keep those cases (regardless of ignore)
         # otherwise, drop any case marked ignore
         if single:
-            if case not in single.split(','):
+            if not sum([s in case for s in single.split(',')]):
```
patrickbrown4: Hm, this might give weird behavior since some case names are subsets of other case names. I think my suggestion would just be to ignore the issue I raised about using `--single` for Monte Carlo, since it's kind of ill-defined for multi-case MC runs anyway, and is not directly related to the rest of the PR.

reeds/input_processing/mcs_sampler.py
patrickbrown4: Ok thanks, that all makes sense. Sorry to complicate things. I think the way you have it now, with the `MCS_seed` being used by both the random and LHS methods, is great.

runreeds.py
bsergi: sounds good, will revert

Issue #38, opened 2026-04-20
Move data loaded into `b_inputs.gms` from `.csv` files to one `inputs.h5` → `inputs.gdx` file


patrickbrown4:
## Summary

`b_inputs.gms` is almost 7000 lines long, of which almost 2000 lines are just for reading in ~270 `.csv` files. `.csv` files are also much less storage-space-efficient than numeric formats like `.h5`. To support the long-term aspiration of a free software base (i.e. not GAMS) for ReEDS and make the ReEDS inputs more portable, we'd like to start moving away from `.csv` files for intermediate data handling in ReEDS. (We'd still use them for user-facing input files, but less for reading data into GAMS.)

## Overview of proposed structure

[1103](https://github.nrel.gov/ReEDS/ReEDS-2.0/pull/1103) on the old repo tried something similar in terms of simplifying `b_inputs.gms`, but was never merged, and still relied on `.csv` files. Here, the idea is that instead of writing `.csv` files from the `input_processing` scripts, we create an `inputs.h5` container, write the inputs to that, convert it to `inputs.gdx`, and then load it into `b_inputs.gms` in one line. A graphical outline of the proposed structure is below:

<img width="2180" height="516" alt="Image" src="https://github.com/user-attachments/assets/8440f636-6648-49da-a88e-28b70574ae45" />

## Components

- [x] Create `inputs.h5` container and the functions to write to it easily (start from the `write_input_to_h5()` [function](https://github.nrel.gov/ReEDS/ReEDS-2.0/commit/da5aebb1075ba39d9898524dcee6143a394b8732))
- [x] Create `h5_to_gdx.py` script to convert `inputs.h5` to `inputs.gdx` and load `inputs.gdx` into `b_inputs.gms`
  - [ ] Compare `gamspy`, `gams.transfer`, and `gdxpds` to see which is fastest and use the winner for `h5_to_gdx.py`
- [ ] Modify `input_processing` scripts to write to `inputs.h5` instead of `.csv` files

Then we would start moving other data processing from `b_inputs.py` into upstream python scripts, but that will be tracked in a separate issue.

(moved from https://github.nrel.gov/ReEDS/ReEDS-2.0/issues/1947)

patrickbrown4:
Addressed by #61

PR #61, opened 2026-04-26
Create an `inputs.h5` -> `inputs.gdx` pipeline to help move processing out of `b_inputs.gms`

patrickbrown4:
## Summary

This PR addresses #38 by creating a new `inputs.h5` container and an `h5_to_gdx.py` script that feeds it into GAMS. This structure helps facilitate the exploration of non-GAMS approaches for ReEDS by letting input-processing calculations be moved out of `b_inputs.gms` without relying on hundreds of new `.csv` files for data transfer.

### Implementation notes 

- The interface to the new data structure is provided a pair of new functions, `reeds.io.write_input_to_h5()` and `reeds.io.read_input()`
  - For example, to write an input file and load it into GAMS, the old approach of:
    ```python
    co2_cap.round(0).to_csv(os.path.join(inputs_case, 'co2_cap.csv'))
    ```
    in python followed in GAMS by
    ```gams
    parameter co2_cap(allt)      "--metric tons-- CO2 emissions cap used when Sw_AnnualCap is on"
    /
    $offlisting
    $ondelim
    $include inputs_case%ds%co2_cap.csv
    $offdelim
    $onlisting
    / ;
    ```
    is now
    ```python
    reeds.io.write_input_to_h5(
        co2_cap, 'co2_cap', inputs_case, gamstype='parameter',
        comment='--metric tons-- CO2 emissions cap used when Sw_AnnualCap is on',
    )
    ```
    with the GAMS declaration/load happening automatically.
  - Similarly, when loading input parameters, `val_r = pd.read_csv(os.path.join(case,'inputs_case','val_r.csv'), header=None)` is now `val_r = reeds.io.read_input(case, 'r')`
- A new script, `h5_to_gdx.py`, runs as the last `input_processing` script and converts `inputs.h5` to `inputs_0.gdx`
  - It also writes `b_declare_(sets|parameters).gms` and `b_load_(sets|parameters).gms` files to facilitate reading the `.gdx` file in `b_inputs.gms` (I couldn't figure out how to get it to work with `$declareAndLoad inputs_case%ds%inputs_0.gdx` alone; it wasn't recognizing subsets for domain checking)
- New guidelines:
  - New aliases should be added to `inputs/sets/_aliases.csv`
  - New sets should follow the guidelines in `inputs/sets/README.md`
  - Use `reeds.io.write_input_to_h5()` when adding new inputs instead of writing them to a `.csv` file and loading explicitly in `b_inputs.gms`
  - New input-defining code should happen in python (feel free to add new scripts to `input_processing` as necessary) instead of `b_inputs.gms`
- I started moving some sets/parameters to the new structure as examples but didn't want to make this PR too huge. The rest can be distributed across smaller PRs.
  - I'll open a new issue tracking next steps for transitioning to this new structure and moving more processing out of `b_inputs.gms`/`e_report.gms` once this PR is merged

### Additional changes 

- Fixed bokeh processing of health impacts
- Renamed `sw.csv` -> `wst_surface.csv` to avoid confusion with switches (which are abbreviatd in python as `sw`)
- Renamed `gen_mandate_tech_list.csv` -> `nat_gen_tech_frac.csv` to match its name in GAMS

### Issues resolved 

#38

Zero change for the Pacific test case:
[results-v20260426_mainM0_Pacific,v20260426_inputsM1_Pacific.pptx](https://github.com/user-attachments/files/27104490/results-v20260426_mainM0_Pacific.v20260426_inputsM1_Pacific.pptx)

The only changes to `inputs.gdx` (aside from some minor parameter renaming) are due to rounding: [diff_inputs-v20260426_mainM0_Pacific-v20260426_inputsM1_Pacific.gdx.zip](https://github.com/user-attachments/files/27104527/diff_inputs-v20260426_mainM0_Pacific-v20260426_inputsM1_Pacific.gdx.zip)

Near-zero change (rounding differences only) for the USA_defaults case: [results-mainK0_USA_defaults,inputsK0_USA_defaults.pptx](https://github.com/user-attachments/files/27538776/results-mainK0_USA_defaults.inputsK0_USA_defaults.pptx)

### 20260522
Double checked, and still only rounding-error differences in the Pacific test case. I also made sure the MonteCarlo_LHS case still works.

### Details to double-check

  - [x] Units are specified
  - [x] New large data files handled with .h5 instead of .csv

No


reeds/io.py
```diff
@@ -1676,6 +1723,97 @@ def write_to_h5(
...
+    if isinstance(dfwrite.index, pd.MultiIndex) or dfwrite.index.name:
+        dfwrite = dfwrite.reset_index()
+    ## Format for GAMS: The final column of a parameter should be named 'Value' and
+    ## should contain the data as floats; all the other columns are treated as indices
+    if gamstype == 'parameter':
+        dfwrite.columns = dfwrite.columns.tolist()[:-1] + ['Value']
+    ### Write record to h5 file
+    attrs = {'gamstype': gamstype.lower()}
+    if len(comment):
+        attrs['comment'] = comment
```
patrickbrown4: What do you think about adding two additional (optional) attributes:
- `units`
  - I've wondered for a while if we could build in explicit unit-checking (using one of a variet of python packages like `pint` or `astropy.units`) to the constraint formulations once we switch to python, to avoid mistakes like adding MW to MWh. Adding the units as we standardize the inputs would give us that option.
- `written_by`
  - It's a little harder to find where some sets/parameters are defined now that you can't search for `val_r.csv` (for example), so this might be helpful for debugging

Both could get parsed into the GAMS comment as `[{units}] {comment} (written by {written_by})`

wesleyjcole: I like this idea.  I also think that it would be good for them not to be optional.  We already "require" units to be specified, even if it is unitless.  The `written_by` default would be `copy_files.py`, so I think requiring it would also work.

input_processing/writecapdat.py
```diff
@@ -830,17 +841,27 @@ def main(reeds_path, inputs_case, agglevel, regions):
...
     if agglevel_variables['lvl'] == 'mult':
         for resolution in agglevel_variables['agglevel']:
             if resolution == 'aggreg':
-                aggreg_data  = main(reeds_path, inputs_case, agglevel=resolution, 
-                                     regions=agglevel_variables['ba_regions'] )
+                aggreg_data, _comments = main(
+                    reeds_path, inputs_case, agglevel=resolution, 
+                    regions=agglevel_variables['ba_regions'],
+                )
+                coments = {**comments, **_comments}
```
wesleyjcole: "coments" is a typo here and below.

```suggestion
                comments = {**comments, **_comments}
```

inputs/national_generation/nat_gen_tech_frac.csv
```diff
@@ -0,0 +1,140 @@
...
+csp1_4,1,1,1,1
+csp1_5,1,1,1,1
+csp1_6,1,1,1,1
+csp1_7,1,1,1,1
+csp1_8,1,1,1,1
+csp1_9,1,1,1,1
+csp1_10,1,1,1,1
+csp1_11,1,1,1,1
+csp1_12,1,1,1,1
+csp1_10,1,1,1,1
```
wesleyjcole: Looks like this line was unintentionally added twice (appears 3 lines up).

```suggestion
```

input_processing/h5_to_gdx.py
```diff
@@ -0,0 +1,190 @@
...
+import numpy as np
+import pandas as pd
+from pathlib import Path
+from typing import Literal
+sys.path.append(str(Path(__file__).parent.parent))
+import reeds
+
+
+#%% Functions
+def read_inputs(case):
```
wesleyjcole: Your other functions in this script have docstrings, but this one doesn't.  Probably worth adding for consistency.

input_processing/h5_to_gdx.py
```diff
@@ -0,0 +1,190 @@
...
+        if len(writelist):
+            for line in writelist:
+                f.write(f'{gamstype} {line} ;\n')
+                name = line.split('(')[0]
+                for alias in aliases.get([name], []):
+                    f.write(f'alias({name},{alias}) ;\n')
+    print(f'Wrote {fpath}')
+
+
+def write_load(
```
wesleyjcole: Not sure this is worth a change, but when I saw this function name, I thought it was a function to write the electricity load data, not to write the load statements.

input_processing/writecapdat.py
```diff
@@ -830,17 +841,27 @@ def main(reeds_path, inputs_case, agglevel, regions):
...
             if resolution == 'ba':
-                ba_data = main(reeds_path, inputs_case, agglevel=resolution, 
-                                regions=agglevel_variables['ba_regions'])
+                ba_data, _comments = main(
+                    reeds_path, inputs_case, agglevel=resolution, 
+                    regions=agglevel_variables['ba_regions'],
+                )
+                coments = {**comments, **_comments}
```
wesleyjcole:
```suggestion
                comments = {**comments, **_comments}
```

input_processing/writecapdat.py
```diff
@@ -830,17 +841,27 @@ def main(reeds_path, inputs_case, agglevel, regions):
...
             if resolution == 'county':
-                county_data = main(reeds_path, inputs_case, agglevel=resolution,
-                                     regions=agglevel_variables['county_regions'],)
+                county_data, _comments = main(
+                    reeds_path, inputs_case, agglevel=resolution,
+                    regions=agglevel_variables['county_regions'],
+                )
+                coments = {**comments, **_comments}
```
wesleyjcole:
```suggestion
                comments = {**comments, **_comments}
```

reeds/input_processing/writecapdat.py
```diff
@@ -884,11 +905,20 @@ def main(reeds_path, inputs_case, agglevel, regions):
         'cap_cspns': True,
         'can_imports_capacity': True,
     }
+    gamstype = {
+        'pcat': 'set',
```
wesleyjcole, line 909: Just flagging that `pcat` is going away in #12.

inputs/sets/_aliases.csv
wesleyjcole, line 1: Why the underscore for aliases and pcat?  None of the others use that convention.

inputs/sets/ctt.csv
```diff
@@ -1,5 +1,5 @@
-o "once through",
```
wesleyjcole, line 1: Is it feasible to keep comments in the set files?  I see you moved these to the ReadMe--it just seems more useful here than there, but keeping it in the file might be a hassle.

inputs/sets/README.md
```diff
@@ -1,4 +1,48 @@
...
+  - `ss`: saline surface water
+  - `ww`: wastewater effluent
+
+## Special-case files
+
+- `_aliases.csv`: aliases (extra names for the same set) used in GAMS
+  - Aliases of primary sets should be added here
+  - Aliases of sets defined in `b_inputs.gms` (e.g., `h`→`hh`) should instead be defined in GAMS after the set definition
+- `_pcat.csv`: prescribed capacity categories
+  - The `pcat` set in GAMS (defined in `writecapdat.py`) includes the members of the `i` set; this file includes only the *extra* elements on top of the `i` set
```
wesleyjcole, line 48: I see here that these are special case file.  Is that why the underscore is used, as a flag that this won't exactly match the GAMS set?

reeds/io.py
```diff
@@ -1679,6 +1727,101 @@ def write_to_h5(
...
+    key:str,
+    case:str|Path,
+    gamstype:Literal['set','parameter']='set',
+    comment:str='',
+    units:str='',
+    verbose=1,
+    **kwargs,
+):
+    """
+    """
```
wesleyjcole, line 1809: Would you add text to the docstring?

wesleyjcole (commented):
I like how tidy this makes the change.

I did see that the comments are not printed in the b_declare_{}.gms files.  Would that be easy to add?  I don't know if it's important, but there might be an instance where having comments there would be helpful.

inputs/sets/README.md
patrickbrown4: Yeah that's right; I thought it would be confusing if all the csv files in `inputs/sets` match the sets used in the model except for these files.

inputs/sets/ctt.csv
patrickbrown4: I'm not sure if `gdxpds` supports element-level comments, and it would add a lot of overhead to propagate them to the `inputs.h5` file. To me it doesn't seem worth it for something that's only used in two sets. I also feel like `README.md` is a more suitable place for documentation than raw .csv files.

reeds/input_processing/writecapdat.py
patrickbrown4: I thought that was the case, but it's still used in `b_inputs.gms`; here's a snippet from that PR branch: https://github.com/ReEDS-Model/ReEDS/blob/06c948a863bcc926bb07378c425d9c7048a7c0f5/b_inputs.gms#L1179-L1208

inputs/sets/_aliases.csv
patrickbrown4: Just to set them apart because they are not read directly into GAMS (explained in `inputs/sets/README.md`)

patrickbrown4:
> I did see that the comments are not printed in the b_declare_{}.gms files. Would that be easy to add? I don't know if it's important, but there might be an instance where having comments there would be helpful.

I wasn't thinking of `b_declare_{sets or parameters}.gms` as files that people would actually want or need to look at; for me they're just a workaround to avoid domain violations for subsets when loading from a `.gdx` file. Do you think it's enough to have the comments in the `inputs.gdx`/`inputs.h5` files instead? (In my mind, the code itself should ideally not be user facing, while these kinds of uniformly formatted data containers would be.)

b_inputs.gms
```diff
@@ -131,109 +82,16 @@ $onlisting
...
-$include inputs_case%ds%val_hurdlereg.csv
-/
-;
+* Written by h5_to_gdx.py
+$include b_declare_sets.gms
+$include b_declare_parameters.gms
+$gdxin inputs_case%ds%inputs_0.gdx
+$include b_load_sets.gms
+$include b_load_parameters.gms
+$gdxin
```
patrickbrown4: This approach could be simplified by using `gams.transfer` to write regular (not relaxed) sets, and then loading the whole file with `$declareAndLoad`. But since it would require an update to `environment.yml` and only affects the code in two places (these lines and `h5_to_gdx.py`), my preference is to do it in a followup PR and move ahead with the current approach for now, so we can start adopting the new input formalism sooner rather than later.

reeds/input_processing/writecapdat.py
wesleyjcole: This has now been removed from #12, but I'm assuming #12 will come in after this PR, so we'll just need to address it when we merge the two together.

wesleyjcole (approved):

reeds/io.py
```diff
@@ -450,6 +452,60 @@ def get_h2_storage_sites(h2_storage_type="salt"):
...
+    if that doesn't exist, an error is raised.
+
+    Args:
+        case (str or Path): Absolute path to a ReEDS case
+        name (str): Name of the ReEDS parameter or {case}/inputs_case/{name}.csv
+
+    Returns:
+        pd.DataFrame
+    """
+    key = Path(name).stem
```
kodiobika, line 489: Just curious - are there any cases where `name` would be different from `key`? Or is this just in case someone passes a relative file path instead of just a name?

reeds/io.py
```diff
@@ -1724,6 +1772,120 @@ def write_to_h5(
...
+        key,
+        h5path,
+        attrs=attrs,
+        **kwargs,
+    )
+    if verbose:
+        print(f'{Path(h5path).name}: Wrote {key} from {calling_file}')
+
+
+def write_csv_to_h5(
```
kodiobika: Could you either rename this to `write_csv_to_inputs_h5` or something similar (since it only ever writes to `inputs.h5`), or add an argument to specify an .h5 filepath? Also "copy" rather than "write" would be a bit clearer to me as the first word of the function name, but I think either's fine

reeds/io.py
```diff
@@ -1724,6 +1772,120 @@ def write_to_h5(
                 )
 
 
+def write_input_to_h5(
```
kodiobika: Similar to the other comment, I'm wondering if the function name makes it clear enough that this only writes to `inputs.h5` as opposed to writing an input to an arbitrary .h5 file. Maybe something like `write_to_inputs_h5` would be clearer?

reeds/input_processing/WriteHintage.py
```diff
@@ -598,7 +592,47 @@ def main(reeds_path, inputs_case):
...
+    # 1. initial capacity: init-1, init-2, ..., init-n
+    # 2. prescribed capacity: prescribed
+    # 3. new capacity: new
+    # This allows us to distinguish between existing, prescribed, and model-chosen builds
+    # The number of classes is set by numhintage for initial capacity and numclass for new capacity
+    # Under the Clean Air Act, all coal plants are regulated individually.
+    # Therefore, we need a large value of hintages to represent these plants
+    max_hintage_number = tdat['bin'].max()
+    if sw.numhintage in ['unit', 'group']:
+        numhintage = 300
```
kodiobika: Can we specify this in `scalars.csv`?

reeds/input_processing/recf.py
```diff
@@ -523,8 +523,11 @@ def main(reeds_path, inputs_case):
     ### Overwrite the original hierarchy.csv based on capcredit_hierarchy_level
     hierarchy.rename_axis('*r').to_csv(
         os.path.join(inputs_case, 'hierarchy.csv'), index=True, header=True)
-    pd.Series(hierarchy.ccreg.unique()).to_csv(
-        os.path.join(inputs_case,'ccreg.csv'), index=False, header=False)
+    ccreg = pd.Series(hierarchy.ccreg.unique(), name='*')
```
kodiobika: Does the series always need a name when calling `write_input_to_h5` with `gamstypes='set'`? If so, could we either add a check to the function raising an error for unnamed series or rename unnamed series to "*" within the function?

reeds/io.py
```diff
@@ -450,6 +452,60 @@ def get_h2_storage_sites(h2_storage_type="salt"):
 
 
 ### Read files from a ReEDS case
+def read_h5(h5path:str|Path, key:str) -> pd.DataFrame():
```
kodiobika:
```suggestion
def read_h5(h5path:str|Path, key:str) -> pd.DataFrame:
```

kodiobika (approved):
Looks great, thanks!

reeds/io.py
patrickbrown4: Yeah, the idea was just to standardize the input so the user can provide either `name` or `{case}/inputs_case/{name}.csv` or `{name}.csv`. I can't remember if there's a specific place where it's used that second way in the code though. If you think it's unnecessary/confusing to have that flexibility, we could just go with the direct `name` input instead.

kodiobika: Got it thanks, and no worries I think it's fine as-is

reeds/io.py
patrickbrown4: Sounds good, I went with `write_csv_to_inputs_h5()` because it's not exactly a copy given the dtype/name changes, but I agree it's better to specify `inpugs_h5` instead of just `h5`

Issue #34, opened 2026-04-17
Compare cases and plotting changes


Max-Vanatta:
### Summary
This issue is to document and outline the potential work which still needs to be done for compare_cases and other connected plotting functions. 

### Proposed Changes

- [ ] update ppt to new NLR ppt template
- [ ] Change technology colormap for the H2-CC/CT overlap with bio sources (see [1818](https://github.nrel.gov/ReEDS/ReEDS-2.0/issues/1818) on the old repo)
- [ ] Discussion: Is H2-CC/CT a) generation or b) long term storage from H2 production to H2 generation
- [ ] rewrite the 2-plot function from plots.py to be more integrated with compare_cases
- [ ] move the output collection out of Bokeh for compare_cases and other plots.
- [ ] clarify the formulation of NPV for each method and include in results.

nice to have (from Patrick):

- [ ] Either put data in dictionaries like diagnostic_plots.py or create output container class like uncertainty_plots.py
- [ ] Allow user to specify which premade plots from {casename}/outputs/figures to include in the output presentation
- [ ] Move some functionality from uncertainty_plots.py into compare_cases.py

PR #65, opened 2026-04-28
Implement daily temperature-based gas price adjustments 

SoLaraS2:
## Summary
This PR calculates daily gas price multipliers based on a linear regression fitting heating and cooling degree days on daily deviations from annual regional gas prices (the parameters of which are in `inputs/fuelprices/gasreg_price_adj_regression_parameters.csv`). The user can choose between these daily regional multipliers, the existing national wintertime markup, and no adjustment.

## Technical details
- Added `gasreg` hierarchy level, which represents the regions at which the degree day / price deviation regressions were fitted. These are mostly just census divisions, with the exceptions being "California" and "Northwest", which are components of the "Pacific" cendiv, and "Southwest" and "Mountain", which are components of the "Mountain" cendiv.
- Daily `gasreg`-level gas price multipliers are calculated in `reeds/input_processing/fuelcostprep.py` by calculating state-level daily degree days for the weather years of the given run and aggregating them via population-weighted average. To create multipliers at the `r` level, the `gasreg`-level multipliers are copied to their constituent zones. To create multipliers at the `cendiv` level, the `gasreg`-level multipliers are aggregated via population-weighted average.
- Added `get_daily_gas_price_multipliers` function to hourly_writetimeseries.py to get a GAMS compatible csv with representative hourly NG price multipliers. In this function, the multipliers for each gasreg are renormalized so that the average across the representative periods is 1, ensuring the year-round average gas price doesn't change.

### Additional changes
- Added a new switch `GSw_GasPriceAdjMethod` which controls whether the daily price multipliers are applied, the national wintertime markup is applied, or no adjustment is applied.
- In `main`, the wintertime markup of 1.054 (`szn_adj_gas_winter` in `scalars.csv`) mentioned in the model documentation is only being applied when `GSw_GasCurve = 1`. A separate wintertime multiplier of 1.04 (`gasprice_ref_frac_adder` in `scalars.csv`) is being applied when `GSw_GasCurve = 3`. No multiplier is applied when `GSw_GasCurve = 0 or 2`. In this PR, the `gasprice_ref_frac_adder` multiplier is deleted and the `szn_adj_gas` multiplier is applied in every `GSw_GasCurve` scenario (assuming `GSw_GasPriceAdjMethod = 1`).
- In `main`, the wintertime markup only increases the wintertime price but doesn't make corresponding adjustments to maintain the pre-markup year-round average price. As a result, the year-round natural gas price is being erroneously inflated. In this PR, the seasonal multipliers are renormalized so that wintertime prices are higher than nonwinter prices while the year-round average price remains unchanged.

- Added `GSw_GasPriceAdjMethod` (0 = no adjustment, 1 = national wintertime markup, 2 = daily adjustments based on regional temperatures)

Results comparing `main` and this branch using different `GSw_GasPriceAdjMethod` and `GSw_GasCurve` values are below.

`GSw_GasCurve = 1` - More gas is built due to the fact that the wintertime markup is now renormalized so that the year-round average gas price remains unchanged (previously we only increased the wintertime gas price without a commensurate decrease in the non-wintertime price, which increased the year-round average price). The non-adjustment and seasonal adjustment cases see up to a ~30 TWh increase in gas generation, while the daily price adjustment case sees up to a ~60 TWh increase.
- `GSw_GasPriceAdjMethod = 1`: [results-0707_Main_USA_defaults,0707_NatGas_USA_defaults.pptx](https://github.com/user-attachments/files/29848317/results-0707_Main_USA_defaults.0707_NatGas_USA_defaults.pptx)
- `GSw_GasPriceAdjMethod = 2`: [results-0707_Main_USA_defaults,0707_NatGas_USA_DailyAdj.pptx](https://github.com/user-attachments/files/29848306/results-0707_Main_USA_defaults.0707_NatGas_USA_DailyAdj.pptx)
- `GSw_GasPriceAdjMethod = 0`: [results-0707_Main_USA_defaults,0707_NatGas_USA_NoAdj.pptx](https://github.com/user-attachments/files/29848314/results-0707_Main_USA_defaults.0707_NatGas_USA_NoAdj.pptx)

`GSw_GasCurve = 2` - Differences range from no change to more gas buildout. The non-adjustment case results in no changes, which makes sense because no multiplier was being applied initially. The seasonal adjustment case has slightly lower gas generation in early years and slightly higher gas and wind generation replacing solar in later years. The daily adjustment case sees up to a ~50 TWh increase in gas generation.
- `GSw_GasPriceAdjMethod = 1`: [results-0707_Main_USA_GasCurve2,0707_NatGas_USA_GasCurve2.pptx](https://github.com/user-attachments/files/29848304/results-0707_Main_USA_GasCurve2.0707_NatGas_USA_GasCurve2.pptx)
- `GSw_GasPriceAdjMethod = 2`: [results-0707_Main_USA_GasCurve2,0707_NatGas_USA_GasCurve2_DailyAdj.pptx](https://github.com/user-attachments/files/29848291/results-0707_Main_USA_GasCurve2.0707_NatGas_USA_GasCurve2_DailyAdj.pptx)
- `GSw_GasPriceAdjMethod = 0`: [results-0707_Main_USA_GasCurve2,0707_NatGas_USA_GasCurve2_NoAdj.pptx](https://github.com/user-attachments/files/29848295/results-0707_Main_USA_GasCurve2.0707_NatGas_USA_GasCurve2_NoAdj.pptx)

`GSw_GasCurve = 3` - Differences are fairly minimal across the board. The largest change is in the non-adjustment case, where 2040 sees a ~6 TWh change in generation, which is mostly a shift from gas-cc to gas-ct. The other cases just have changes on the scale of rounding errors.
- `GSw_GasPriceAdjMethod = 1`: [results-0707_Main_USA_GasCurve3,0707_NatGas_USA_GasCurve3.pptx](https://github.com/user-attachments/files/29848287/results-0707_Main_USA_GasCurve3.0707_NatGas_USA_GasCurve3.pptx)
- `GSw_GasPriceAdjMethod = 2`: [results-0707_Main_USA_GasCurve3,0707_NatGas_USA_GasCurve3_DailyAdj.pptx](https://github.com/user-attachments/files/29848279/results-0707_Main_USA_GasCurve3.0707_NatGas_USA_GasCurve3_DailyAdj.pptx)
- `GSw_GasPriceAdjMethod = 0`: [results-0707_Main_USA_GasCurve3,0707_NatGas_USA_GasCurve3_NoAdj.pptx](https://github.com/user-attachments/files/29848283/results-0707_Main_USA_GasCurve3.0707_NatGas_USA_GasCurve3_NoAdj.pptx)

`GSw_GasCurve = 0` - Differences range from no change to more gas buildout. The non-adjustment case results in no changes, which makes sense because no multiplier was being applied initially. The seasonal adjustment case sees up to a ~20 TWh increase in gas generation, and the daily adjustment case sees up to a ~70 TWh increase.
- `GSw_GasPriceAdjMethod = 1`: [results-0707_Main_USA_GasCurve0,0707_NatGas_USA_GasCurve0.pptx](https://github.com/user-attachments/files/29873694/results-0707_Main_USA_GasCurve0.0707_NatGas_USA_GasCurve0.pptx)
- `GSw_GasPriceAdjMethod = 2`: [results-0707_Main_USA_GasCurve0,0707_NatGas_USA_GasCurve0_DailyAdj.pptx](https://github.com/user-attachments/files/29866556/results-0707_Main_USA_GasCurve0.0707_NatGas_USA_GasCurve0_DailyAdj.pptx)
- `GSw_GasPriceAdjMethod = 0`: [results-0707_Main_USA_GasCurve0,0707_NatGas_USA_GasCurve0_NoAdj.pptx](https://github.com/user-attachments/files/29866538/results-0707_Main_USA_GasCurve0.0707_NatGas_USA_GasCurve0_NoAdj.pptx)

### Details to double-check

  - [x] Units are specified
  - [x] Preprocessing steps have been documented and committed to [ReEDS_Input_Processing](https://github.com/ReEDS-Model/ReEDS_Input_Processing): https://github.com/ReEDS-Model/ReEDS_Input_Processing/tree/main/aeo_updates/temperature_gas_price_adj_regression
- Verify all the temporal tests from `cases_test.csv` still work
  - [x] Pacific_CC
  - [x] Pacific_weks
  - [x] Pacific_full_year
  - [x] Interday_storage
  - [x] Pacific_rep15

- [X] No large data file(s) added/modified
- [X] No change to process flow (runbatch.py, d_solve_iterate.py)
- [X] No change to code organization
- [X] No change to package requirements (environment.yml or Project.toml)

- LLMs used for timestamp compatibility between different outputs to ensure consistency in the final csv outputted
- LLMs used to run through the files and find relevant predefined functions I (Lara) may have originally missed
- LLMs used to debug h5 file handling


reeds/io.py
```diff
@@ -1909,3 +1909,237 @@ def get_folder_size(casedir):
     # convert to GB
     total_size /= 1e9
     return total_size
+
+def get_degree_days(case, base_temp_c=18.3333333333, hourly_formula=False):
```
patrickbrown4: `io.py` is kind of crowded but it's meant to be focused on file in/out rather than actual calculations. Could you either create a new `reeds/temperature.py` module and add this function there, or define it in `fuelcostprep.py` since that's the only place it's used? (#36 will move the `input_processing` scripts into `reeds`, so the latter approach would still make it accessible to other scripts once that PR is merged.)

inputs/zones/state_groups.csv
```diff
@@ -1,49 +1,49 @@
...
-TN,USA,East_South_Central,appalachia,Southeast
-TX,USA,West_South_Central,southern-plains,Texas
-UT,USA,Mountain,mountain,Mountain
-VA,USA,South_Atlantic,appalachia,MidAtlantic
-VT,USA,New_England,northeast,NewEngland
-WA,USA,Pacific,pacific,Northwest
-WI,USA,East_North_Central,lake-states,Midwest
-WV,USA,South_Atlantic,appalachia,MidAtlantic
-WY,USA,Mountain,mountain,Mountain
+st,country,cendiv,usda_region,h2ptcreg,ngreg
```
patrickbrown4: 1. How about `gasreg` instead of `ngreg`, since the techs are `gas-cc`/`gas-ct` and we don't use `ng` much throughout the code?
2. We'll need to add the new region level to <https://reeds-model.github.io/ReEDS/model_documentation.html#figure-hierarchy> (and describe the new method in the docs). If the regions are finalized, let me know and I can remake the figure.

reeds/io.py
```diff
@@ -1909,3 +1909,237 @@ def get_folder_size(casedir):
...
+
+    for t in model_years:
+        idx = pd.date_range(f'{t}-01-01', f'{t}-12-31', freq='D')
+
+        # 365-day calendar excludes Dec 31, not Feb 29
+        idx = idx[~((idx.month == 12) & (idx.day == 31))]
+
+        # only needed if model year is leap year
+        if len(idx) > 365:
+            idx = idx[~((idx.month == 2) & (idx.day == 29))]
```
patrickbrown4: Normal years should keep 12/31; leap years should drop 12/31 and keep 2/29

input_processing/hourly_writetimeseries.py
```diff
@@ -225,6 +225,70 @@ def get_yearly_demand(sw, hmap_myr, hmap_allyrs, inputs_case, periodtype='rep'):
 
     return load_in, load_out
 
+
+def get_natgas_price_diffs(sw, hmap_myr, hmap_allyrs, inputs_case, periodtype='rep'):
```
patrickbrown4: Since this file is modified, can you run all the temporal tests from `cases_test.csv` and make sure they still work?
- [x] Pacific_CC
- [x] Pacific_weks
- [x] Pacific_full_year
- [x] Interday_storage
- [x] Pacific_rep7

inputs/zones/state_groups.csv
kodiobika: `gasreg` SGTM and thanks for the heads up on the docs. The regions are finalized so if you could remake the figure that'd be great, thanks!

input_processing/fuelcostprep.py
```diff
@@ -138,6 +140,296 @@
 fuel = fuel.merge(h2fuel,on=['t','r'],how='left')
 fuel = fuel.sort_values(['t','r'])
 
+#%%#################################### 
+### Natural Gas Price Diffs ###
+
+def get_degree_days(case, base_temp_c=18.3333333333, hourly_formula=False):
```
kodiobika: A nit-picky suggestion that I think makes this line a bit more clean/readable (in future comments, I'll introduce similar suggestions with "nit:" - see https://www.augmentcode.com/guides/what-does-nit-mean-in-code-review for context). 18.33 is a slightly less precise but that's alright, I think it's worth making it look cleaner (and we round a lot of figures throughout the model anyway)

```suggestion
def get_degree_days(case, base_temp_celsius=18.33, hourly_formula=False):
```

kodiobika: For clarity on what's going into the function, can we add `reeds_path` as an argument and change `case` to `inputs_case`?

input_processing/fuelcostprep.py
```diff
@@ -138,6 +140,296 @@
...
+    Return daily HDD/CDD by gasreg for the modeled years in this case,
+    using temperature shapes from GSw_HourlyWeatherYears and annual totals
+    from gasreg_hdd.csv / gasreg_cdd.csv.
+    """
+    case = reeds.io.standardize_case(case)
+    inputs_case = os.path.join(case, 'inputs_case')
+    sw = reeds.io.get_switches(case)
+
+    # modeled years for outputs
+    model_years = reeds.io.get_years(case)
```
kodiobika: nit: Since this function is typically going to be called with `inputs_case`, I think we can delete the first two lines. Also `get_switches` already standardizes the case internally, so we can pass `inputs_case` there, and then we can pass the parent directory to `get_years` (functionally no change but I think it's slightly cleaner to use `inputs_case` everywhere)

```suggestion
    sw = reeds.io.get_switches(inputs_case)

    # modeled years for outputs
    model_years = reeds.io.get_years(os.path.dirname(inputs_case))
```

input_processing/fuelcostprep.py
```diff
@@ -138,6 +140,296 @@
...
+    model_years = np.array([y for y in model_years if y <= int(sw.endyear)], dtype=int)
+
+    # weather years for temperature shapes
+    weather_years = np.array(
+        [int(y) for y in str(sw.GSw_HourlyWeatherYears).split('_') if str(y).strip()],
+        dtype=int,
+    )
```
kodiobika: nit: Suggestion to make this a little simpler:
```suggestion
    weather_years = [int(y) for y in sw.GSw_HourlyWeatherYears.split('_')]
```

input_processing/fuelcostprep.py
```diff
@@ -138,6 +140,296 @@
...
+        .rename(columns={'state': 'st'})
+        .dropna(subset=['st'])
+        .groupby('st', as_index=True)['population']
+        .sum()
+        .rename_axis('st')
+        .rename('population')
+        .to_frame()
+    )
+
+    pop_state = pop_state.loc[pop_state.index.intersection(valid_states)].copy()
```
kodiobika: nit: To avoid having to read in the valid states, can we move the temperature calculation to be above the population calculation and then filter on the temperature columns (which will already be the valid states)?

```suggestion
    pop_state = pop_state.loc[temp.columns].copy()
```

input_processing/fuelcostprep.py
```diff
@@ -138,6 +140,296 @@
...
+        .tz_localize('UTC')
+        .tz_convert('Etc/GMT+6')
+    )
+
+    # subset to selected weather years after timezone conversion
+    temp = temp.loc[temp.index.year.isin(weather_years)].copy()
+
+
+    # keep valid states only
+    temp = temp[[c for c in temp.columns if c in valid_states]].copy()
```
kodiobika: We actually already have a `get_temperatures` function so let's use that instead. If the population calculation is relocated (see other comment), this also lets us remove the valid_states code above since the function does the filtering internally

```suggestion
    temp = reeds.io.get_temperatures(inputs_case)
    temp = temp.loc[temp.index.year.isin(weather_years)]
```

input_processing/fuelcostprep.py
```diff
@@ -138,6 +140,296 @@
...
+    ddh = pd.read_csv(
+        os.path.join(reeds_path, 'inputs', 'fuelprices', 'gasreg_hdd.csv'),
+        index_col=0,
+    )
+    ddc = pd.read_csv(
+        os.path.join(reeds_path, 'inputs', 'fuelprices', 'gasreg_cdd.csv'),
+        index_col=0,
+    )
+    ddh.index = ddh.index.astype(int)
+    ddc.index = ddc.index.astype(int)
```
kodiobika: Let's omit these just to cut down on # of lines. I think we can expect the index to at least be numeric (even if non-int) already, which is good enough

```suggestion
```

input_processing/fuelcostprep.py
```diff
@@ -138,6 +140,296 @@
...
+
+    # annual gasreg totals
+    ddh = pd.read_csv(
+        os.path.join(reeds_path, 'inputs', 'fuelprices', 'gasreg_hdd.csv'),
+        index_col=0,
+    )
+    ddc = pd.read_csv(
+        os.path.join(reeds_path, 'inputs', 'fuelprices', 'gasreg_cdd.csv'),
+        index_col=0,
+    )
```
kodiobika: nit: Suggestion to make the variable names a bit clearer
```suggestion
    gasreg_hdd = pd.read_csv(
        os.path.join(reeds_path, 'inputs', 'fuelprices', 'gasreg_hdd.csv'),
        index_col=0,
    )
    gasreg_cdd = pd.read_csv(
        os.path.join(reeds_path, 'inputs', 'fuelprices', 'gasreg_cdd.csv'),
        index_col=0,
    )
```

input_processing/fuelcostprep.py
```diff
@@ -138,6 +140,296 @@
...
+
+        hdd_t['t'] = t
+        cdd_t['t'] = t
+        hdd_t['ddtype'] = 'hdd'
+        cdd_t['ddtype'] = 'cdd'
+
+        out.append(hdd_t.reset_index().rename(columns={'index': 'date'}))
+        out.append(cdd_t.reset_index().rename(columns={'index': 'date'}))
+
+    daily_dds = pd.concat(out, ignore_index=True)
```
kodiobika: For simplicity, maybe something like this (it would also be ideal to make it a function, to avoid duplicate code for both HDD and CDD)

```suggestion
    hdd_daily_gasreg = (
        pd.merge(
            hdd_shape.reset_index(),
            gasreg_hdd.reset_index(),
            how='cross',
            suffixes=('_shape', '_magnitude')
        )
        .set_index(['t', 'timestamp'])
        .rename_axis(['year', 'datetime'])
        .sort_index()
    )
    for gasreg in gasreg_hdd.columns:
        hdd_daily_gasreg[gasreg] = (
            hdd_daily_gasreg[f"{gasreg}_shape"]
            * hdd_daily_gasreg[f"{gasreg}_magnitude"]
        )
        hdd_daily_gasreg = hdd_daily_gasreg.drop(
            columns=[f"{gasreg}_shape", f"{gasreg}_magnitude"]
        )
```

input_processing/fuelcostprep.py
kodiobika: Could we move this function to the top (after all the import statements) to distinguish it from the "main" code in this script?

input_processing/fuelcostprep.py
```diff
@@ -138,6 +140,296 @@
...
+    )
+    ddc = pd.read_csv(
+        os.path.join(reeds_path, 'inputs', 'fuelprices', 'gasreg_cdd.csv'),
+        index_col=0,
+    )
+    ddh.index = ddh.index.astype(int)
+    ddc.index = ddc.index.astype(int)
+
+    ddh = ddh.loc[ddh.index.intersection(model_years)].copy()
+    ddc = ddc.loc[ddc.index.intersection(model_years)].copy()
```
kodiobika: nit: If we move this further down (i.e., after `historical_hdd_daily_gasreg` and `historical_cdd_daily_gasreg` are defined -- see other comment) and filter on the columns upfront, we can avoid having to keep track of the valid gasregs (so `active_gasregs` can be deleted)
```suggestion
    ddh = ddh.loc[ddh.index.intersection(model_years)][historical_hdd_daily_gasreg.columns].copy()
    ddc = ddc.loc[ddc.index.intersection(model_years)][historical_cdd_daily_gasreg.columns].copy()
```

input_processing/fuelcostprep.py
```diff
@@ -138,6 +140,296 @@
...
+
+# Regression parameters for calculating natural gas price differences across regions based on degree days
+params = pd.read_csv(os.path.join(reeds_path,'inputs', 'fuelprices', 'temperature_price_regression_parameters.csv'), index_col='param')
+
+# Daily degree days by price region
+daily_dd = get_degree_days(inputs_case, hourly_formula=False)
+
+#apply the regional regression params to get daily multiplicative price differences from the annual price 
+daily_dd['date'] = pd.to_datetime(daily_dd['date'])
+daily_dd['month'] = daily_dd['date'].dt.month
```
kodiobika: To avoid having to create the `month_map` below, we can also reformat the "month" upfront to be in the "JAN", "FEB", etc. format
```suggestion
daily_dd['month'] = daily_dd['date'].dt.strftime('%b').str.upper()
```

input_processing/fuelcostprep.py
```diff
@@ -138,6 +140,296 @@
...
+    month_map = {
+        i: params.loc[f'alpha_{m}', r]
+        for i, m in enumerate(
+            ['JAN','FEB','MAR','APR','MAY','JUN','JUL','AUG','SEP','OCT','NOV','DEC'],
+            start=1
+        )
+    }
+
+    #Align index of month effects with index of hdd/cdd
+    month_effect = hdd.index.month.map(month_map)
```
kodiobika: Follow-up to suggestion in above comment
```suggestion
    # monthly effects
    month_effects_map = params.loc[params.index.str.contains('alpha_')][r]
    month_effects_map.index = month_effects_map.index.str.removeprefix('alpha_')
    month_effect = hdd.month.map(month_effects_map)
```

input_processing/fuelcostprep.py
```diff
@@ -138,6 +140,296 @@
...
+out_datetime = out.index
+if out_datetime.tz is None:
+    out_datetime = out_datetime.tz_localize('Etc/GMT+6')
+
+out = out.astype(np.float32)
+out.index = pd.MultiIndex.from_arrays(
+    [out_year.astype(np.int32), out_datetime],
+    names=['year', 'datetime'],
+)
+out = out.sort_index()
```
kodiobika: With the formatting changes from the other comments, I think these lines can be removed
```suggestion
```

input_processing/fuelcostprep.py
kodiobika: Since this function is doing a couple of different things, can we split its functionalities into multiple smaller functions (e.g., a `get_state_weights` function which calculates the weights from the population data, a `calculate_daily_state_degree_days` function which calculates the daily state-level degree days from the hourly temperature data), and then call those smaller functions in this one? And I know I initially suggested `get_degree_days` but now that I'm looking closer at this function, maybe it can be renamed to `calculate_population_weighted_daily_degree_days` or something similar

input_processing/fuelcostprep.py
```diff
@@ -138,6 +140,296 @@
...
+        {
+            st: cdd_daily_st[st] * state_weights[st]
+            for st in cdd_daily_st.columns
+            if st in state_weights
+        },
+        axis=1,
+    )
+    cdd_daily_reg.columns = [state_gasreg[st] for st in cdd_daily_reg.columns]
+    cdd_daily_reg = cdd_daily_reg.groupby(level=0, axis=1).sum()
+    cdd_daily_reg = cdd_daily_reg[[c for c in cdd_daily_reg.columns if c in active_gasregs]]
```
kodiobika: nit: Maybe something like this to cut down on the number of lines / make it simpler? And to avoid duplicate code, could we make this a function that takes a state daily profile (hdd_daily_st / cdd_daily_st) and state weights (pop_state['weight'] in this case)?

```suggestion
    historical_hdd_daily_gasreg = (
        (hdd_daily_st * pop_state['weight'])
        .transpose()
        .rename(st2gasreg)
        .groupby(level=0)
        .sum()
        .transpose()
    )
```

kodiobika (changes requested):
Thanks for your work on this! It's looking great so far -- mostly just some stylistic suggestions

input_processing/fuelcostprep.py
```diff
@@ -138,6 +140,296 @@
...
+    else:
+        temp_daily_st = temp.resample('D').agg(['min', 'max'])
+
+        tavg_daily_st = (
+            temp_daily_st.xs('min', axis=1, level=1)
+            + temp_daily_st.xs('max', axis=1, level=1)
+        ) / 2
+
+        hdd_daily_st = (base_temp_c - tavg_daily_st).clip(lower=0)
+        cdd_daily_st = (tavg_daily_st - base_temp_c).clip(lower=0)
```
kodiobika: Nice! I wasn't really familiar with `.resample()` and `.xs()` before now, and your use of them here is really elegant. Just some small changes to account for the renamed base temperature variable (see other comment) and a suggestion to make the daily average temperature variable slightly clearer
```suggestion
    if hourly_formula:
        hdd = (base_temp_celsius - temp).clip(lower=0)
        cdd = (temp - base_temp_celsius).clip(lower=0)

        hdd_daily_st = hdd.resample('D').sum()
        cdd_daily_st = cdd.resample('D').sum()
    else:
        temp_daily_st = temp.resample('D').agg(['min', 'max'])

        avg_temp_daily_st = (
            temp_daily_st.xs('min', axis=1, level=1)
            + temp_daily_st.xs('max', axis=1, level=1)
        ) / 2

        hdd_daily_st = (base_temp_celsius - avg_temp_daily_st).clip(lower=0)
        cdd_daily_st = (avg_temp_daily_st - base_temp_celsius).clip(lower=0)
```

cases.csv
```diff
@@ -143,6 +143,7 @@ GSw_GasCurve,"Select natural gas supply curve (0 = cendiv, 1 = national + cendiv
 GSw_GasCC_H_1x1,Turn on/off both gas-CC_H_1x1 and gas-CC_H_1x1-CCS_mod/max plant options,0; 1,0,
 GSw_GasCC_H_2x1,Turn on/off both gas-CC_H_2x1 and gas-CC_H_2x1-CCS_mod/max plant options,0; 1,0,
 GSw_GasCT_Aero,Turn on/off gas-CT_aeroderivative technology option,0; 1,0,
+GSw_GasPriceAdjMethod,"Select method for adjusting gas prices (0 = national wintertime markup, 1 = daily adjustments based on regional temperatures)",0; 1,0,
```
wesleyjcole: Given that "0" is used for off, it makes sense to me that these options would be 1 and 2 rather than 0.  0 to me implies no adjustment at all (flat price for the full year).

reeds/core/terminus/report.gms
```diff
@@ -529,19 +529,19 @@ repgasquant_nat(t)$tmodel_new(t) = sum{cendiv, repgasquant(cendiv,t) } ;
 *for reported gasprice (not that used to compute system costs)
 *scale back to $ / mmbtu
 repgasprice(cendiv,t)$[(Sw_GasCurve = 0)$tmodel_new(t)$repgasquant(cendiv,t)] =
-    smax{gb$[sum{h, GASUSED.l(cendiv,gb,h,t) }], gasprice(cendiv,gb,t) } / gas_scale ;
+    smax{gb$[sum{h, GASUSED.l(cendiv,gb,h,t) * gasprice_adj_cendiv(cendiv,h,t) }], gasprice(cendiv,gb,t) } / gas_scale ;
 
 repgasprice(cendiv,t)$[(Sw_GasCurve = 2)$tmodel_new(t)$repgasquant(cendiv,t)] =
     sum{(i,v,r,h)$[r_cendiv(r,cendiv)$valgen(i,v,r,t)$gas(i)$heat_rate(i,v,r,t)],
-          hours(h)*heat_rate(i,v,r,t)*fuel_price(i,r,t)*GEN.l(i,v,r,h,t)
+          hours(h)*heat_rate(i,v,r,t)*fuel_price(i,r,t)*GEN.l(i,v,r,h,t)*gasprice_adj_r(r,h,t)
```
wesleyjcole:
```suggestion
          hours(h) * heat_rate(i,v,r,t) * fuel_price(i,r,t) * GEN.l(i,v,r,h,t) * gasprice_adj_r(r,h,t)
```

docs/source/model_documentation.md
```diff
@@ -3609,6 +3628,23 @@ and $\rho$ and $\sigma$ are the seasonal multipliers for winter and nonwinter, r
...
+Otherwise, annual degree day projections are calculated by taking historical (1995-2025) state-level degree days from {cite}`noaaDailyDegreeDays`, projecting them out to 2050 using a 30-year linear trend, and then aggregating them to the scope of the regression regions via population-weighted average.
+For purposes of calculating this population-weighted average, state-level population projections for 2030, 2040, and 2050 are taken from {cite}`uvaWeldonCooperCenterPopulationProjections` and in-between years are linearly interpolated.
+
+Depending on the spatial resolution of the gas prices being used in the model, the daily gas price adjustments are either downscaled to the zone level by copying each regression region's adjustments to their constituent zones or upscaled to the census division level via population-weighted average.
+Once representative periods are selected in the model, the daily adjustments are filtered to include only the representative periods and then renormalized so that the average price multiplier for each zone or census division and each model year is one, thus ensuring the year-round average gas price remains unchanged.
```
wesleyjcole: Would you include an example natural gas price adjustment figure (for example, one from your PR text)?  I think that would be really helpful for the user to understand what is going on.

docs/sources.csv
```diff
@@ -263,6 +263,8 @@ RelativeFilePath,RelativeFolderPath,FileName_new,FileExtension,Description_new,I
 /inputs/fuelprices/coal_AEO_2026_altelec.csv,/inputs/fuelprices,coal_AEO_2026_altelec,.csv,AEO2026 Alternative Electricity case census division fuel price of coal with missing New England values set to Mid Atlantic,"t,cendiv",2025,AEO2026,Input,$/MMBtu
 /inputs/fuelprices/coal_AEO_2026_baseline.csv,/inputs/fuelprices,coal_AEO_2026_baseline,.csv,AEO2026 Counterfactual Baseline case census division fuel price of coal with missing values forward-filled from earlier years and missing New England values set to Mid Atlantic,"t,cendiv",2025,AEO2026,Input,$/MMBtu
 /inputs/fuelprices/dollaryear.csv,/inputs/fuelprices,dollaryear,.csv,Dollar year mapping for each fuel price scenario,,,,,
+/inputs/fuelprices/gasreg_degree_day_price_mult_regression_params.csv,/inputs/fuelprices,gasreg_degree_day_price_mult_regression_params,.csv,"Coefficients and intercepts derived from regression regressing daily gasreg heating/cooling degree days on daily deviations of gas prices from their annual average price with monthly fixed effects. ""Beta"" values are HDD/CDD coefficients and ""alpha"" values are intercepts and monthly fixed effects.",param,,,Input,
```
wesleyjcole: Would you point to the relevant folder in the Input_Processing repo?

reeds/core/solve/2_temporal_params.gms
```diff
@@ -879,15 +879,42 @@ peakdem_static_h(r,h,t) = peak_h(r,h,t) * (1 - sum{flex_type, flex_demand_frac(f
...
 szn_adj_gas(h)$frac_h_quarter_weights(h,"wint") =
     szn_adj_gas(h) + frac_h_quarter_weights(h,"wint") * szn_adj_gas_winter ;
+* Renormalize so hour-weighted average of szn_adj_gas is 1
+scalar szn_adj_gas_avg "--unitless-- hour-weighted average of natural gas seasonal adjustment" ;
+szn_adj_gas_avg = sum{h, szn_adj_gas(h) * hours(h) } / sum{h, hours(h) } ;
+szn_adj_gas(h) = szn_adj_gas(h) / szn_adj_gas_avg ;
+
+* If GSw_GasPriceAdjMethod = 0, replace daily regional adjustments with the national wintertime markup
+gasprice_adj_r(r,h,t)$(Sw_GasPriceAdjMethod = 0) = szn_adj_gas(h) ;
+gasprice_adj_cendiv(cendiv,h,t)$(Sw_GasPriceAdjMethod = 0) = szn_adj_gas(h) ;
```
wesleyjcole: If you want to add a "no adjustment" for a zero input you could do:

`gasprice_adj_r(r,h,t)$(Sw_GasPriceAdjMethod = 0) = 1 ;`

reeds/core/solve/6_data_dump.gms
```diff
@@ -194,19 +194,19 @@ repgasquant(cendiv,t)$[(Sw_GasCurve = 1 or Sw_GasCurve = 2)$tcur(t)] =
        } ;
 
 repgasprice(cendiv,t)$[(Sw_GasCurve = 0)$tcur(t)] =
-    smax{gb$[sum{h, GASUSED.l(cendiv,gb,h,t) }], gasprice(cendiv,gb,t) } ;
+    smax{gb$[sum{h, GASUSED.l(cendiv,gb,h,t) * gasprice_adj_cendiv(cendiv,h,t) }], gasprice(cendiv,gb,t) } ;
 
 repgasprice(cendiv,t)$[(Sw_GasCurve = 2)$tcur(t)$repgasquant(cendiv,t)] =
     sum{(i,v,r,h)$[r_cendiv(r,cendiv)$valgen(i,v,r,t)$gas(i)$heat_rate(i,v,r,t)],
-          hours(h)*heat_rate(i,v,r,t)*fuel_price(i,r,t)*GEN.l(i,v,r,h,t)
+          hours(h)*heat_rate(i,v,r,t)*fuel_price(i,r,t)*GEN.l(i,v,r,h,t)*gasprice_adj_r(r,h,t)
```
wesleyjcole:
```suggestion
          hours(h) * heat_rate(i,v,r,t) * fuel_price(i,r,t) * GEN.l(i,v,r,h,t) * gasprice_adj_r(r,h,t)
```

reeds/input_processing/fuelcostprep.py
```diff
@@ -13,13 +13,410 @@
...
+
+    # Create another set of multipliers for census divisions by aggregating
+    # the gasreg-level multipliers via population-weighted average
+    gasreg_cendiv_weights = calculate_region_aggregion_population_weights(
+        inputs_case,
+        region_level='gasreg',
+        aggregion_level='cendiv'
+    )
+    gasreg_cendiv_map = dict(zip(hierarchy['gasreg'], hierarchy['cendiv']))
+    df_out_cendiv = aggregate_regional_degree_days_to_aggregion(
```
wesleyjcole: This is not doing a degree day aggregation.  It works, but maybe it's worth generalizing the function name to reflect that it isn't only used on degree days.

wesleyjcole (approved):
I noticed that the z error_check went down when you turned on the daily natural gas price adjustments with both `GSw_GasCurve = 1` and `GSw_GasCurve = 2`. I couldn't see any reasons why that would change.  Any thoughts as to why that might be?  It moved in a good direction, so if you don't have ideas for why it changed, that's fine.

reeds/core/solve/2_temporal_params.gms
kodiobika: thanks!

inputs/zones/state_groups.csv
patrickbrown4, line 1: Could you add a bullet to this section in `inputs/zones/README.md` with a link to the source used for the new `gasreg`? https://github.com/ReEDS-Model/ReEDS/blob/dce4e296f146377c718258d623a62e0043e13344/inputs/zones/README.md?plain=1#L79-L83

kodiobika: There's actually not a single source for these; they were inspired partly by the census divisions, the EIA NEMS natural gas flow/capacity regions, and the EIA NEMS natural gas-electricity market module regions (figs 2.5 and 2.7 of https://www.eia.gov/outlooks/aeo/nems/documentation/ngmm/pdf/NGMM_AEO2025.pdf), but don't follow any of them strictly. Should I just note that in the readme?

patrickbrown4: Yeah I think it'd be helpful to note all of that, including links to each of the sources (ok to reproduce the [Census divisions](https://www2.census.gov/geo/pdfs/maps-data/maps/reference/us_regdiv.pdf) link from above). Thanks!

kodiobika:
> I noticed that the z error_check went down when you turned on the daily natural gas price adjustments with both `GSw_GasCurve = 1` and `GSw_GasCurve = 2`. I couldn't see any reasons why that would change. Any thoughts as to why that might be? It moved in a good direction, so if you don't have ideas for why it changed, that's fine.

Weird, yeah I'm not sure why it changed

kodiobika (dismissed):

docs/source/model_documentation.md
```diff
@@ -1650,8 +1661,16 @@ an exponential decay length of 150 km is used by default.
...
-The natural gas fuel prices also include a seasonal price adjustor, making winter prices higher than the natural gas prices seen during the other seasons of the year.
+The natural gas fuel prices also include time-based price adjustors.
+One option is a seasonal price adjustor, which makes winter prices higher than the natural gas prices seen during the other seasons of the year CONUS-wide.
 For details, see the [Seasonal Natural Gas Price Adjustments section](#seasonal-natural-gas-price-adjustments) of the appendix.
+The other option is a daily price adjustor, which adjusts prices in accordance with regional temperatures using coefficients developed through a linear regression analysis regressing daily heating and cooling degree days on daily deviations of natural gas spot prices from their annual average price.
+For details, see the [Daily Natural Gas Price Adjustments section](#daily-natural-gas-price-adjustments) of the appendix.
+
+```{admonition} Natural gas price adjustments
+The switch `GSw_GasPriceAdjMethod` controls the choice of natural gas price adjustments.
+0 = no adjustment, 1 = national wintertime markup, 2 = daily adjustments based on regional temperatures
```
patrickbrown4: Can you note the default setting?

docs/source/model_documentation.md
```diff
@@ -3613,6 +3632,31 @@ and $\rho$ and $\sigma$ are the seasonal multipliers for winter and nonwinter, r
...
+The Pacific census division is broken up into the subregions "Northwest" (Oregon and Washington) and "California" (California).
+The Mountain census division is broken up into the subregions "Southwest" (Arizona and New Mexico) and "Mountain" (all remaining states in the Mountain census division).
+
+To derive daily gas price adjustments, the regression parameters are applied to projections of daily heating and cooling degree days.
+These projections are derived by rescaling historical daily heating and cooling degree days (calculated using hourly average temperatures observed during the weather years corresponding to representative periods) to match projections of annual degree days.
+In cases where the regression regions correspond to census divisions, annual degree day projections are taken from AEO.
+Otherwise, annual degree day projections are calculated by taking historical (1995-2025) state-level degree days from {cite}`noaaDailyDegreeDays`, projecting them out to 2050 using a 30-year linear trend, and then aggregating them to the scope of the regression regions via population-weighted average.
```
patrickbrown4: Have you shown that these two methods match for the cendiv regions? I worry a bit about using different methods for different regions - personally my preference would be to either use `cendiv` (if that's the resolution at which you have the AEO data) or use the single NOAA method everywhere.

docs/source/model_documentation.md
```diff
@@ -3613,6 +3632,31 @@ and $\rho$ and $\sigma$ are the seasonal multipliers for winter and nonwinter, r
...
 
+### Daily Natural Gas Price Adjustments
+
+Daily gas price adjustments use degree day coefficients derived from an ordinary least squares regression regressing daily regional heating and cooling degree days on daily deviations of regional natural gas spot prices from their annual average price with monthly fixed effects.
+The regions used in the regression mostly correspond to census divisions, except in two cases where census divisions are broken up into two smaller regions.
+The Pacific census division is broken up into the subregions "Northwest" (Oregon and Washington) and "California" (California).
+The Mountain census division is broken up into the subregions "Southwest" (Arizona and New Mexico) and "Mountain" (all remaining states in the Mountain census division).
+
+To derive daily gas price adjustments, the regression parameters are applied to projections of daily heating and cooling degree days.
+These projections are derived by rescaling historical daily heating and cooling degree days (calculated using hourly average temperatures observed during the weather years corresponding to representative periods) to match projections of annual degree days.
```
patrickbrown4: To make sure I understand, is the projection step to capture climate change? Is a particular climate scenario assumed (if so, it'd be good to note it)?

docs/source/model_documentation.md
```diff
@@ -3613,6 +3632,31 @@ and $\rho$ and $\sigma$ are the seasonal multipliers for winter and nonwinter, r
...
+In cases where the regression regions correspond to census divisions, annual degree day projections are taken from AEO.
+Otherwise, annual degree day projections are calculated by taking historical (1995-2025) state-level degree days from {cite}`noaaDailyDegreeDays`, projecting them out to 2050 using a 30-year linear trend, and then aggregating them to the scope of the regression regions via population-weighted average.
+For purposes of calculating this population-weighted average, state-level population projections for 2030, 2040, and 2050 are taken from {cite}`uvaWeldonCooperCenterPopulationProjections` and in-between years are linearly interpolated.
+
+Depending on the spatial resolution of the gas prices being used in the model, the daily gas price adjustments are either downscaled to the zone level by copying each regression region's adjustments to their constituent zones or upscaled to the census division level via population-weighted average.
```
patrickbrown4, line 3643: It might be helpful to note the default treatment (the one used in the USA_defaults case)

cases.csv
```diff
@@ -143,6 +143,7 @@ GSw_GasCurve,"Select natural gas supply curve (0 = cendiv, 1 = national + cendiv
 GSw_GasCC_H_1x1,Turn on/off both gas-CC_H_1x1 and gas-CC_H_1x1-CCS_mod/max plant options,0; 1,0,
 GSw_GasCC_H_2x1,Turn on/off both gas-CC_H_2x1 and gas-CC_H_2x1-CCS_mod/max plant options,0; 1,0,
 GSw_GasCT_Aero,Turn on/off gas-CT_aeroderivative technology option,0; 1,0,
+GSw_GasPriceAdjMethod,"Select method for adjusting gas prices (0 = no adjustment, 1 = national wintertime markup, 2 = daily adjustments based on regional temperatures)",0; 1; 2,1,
```
patrickbrown4, line 146: Do you think it'd be worthwhile to add a test case for the new daily adjustment method to `cases_test.csv`? If there's a lot of processing that only runs when `GSw_GasPriceAdjMethod = 2`, having a test case might help to keep it working. If the differences are relatively minor and/or you don't think the feature will be used often, it's ok to leave it out.

tests/objective_function_params.yaml
patrickbrown4: `szn_adj_gas` is no longer used in the objective function, so I think it'd be best to replace this block with a check on the parameters that do get used. It looks like that's:
- `gasprice_adj_r`, I think for all switch values?
- `gasprice_adj_cendiv` if `GSw_GasCurve` is 0 or 3

reeds/core/setup/d_objective.gms
```diff
@@ -240,27 +240,27 @@ eq_Objfn_op(t)$tmodel(t)..
...
                    * gasprice(cendiv,gb,t)
                    }$(Sw_GasCurve = 0)
 
 *Sw_GasCurve = 3 (national supply curve for natural gas prices with census division multipliers)
               + sum{(h,cendiv,gb), hours(h) * GASUSED(cendiv,gb,h,t)
-                   * gasadder_cd(cendiv,t,h) + gasprice_nat_bin(gb,t)
+                   * gasadder_cd(cendiv,t,h) * gasprice_adj_cendiv(cendiv,h,t) + gasprice_nat_bin(gb,t)
                    }$(Sw_GasCurve = 3)
 
 *Sw_GasCurve = 1 (national and census division supply curves for natural gas prices)
```
patrickbrown4, line 259: @wesleyjcole The comment says `Sw_GasCurve = 1 ` but I don't see a `Sw_GasCurve` conditional - am I missing something or are we applying this twice?

reeds/input_processing/fuelcostprep.py
```diff
@@ -21,6 +22,402 @@
 from pathlib import Path
 sys.path.append(str(Path(__file__).parent.parent.parent))
 import reeds
+
+
+def calculate_region_aggregion_population_weights(
+    inputs_case: str,
```
patrickbrown4: (nitpick) In general it'd be nice to allow Paths as well (I haven't worked through all the old functions yet but it'd be nice to at least start doing it for new functions)
```suggestion
    inputs_case: str | Path,
```

reeds/input_processing/fuelcostprep.py
```diff
@@ -13,6 +13,7 @@
 ### ===========================================================================
 
 import pandas as pd
+import numpy as np
```
patrickbrown4: Already imported below
```suggestion
```

reeds/input_processing/fuelcostprep.py
```diff
@@ -21,6 +22,402 @@
...
+    # Get baseline temperature for calculating degree days
+    scalars = reeds.io.get_scalars(inputs_case)
+    base_temp = scalars['degree_days_base_temperature']
+
+    # Calculate each state's average temperature for each day
+    temp_daily = temp_hourly.resample('D').agg(['min', 'max'])
+    avg_temp_daily = (
+        temp_daily.xs('min', axis=1, level=1)
+        + temp_daily.xs('max', axis=1, level=1)
+    ) / 2
```
patrickbrown4: The average of the min and max is not the same as the average over all hours. If you actually want the average, I would just add 'mean' to the list of agg funcs.

It looks like the max and min aren't used though, so you can just take the simple mean:

```suggestion
    # Calculate each state's average temperature for each day
    avg_temp_daily = temp_hourly.resample('D').mean()
```

reeds/input_processing/hourly_writetimeseries.py
```diff
@@ -1313,6 +1364,35 @@ def main(sw, reeds_path, inputs_case, periodtype='rep', make_plots=1, logging=Tr
...
+            .rename('multiplier')
+        )
+        # Renormalize so the average for each region and year is 1,
+        # ensuring the year-round average gas price doesn't change.
+        df = (
+            df.div(df.groupby(level=[region_level, 't']).mean())
+            .reset_index()
+            [[region_level, 'h', 't', 'multiplier']]
+        )
+        daily_gasprice_multipliers_dict[region_level] = df
```
patrickbrown4: If you use multiple representative years (e.g. GSw_HourlyWeatherYears=2012_2013), it looks like you'll get different average prices in the different weather years (which is what I would want to happen, since it does actually happen) - is that right?

reeds/input_processing/fuelcostprep.py
```diff
@@ -21,6 +22,402 @@
...
+        inputs_case: Path to the inputs case directory.
+
+    Returns:
+        (pd.DataFrame, pd.DataFrame)
+    """
+    # Get hourly state-level temperatures for the given weather year(s)
+    sw = reeds.io.get_switches(inputs_case)
+    weather_years = [int(y) for y in sw.GSw_HourlyWeatherYears.split('_')]
+    temp_hourly = reeds.io.get_temperatures(inputs_case)
+    temp_hourly = temp_hourly.loc[temp_hourly.index.year.isin(weather_years)]
```
patrickbrown4: I checked the versions of `daily_gasprice_multipliers_cendiv.csv` and `daily_gasprice_multipliers_r.csv` saved to the `inputs_case/stress{year}i0` folders, and they're all empty. I think that means we multiply gas prices by zero during stress periods, which will mess up the dispatch order during stress periods.

So I think you want `resource_adequacy_years` here instead of `GSw_HourlyWeatherYears`. I'm not sure if the downstream code in `hourly_writetimeseries.py` will also need to be adjusted.

reeds/core/setup/d_objective.gms
wesleyjcole: It's just a longer block, so the conditional isn't for several lines down.  You'll see a `$[Sw_GasCurve = 1]` at the end of the block.

patrickbrown4: Ah got it, I missed the bracket, thanks.

reeds/input_processing/fuelcostprep.py
kodiobika: We initially wrote it this way to mimic the traditional method for calculating degree days (https://www.eia.gov/energyexplained/units-and-calculators/degree-days.php), but yeah I think it'd be better to just take the average since we have hourly data. Will update that

patrickbrown4: Ah I see, if you want to keep this method then you could instead just adjust the comment and variable names so they match the code, up to you

reeds/input_processing/hourly_writetimeseries.py
kodiobika: Yep that's right - the average price within a model year will be the same, but the average price across different weather years can differ

docs/source/model_documentation.md
kodiobika: They differ slightly due to AEO using different (paywalled) population projections and AEO's version of the Pacific `cendiv` including AK and HI. But good point, just using one method would be better for consistency and easier to explain - I'll update that to use just the NOAA method

docs/source/model_documentation.md
kodiobika: That's right, specifically we wanted our HDD/CDD baseline for each year to match what's in the AEO. So I guess the scenario would just be AEO 2026. I'll update the documentation to note that that's where the projection method comes from

reeds/core/solve/6_data_dump.gms
```diff
@@ -194,19 +194,19 @@ repgasquant(cendiv,t)$[(Sw_GasCurve = 1 or Sw_GasCurve = 2)$tcur(t)] =
        } ;
 
 repgasprice(cendiv,t)$[(Sw_GasCurve = 0)$tcur(t)] =
-    smax{gb$[sum{h, GASUSED.l(cendiv,gb,h,t) }], gasprice(cendiv,gb,t) } ;
+    smax{gb$[sum{h, GASUSED.l(cendiv,gb,h,t) * gasprice_adj_cendiv(cendiv,h,t) }], gasprice(cendiv,gb,t) } ;
```
patrickbrown4: If I'm reading it right, it seems like `gasprice_adj_cendiv` won't have an effect here, since it's only used in the conditional on `gb`. Should it multiply `gasprice` instead?

Although actually, `repgasprice` is only used to calculate `repgasprice_filt`, which used to be used in Augur (probably for the storage arbitrage value calculation) but is no longer used. So this whole block should be able to be deleted. (Or can save for a later cleanup PR, either works.)

reeds/core/terminus/report.gms
```diff
@@ -529,19 +529,19 @@ repgasquant_nat(t)$tmodel_new(t) = sum{cendiv, repgasquant(cendiv,t) } ;
 *for reported gasprice (not that used to compute system costs)
 *scale back to $ / mmbtu
 repgasprice(cendiv,t)$[(Sw_GasCurve = 0)$tmodel_new(t)$repgasquant(cendiv,t)] =
-    smax{gb$[sum{h, GASUSED.l(cendiv,gb,h,t) }], gasprice(cendiv,gb,t) } / gas_scale ;
+    smax{gb$[sum{h, GASUSED.l(cendiv,gb,h,t) * gasprice_adj_cendiv(cendiv,h,t) }], gasprice(cendiv,gb,t) } / gas_scale ;
```
patrickbrown4: Just to double check (I'm not sure I'm interpreting correctly as I don't think I've dealt with this equation before), I don't think I follow how `gasprice_adj_cendiv` is having an effect here, as it shouldn't affect the `gb` conditional. Should there be a `GASUSED.l * gasprice_adj_cendiv * gasprice` term?

cases.csv
kodiobika: The daily adjustment processing gets run in every case and then the multipliers are overwritten if `GSw_GasPriceAdjMethod != 2`, so I think we should be okay

reeds/core/terminus/report.gms
kodiobika: Ah yeah sorry, I think I was just looking for all gas price-related terms that summed over `h` and added the adjustment term there without thinking. Looking at it carefully now I think what this statement is doing is finding the max `gasprice` among gas bins where the total amount of gas used across all hours is non-zero? So I think that `sum` statement should remain unchanged (from `main`) and instead the `gasprice` term should be replaced by ~`gasprice(cendiv,gb,t) * sum{h, gasprice_adj_cendiv(cendiv,h,t)}`~. Does that sound right?

kodiobika: Oh actually we don't want to multiply by the sum there, I think we'd want to find the hour that maximizes the gas price, so instead it'd be `gasprice(cendiv,gb,t) * smax{h, gasprice_adj_cendiv(cendiv,h,t)}`?

kodiobika: But then again, I guess we'd only want to consider hours where `GASUSED.l` is non-zero (assuming my interpretation is right). And actually I'm not sure if we want `repgasprice` to include the adjustment terms if it's just supposed to be "representative", so maybe it should just be reverted to what's in `main`. @wesleyjcole Any opinion on this one?

wesleyjcole: Patrick - thanks for catching this one.

Kodi - I think we need to calculate an annual `gasprice_adj_cendiv` value, weighted by `sum{gb, GASUSED.l(cendiv,gb,h,t) }`.  This would give a weighted average price adjustment based on when gased in used.  I expect that it will be close to one, given that the adjustment was built to not change the annual input price.

docs/source/model_documentation.md
kodiobika: Although now that I think about it, we initially wanted to make sure the HDD/CDD baselines match because we were regressing daily HDD/CDDs on daily prices, and the projected HDD/CDDs are baked into the AEO prices. But now we're regressing daily HDD/CDDs on the deviation of daily prices from an annual average (i.e., on daily multipliers), so I think we can just remove the projections entirely and just apply the regression coefficients to the historical (weather year) daily degree days. @wesleyjcole Does this sound right to you?

kodiobika: Actually maybe not, thinking about it more - if 2050 is a hotter year for a region than 2025, we would want its price multipliers to be higher on average in 2050, even if the hourly weather patterns are the same right?

kodiobika: Thinking about it more (again), if the HDD/CDD projections are already baked into the year-round average AEO prices, then actually we might be double-counting the hotness or coldness of the model years (maybe the normalization of the multipliers makes it all moot anyway though?). So I think we should just be using the historical degree days

wesleyjcole: Thanks for putting your thinking in this thread.  This is complicated, and I think I can convince myself that either approach is reasonable.  In the end, I don't think it will matter that much, and it would be much simpler (and easier to maintain) if we didn't use the projections of the HDD/CDD, so I'm for taking your suggestion and just use the historical degree days.

patrickbrown4: Yeah, I'm not sure I follow everything, but that sounds most in keeping with the current approach of using historical weather for wind/solar/load/outages. Eventually it'd be great to have a unified approach for incorporating defined climate projections into all of these, so I like the idea of keeping the gas price treatment more straightforwardly historical for now, and then thinking more about how to update them all together.

kodiobika:
@patrickbrown4 @wesleyjcole I'm in the middle of creating new comparison reports, but the code is ready for re-review whenever you get a chance, thanks!

kodiobika:
> @patrickbrown4 @wesleyjcole I'm in the middle of creating new comparison reports, but the code is ready for re-review whenever you get a chance, thanks!

Oop actually there's a bug in `report.gms` but everything else is ready for review

kodiobika:
> > @patrickbrown4 @wesleyjcole I'm in the middle of creating new comparison reports, but the code is ready for re-review whenever you get a chance, thanks!
> 
> Oop actually there's a bug in `report.gms` but everything else is ready for review

Should be resolved now

docs/source/figs/docs/spatial_layers_zones.png
patrickbrown4, line 1: I could remake this figure for the 90 zones, but we'll probably end up wanting to redo the layout (both of the figure and the discussion in the docs) to illustrate the different possible resolutions, so I think it can wait for a followup PR

docs/source/model_documentation.md
```diff
@@ -1650,8 +1661,16 @@ an exponential decay length of 150 km is used by default.
 If `GSw_GasRegionSmooth` is set to 0, the 1:1 zone:census-division mapping in {numref}`figure-hierarchy` is instead used directly.
 ```
 
-The natural gas fuel prices also include a seasonal price adjustor, making winter prices higher than the natural gas prices seen during the other seasons of the year.
+The natural gas fuel prices also include time-based price adjustors.
+One option is a seasonal price adjustor, which makes winter prices higher than the natural gas prices seen during the other seasons of the year CONUS-wide.
```
patrickbrown4:
```suggestion
The default option is a seasonal price adjustor, which makes winter prices higher than the natural gas prices seen during the other seasons of the year CONUS-wide.
```

reeds/core/terminus/report.gms
```diff
@@ -527,21 +527,25 @@ repgasquant_irt(i,r,t)$tmodel_new(t) =
 repgasquant_nat(t)$tmodel_new(t) = sum{cendiv, repgasquant(cendiv,t) } ;
 
 *for reported gasprice (not that used to compute system costs)
-*scale back to $ / mmbtu
+*scale back to $ / mmbtu and apply annual consumption-weighted gas price multipliers
+gas_consumption(cendiv,gb,t) = sum{h, GASUSED.l(cendiv,gb,h,t) } ;
```
patrickbrown4: Since `gas_scale` is used here, and from my read of `eq_gasused`, it looks like we need to multiply by `gas_scale` to get the units in MMBtu:
https://github.com/ReEDS-Model/ReEDS/blob/80d1ec04f89c505ed89126d71a5512785fefc9d5/reeds/core/terminus/report.gms#L506-L507

reeds/core/terminus/report_params.csv
```diff
@@ -77,6 +77,7 @@ error_check(*),unitless,set of checks to determine if there is an error - values
 "flex_load_out(flex_type,r,allh,t)",MWh,flexible load consumed in each timeslice,,,
 forced_outage(i),fraction,average forced outage rate (h-weighted; r-unweighted) by technology during representative periods,,,
 planned_outage(i),fraction,average scheduled outage rate (h-weighted) by technology during representative periods,,,
+"gas_consumption(cendiv,gb,t)",MMBtu,annual gas used by census division and gas bin,,,
```
patrickbrown4: See note on `report.gms` about double-checking the use of `gas_scale` to make sure the units are in MMBtu

tests/objective_function_params.yaml
patrickbrown4, line 1: Thanks for adding these!
I don't have a good system for remembering to do this. Maybe we should add another checkbox to "Details to double-check" in the PR template. (Not here, just thinking out loud.)

inputs/scalars.csv
```diff
@@ -23,12 +23,12 @@ csp_sm_2,2.4,"--fraction-- solar multiple for csp2"
 csp_sm_3,1.3,"--fraction-- solar multiple for csp3"
 csp_sm_4,1.0,"--fraction-- solar multiple for csp4"
 csp_sm_baseline,2.4,"--fraction-- baseline solar multiple to normalize other CSP solar multiples by"
+degree_days_base_temperature,18.33,"--temperature in degrees Celsius to use as the baseline temperature for calculating degree days"
```
patrickbrown4: Could add a note on the source:
```suggestion
degree_days_base_temperature,18.33,"--temperature in degrees Celsius to use as the baseline temperature for calculating degree days (65°F from https://www.eia.gov/energyexplained/units-and-calculators/degree-days.php)"
```

reeds/input_processing/fuelcostprep.py
```diff
@@ -86,6 +86,198 @@ def plot_cendivweights(inputs_case, dfmap, cendivweights):
...
+    # (https://www.eia.gov/energyexplained/units-and-calculators/degree-days.php),
+    # but we found that this approach generally resulted in better
+    # predictors of daily deviations from annual average gas prices.
+    hdd_hourly = (base_temp - temp_hourly).clip(lower=0)
+    hdd_daily = hdd_hourly.resample('D').mean()
+
+    cdd_hourly = (temp_hourly - base_temp).clip(lower=0)
+    cdd_daily = cdd_hourly.resample('D').mean()
+
+    return hdd_daily, cdd_daily
```
patrickbrown4: Slight preference to instead return a `{'hdd':hdd_daily, 'cdd':cdd_daily}` dict so the outputs are self-documenting and to help avoid mixups

reeds/core/terminus/report.gms
```diff
@@ -527,21 +527,25 @@ repgasquant_irt(i,r,t)$tmodel_new(t) =
...
 repgasprice(cendiv,t)$[(Sw_GasCurve = 0)$tmodel_new(t)$repgasquant(cendiv,t)] =
-    smax{gb$[sum{h, GASUSED.l(cendiv,gb,h,t) }], gasprice(cendiv,gb,t) } / gas_scale ;
+    smax{gb$[gas_consumption(cendiv,gb,t)],
+        gasprice(cendiv,gb,t)
+        * sum{h, gasprice_adj_cendiv(cendiv,h) * GASUSED.l(cendiv,gb,h,t) / gas_consumption(cendiv,gb,t) / hours(h) }
+        } / gas_scale ;
```
patrickbrown4: If `gas_scale` is added above, it'd be removed here

reeds/core/terminus/report.gms
```diff
@@ -527,21 +527,25 @@ repgasquant_irt(i,r,t)$tmodel_new(t) =
 repgasquant_nat(t)$tmodel_new(t) = sum{cendiv, repgasquant(cendiv,t) } ;
 
 *for reported gasprice (not that used to compute system costs)
-*scale back to $ / mmbtu
+*scale back to $ / mmbtu and apply annual consumption-weighted gas price multipliers
+gas_consumption(cendiv,gb,t) = sum{h, GASUSED.l(cendiv,gb,h,t) } ;
 repgasprice(cendiv,t)$[(Sw_GasCurve = 0)$tmodel_new(t)$repgasquant(cendiv,t)] =
```
patrickbrown4, line 534: Just to double check, do these all look right (expected order of magnitude) in the test runs you did? (No need to add a bunch of plots to the PR, a quick scan of the output file is enough)

I've added a reminder to #34 to add an output plot for gas price (eventually, not here)

patrickbrown4 (approved):
Thanks for the revisions! A couple small comments but otherwise looks good to me.

reeds/core/terminus/report.gms
kodiobika: Oh that's weird, I actually took the units from the description of `GASUSED` in `c_model.gms` (`MMBtu/hour`). I think this implies that the units of either `GASUSED` or `repgasquant` must be labelled incorrectly, right? Regarding `gas_consumption`, my understanding is the units end up not mattering (except for documentation purposes in `report_params.csv`) since we only use it to

- check if consumption is non-zero for a given gas bin (in the `gb$[gas_consumption(cendiv,gb,t)]` conditional) and 
- calculate consumption weights for each hour (`GASUSED.l(cendiv,gb,h,t) / gas_consumption(cendiv,gb,t)`)

Does that sound right to you? And assuming it's correct that the units of `gas_consumption` don't matter, I think we should end up with whatever units for `repgasprice` we had before (but will double-check), which is labelled as `$/mmbtu` but now I'm not sure what they actually are

patrickbrown4: It's probably safest to be sure about the units for `gas_consumption`, since if it's included as an output parameter (which it will be if it's defined in `report_params.csv`), someone will probably use it someday.

I missed it before but I think you're right that it's currently off; since `GASUSED` is in [MMBtu/hour], it should be multiplied by `hours` when taking the sum.

The `gas_scale` makes things weird; it's a 1e6 scaling factor but is only to keep the gas variables to a similar scale as the other model variables. So:

- `GASUSED` = `heat_rate` [MMBtu/MWh] * `GEN` [MW] / `gas_scale` [1e6] = 1e-6*[MMBtu/hour]
- `repgasquant` = `GASUSED` (1e-6*[MMBtu/hour]) * `hours` [hour] * `gas_scale` (1e6) * 1e-9 [quad/MMBtu] = [quad]
- `gas_consumption` = `GASUSED` (1e-6*[MMBtu/hour]) * `hours` [hour] = 1e-6*[MMBtu]

So if you add `* gas_scale` to the definition of `gas_consumption`, you'll get `gas_consumption` in [MMBtu], as stated in `report_params.csv`:
- `gas_consumption(cendiv,gb,t) = sum{h, GASUSED.l(cendiv,gb,h,t) * hours(h) * gas_scale } ;`
- [MMBtu] = 1e-6*[MMBtu/hour] * [hour] * 1e6 = [MMBtu]

For `repgasprice`, note that `gasprice` is [$/MMBtu] multiplied by `gas_scale` in `b_inputs.gms`, so its values are on the order of 1e6. So if you do it as:

- `repgasprice` = `gasprice` (1e6*[$/MMBtu]) * `GASUSED` (1e-6*[MMBtu/hour]) * `hours` [hour] / `gas_consumption` [MMBtu]

(after `gas_consumption` has been descaled to [MMBtu]), then you should just get [$/MMBtu], without the use of `gas_scale` in the `repgasprice` equation.

It does look like the `hours` is currently in the wrong place in the `repgasprice` equation; if `GASUSED` is [MMBtu/hour] and `gas_consumption` is [MMBtu] then it should be in the numerator instead of the denominator.

At least that's my current take; should take another look on Monday!

kodiobika: Thanks! Appreciate the detailed explanation; I think I understand now. I'll make that correction to `gas_consumption`, and I didn't realize `repgasquant` and `gas_consumption` were so similar (only differing in the indices they sum over), so I think I'll rename `gas_consumption` to `repgasquant_gb` so that we have `repgasquant_gb(cendiv,gb,t) = sum{h, GASUSED.l(cendiv,gb,h,t) * hours(h) } * gas_scale ;`

Given that, where I've landed for the `repgasprice` calculation is `gasprice(cendiv,gb,t) * sum{h, gasprice_adj_cendiv(cendiv,h) * GASUSED.l(cendiv,gb,h,t) / repgasquant_gb(cendiv,gb,t) };`

To explain, my understanding is:
- `sum{h, gasprice_adj_cendiv(cendiv,h) / hours(h)}` represents the simple average gas price adjustment
- To make it consumption weighted, we multiply the `gasprice_adj_cendiv(cendiv,h)` term by `GASUSED.l(cendiv,gb,h,t) * hours(h) / repgasquant_gb(cendiv,gb,t)`.
- The `hours(h)` terms in the numerator and denominator cancel out
- Like you said, because `repgasquant_gb` is scaled with `gas_scale`, we can remove the existing `gas_scale` term from the denominator

What do you think?

kodiobika: Actually I'm not sure about that simple average calculation, going to think about it a bit more

kodiobika: I think it would actually be `sum{h, gasprice_adj_cendiv(cendiv,h) * hours(h)} / sum{h, hours(h)}`; I'll take another look at the `repgasprice` calculation

kodiobika: Nevermind I think I was overthinking it; if I'm understanding correctly now, the simple average shouldn't matter and `gasprice(cendiv,gb,t) * sum{h, gasprice_adj_cendiv(cendiv,h) * GASUSED.l(cendiv,gb,h,t) * hours(h) / repgasquant_gb(cendiv,gb,t) }` should work for `repgasprice`

reeds/input_processing/fuelcostprep.py
kodiobika: Good call, thanks - done

reeds/core/terminus/report_params.csv
kodiobika: Updated to `repgasquant_gb` with units of Quads (to match the other `repgasquant*` parameters)

reeds/core/terminus/report.gms
patrickbrown4: Yep, the `gas_scale`s embedded in `gasprice` and `GASUSED` cancel out, and sums of `GASUSED` over `h` should always be handled as `GASUSED * hours`, so the current version looks right to me. Thanks!

PR #79, opened 2026-05-01
ci(dependabot): reduce noise and skip ReEDS run for dependabot updates

pesap:
## Summary
- reduce Dependabot update noise by grouping updates and slowing cadence
- run github-actions updates every 14 days via cron
- skip ReEDS and R2X CI jobs when actor is dependabot bot to avoid license-related failures

## Validation
- ruby YAML parse check for .github/dependabot.yml and .github/workflows/python-app.yaml


.github/dependabot.yml
patrickbrown4, line 1: Thanks Pedro. Can we limit dependabot to the github actions and ignore `environment.yml` entirely? We inflexibly specify the major.minor versions and only update the environment every few years, so it doesn't seem necessary to check every 2 weeks.

pesap: Sounds good. I removed completely the python dependabot update. Now it is isolated to the CI.

pesap:
@patrickbrown4 it should be ready for re-review

.github/dependabot.yml
```diff
@@ -1,60 +1,15 @@
...
-      include: "scope"
-    rebase-strategy: "auto"
-
   - package-ecosystem: "github-actions"
-    directory: /
+    directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "cron"
+      cronjob: "0 6 */14 * *"
```
patrickbrown4: (minor nitpick) Are comments allowed in this file? If so, maybe add one here explaining what this means? (It looks like it will execute twice near the end of the month, on the 29th and then on the 1st - maybe you could just do every second Monday instead? Not a bit deal; up to you.)

patrickbrown4 (approved):
Looks good to me, thanks!

.github/dependabot.yml
pesap: Sounds good. Added a message for that with the suggestion of the cadence see https://github.com/ReEDS-Model/ReEDS/pull/79/commits/8c71c4c81e7ef43e1a330fabf1e64e6fee110328

PR #80, opened 2026-05-01
Migrate reVeal2ReEDS pipeline to Hourlize

kodiobika:
## Summary
This PR updates hourlize to include the reVeal2ReEDS pipeline, which replaces EER data center demand projections with user-specified data center demand projections from reVeal or other similarly formatted sources.

## Technical details 
- Adds `hourlize/reveal2reeds/reveal2reeds.py`, which contains the functions that execute the pipeline

As a test, I ran the updated hourlize workflow with the EER `IRA_low` scenario, replacing its data center demand projections with the `central` scenario from the reVeal results. The differences in overall CONUS-wide load over time are below.

eer_vs_reveal

A comparison between the national reference case in `main` and the national reference case using this updated load profile is below. The overall increase in load results in greater capacity buildout and generation, which is predominantly wind and solar with some batteries and gas.

[results-0529_Main_USA_defaults1,0601_r2R_USA_defaults.pptx](https://github.com/user-attachments/files/28808004/results-0529_Main_USA_defaults1.0601_r2R_USA_defaults.pptx)

### Details to double-check

No


hourlize/load.py
```diff
@@ -283,6 +302,10 @@ def create_hourly_state_load_for_model_year(
     # sectoral load from the raw load profiles
     replacement_load_list = []
     for sector in replace_sectors:
+        # Skip 'data centers' sector, as it was already processed above
+        if sector == 'Data Centers':
+            continue
```
SoLaraS2: I don't understand why the DC replacement can't happen inside this preexisting loop, is this a coding convention? I just don't see a reason on the processing side, it's at the same level as the loop already existing so we aren't skipping or double calculating anything. Ignore if this is just how it has to be :)

hourlize/reveal2reeds/reveal2reeds.py
```diff
@@ -0,0 +1,239 @@
...
+            pd.concat(
+                [df_load, state_subsector_data_center_demand_hourly],
+                ignore_index=True
+            )
+            .groupby(
+                [
+                    'weather_datetime',
+                    'sector',
+                    'subsector',
+                    'dispatch_feeder'
```
SoLaraS2: Actually the "dispatch_feeder" level is generally summed over. I spoke about this with Anne Hamilton when I first started working with the EER files, so you should be fine to drop it as a level distinction in general. I thought I'd share this bit of info! (It doesn't always make a huge difference, but to get the exact same results as current scripts it'd probably be best to sum over it)

hourlize/reveal2reeds/reveal2reeds.py
```diff
@@ -0,0 +1,239 @@
...
+    propagation_by_weather_year = get_propagation_by_weather_year(
+        propagation_source_path,
+        scenario
+    )
+
+    # Estimate national hourly load values for each weather year
+    # by multiplying the propagation factors by national data
+    # center demand for the model year.
+    national_data_center_demand_hourly = pd.DataFrame(
+        index=df_load['weather_datetime'].drop_duplicates()
```
SoLaraS2, line 60: I tried to trace where df_load is coming from at this point, but if it has repeated timestamps it's likely because of dispatch_feeder, and it makes a bigger difference when we are considering the rest-of-economy subsectors, so I generally try to sum them to make unique timestamps rather than dropping. But if by this point you are absolutely sure there are no duplicates of this nature, what are the other options that would cause a need for this process?

SoLaraS2 (commented):
My comments are more so questions I had while walking through the code, they might be less helpful if the "issues" I addressed are overshadowed by something else, and as such the code needs to be as it currently is. Looking good though! :D

hourlize/inputs/load/sector_config.json
```diff
@@ -45,10 +45,12 @@
...
-    "model_years": [2021, 2025, 2030, 2035, 2040, 2045, 2050],
-    "filepaths": ["/kfs2/projects/eerload/challoran/eer_splice/dummy_agg_op_datacenters_by_state.csv"],
-    "unit_conversion_factor": 1,
-    "timezone": "Etc/GMT+6",
-    "regional_scope": "state"
+    "model_years": [2025, 2030, 2035, 2040, 2045, 2050],
+    "scenario": "central",
+    "national_demand_source": "/projects/largeload/geospatial/runs/random_forest_base_weights_01_09_2026/downscaling_2026-01-07_agg64/eer_national_central/eer_national_central_downscaled_projections.csv",
+    "cooling_proportions_source": "/projects/largeload/reVeal2ReEDS/files/{scenario}_dc_cooling_prop.csv",
+    "propagation_source": "/projects/largeload/reVeal2ReEDS/files/weather_year_propagation.csv",
```
clairehalloran: @SoLaraS2 do the original  `cooling_proportions_source` and `propagation_source` files exist on the HPC in the /projects/eerload directory? It would be better to point to those locations than a location in the largeloads project.

hourlize/inputs/load/sector_config.json
```diff
@@ -45,10 +45,12 @@
     "subsectors": {
       "commercial": ["data center cooling", "data center it"]
     },
-    "model_years": [2021, 2025, 2030, 2035, 2040, 2045, 2050],
-    "filepaths": ["/kfs2/projects/eerload/challoran/eer_splice/dummy_agg_op_datacenters_by_state.csv"],
-    "unit_conversion_factor": 1,
-    "timezone": "Etc/GMT+6",
-    "regional_scope": "state"
+    "model_years": [2025, 2030, 2035, 2040, 2045, 2050],
+    "scenario": "central",
```
clairehalloran: It would be good to add a README in [hourlize/inputs/load](https://github.com/ReEDS-Model/ReEDS/tree/main/hourlize/inputs/load) that explains these config options. In particular, does scenario here refer to EER baseline vs. IRA low vs. 100by2050 (which EER calls "central") or central vs. high data center demand?

clairehalloran (changes requested):
Thanks for integrating this feature! I tested this on the HPC and it worked for me.

I'd like to see some documentation added about this in a README in hourlize/input/load similar to the one in hourlize/input/config. In particular, it would be good to clarify that this pipeline expects total data center load, and then splits that into IT and cooling profiles based on EER's assumption of an annual PUE of 1.15. (I think I followed that logic correctly, let me know if not.)

hourlize/load.py
kodiobika: Good call, thanks! I moved it inside the loop

hourlize/reveal2reeds/reveal2reeds.py
kodiobika: Thanks for the heads up! Just dropped it from the list

hourlize/reveal2reeds/reveal2reeds.py
kodiobika: `df_load` at this point is actually just the initial EER load profile so the duplicate timestamps come from there being multiple sectors, subsectors, and dispatch_feeders, which I think is expected and should be fine. And in this case we're just getting the unique timestamps to create an index for the new hourly data center demand profile, so whether we use drop_duplicates or sum and take the unique indices should give the same result, but let me know if you have a different read

hourlize/inputs/load/sector_config.json
kodiobika: Done! I was actually using the scenario as a catch-all for the PUE, state load factors, and weather year propagation inputs but just realized that isn't ideal since they don't have the same sets of scenarios (and someone may want to mix-and-match anyway), so I've updated the config to specify both a filepath and a scenario wherever relevant and added explanations to the readme

kodiobika:
@SoLaraS2 @clairehalloran Thanks for the reviews and sorry for the delay in response! This should be ready for re-review whenever you get a chance

clairehalloran (approved):
Looks good to me, thanks Kodi!

PR #85, opened 2026-05-08
 Derate thermal capacity reserve margin contribution with FOR

clairehalloran:
## Summary
This PR derates the contribution of thermal capacity to the reserve margin based on technology-specific mean forced outage rates (FORs) during the top net peak load hours during each ccseason. This change only applies when the capacity credit resource adequacy method is used, i.e. `GSw_PRM_CapCredit=1`. This update does not change capacity credit for VRE, hydro, or storage.

### Implementation notes 
For each thermal technology, the mean FOR during the top net peak load hours is calculated for each ccseason in `resource_adequacy/capacity_credit.py` based on the same hourly FORs used in PRAS. The mean FOR for each technology is written to the `handoff_{t}.gdx` file for each year that resource adequacy calculations are run. The capacity of thermal generation is multiplied by `(1 - FOR)` in `eq_reserve_margin`.

### Additional changes 
- Added ValueError in `runreeds.py` to disallow `GSw_PRM_UpdateMethod=0`,`GSw_PRM_CapCredit=1`, and `GSw_PRM_StressIterateMax>0`, which iterates to reach an NEUE threshold without updating the PRM values between iterations.
- In the Pacific_CC test case, updated `GSw_PRM_UpdateMethod` from the default of `0` to `1` to avoid raising this ValueError and test endogenously increasing the PRM.

I ran the USA_fast test case with the following switches:
- `GSw_PRM_CapCredit=1`
- `GSw_PRM_UpdateMethod=1`
- `GSw_PRM_scenario=0.1`
- `GSw_PRM_UpdateFraction=0.05`
- `GSw_PRM_StressIterateMax=20`

Derating thermal capacity increased total capacity in most years:
image

In 2050, there is a net increase in generation capacity with combined cycle and steam turbine prime movers (gas-CC, H2-CC, coal, and coal-CCS), which have lower outage rates than combustion turbines at low temperatures. 

Derating also led to lower PRM values:

PRM with derate:
image

PRM without derate:
image

This feature resulted in a small increase in runtime:
image

Full USA_fast_CC comparison here: [results-main,derate.pptx](https://github.com/user-attachments/files/28648186/results-main.derate.pptx)

No change in capacity for USA_defaults:
image

Full USA_defaults comparison here: [results-USA_defaults_main,USA_defaults_derate.pptx](https://github.com/user-attachments/files/28721482/results-USA_defaults_main.USA_defaults_derate.pptx)

### Details to double-check

Yes, as an alternative to looking up pandas documentation and for minor line completion.


reeds/resource_adequacy/capacity_credit.py
```diff
@@ -152,13 +152,35 @@ def reeds_cc(t, tnext, casedir):
     ### Get the non-duplicated profiles
     resource_profiles = resources.drop_duplicates('resource')
 
+    ### Get forced outage profiles, flatten, and melt
+    forced_outage_rate = reeds.io.get_outage_hourly(inputs_case,'forced')
+
+    # drop techs that won't use this capacity credit before stacking
+    techs_to_drop = ['battery_li','can-imports','distpv', 'electrolyzer',
+                    'hydd', 'hyded', 'hydend', 'hydnd', 'hydnpd', 'hydnpnd',
+                    'hydro', 'hydsd', 'hydsnd', 'hydud', 'hydund', 'pumped-hydro']
```
patrickbrown4: Whenever possible, we'd like to avoid hard-coding technology names within Python/GAMS scripts, as it increases the number of places that need to be modified when technologies are added/changed. (I know we do it now but we'd at least like to avoid adding it in new places.) Is there a tech group (or groups) in `inputs/tech-subset-table.csv` (available from `reeds.techs.get_tech_subset_table()` or [like this](https://github.com/ReEDS-Model/ReEDS/blob/c355a9e662824504a17958ba92509967cd220938/reeds/resource_adequacy/prep_data.py#L157-L159) if techs need to be expanded) you could use instead? If not, you could add a tech group and use it here.

reeds/resource_adequacy/capacity_credit.py
```diff
@@ -152,13 +152,35 @@ def reeds_cc(t, tnext, casedir):
...
+                            .stack(level=1)
+                            .reset_index()
+                            .rename(columns={"level_0": "timestamp"})
+                            )
+
+    forced_outage_rate = pd.melt(
+        forced_outage_rate,
+        id_vars=['timestamp','r'], 
+    )
+
```
patrickbrown4: In general it's best to avoid reshaping hourly dataframes with thousands of columns and multiple weather years, as it can be super slow/memory-intensive. It looks like you could instead:
1. Keep `forced_outage_rate` in its shape after the `.drop()`
2. Turn `top_net_load_hours` into a dictionary of timestamps:
```python
    hours = top_net_load_hours.groupby(['ccseason', 'r'])['*timestamp'].agg(list)
    mean_forced_outage_rate = pd.concat({
        (ccseason, r): forced_outage_rate.loc[h].xs(r, 1, 'r').mean()
        for (ccseason, r), h in hours.items()
    }, names=('ccseason', 'r'))
```

There might need to be a few other adjustments but the general idea is to keep the forced outage rates in the original hours x (techs, regions) format since it's much faster to query the data you need than to reshape the whole thing.

Could you give that a shot and see if it speeds up `capacity_credit.py`? Happy to chat about it if I overlooked something.

reeds/resource_adequacy/capacity_credit.py
```diff
@@ -152,13 +152,35 @@ def reeds_cc(t, tnext, casedir):
...
 
+    ### Get forced outage profiles, flatten, and melt
+    forced_outage_rate = reeds.io.get_outage_hourly(inputs_case,'forced')
+
+    # drop techs that won't use this capacity credit before stacking
+    techs_to_drop = ['battery_li','can-imports','distpv', 'electrolyzer',
+                    'hydd', 'hyded', 'hydend', 'hydnd', 'hydnpd', 'hydnpnd',
+                    'hydro', 'hydsd', 'hydsnd', 'hydud', 'hydund', 'pumped-hydro']
+    
+    forced_outage_rate = forced_outage_rate.drop(columns=techs_to_drop, errors='ignore')
```
patrickbrown4:
```suggestion
    forced_outage_rate = forced_outage_rate.drop(columns=techs_to_drop, level=0, errors='ignore')
```

reeds/resource_adequacy/capacity_credit.py
```diff
@@ -464,6 +486,39 @@ def pivot_melt_data(df):
...
+
+    # map hour to timestamp
+    top_net_load_hours = top_net_load_hours.merge(
+        hmap_allyrs[['hour','*timestamp','actual_h']], on='hour', how='left'
+    ).drop('hour', axis=1)
+    top_net_load_hours['*timestamp'] = pd.to_datetime(top_net_load_hours['*timestamp'])
+    top_net_load_hours['h'] = top_net_load_hours['actual_h']
+    
+    # join top hours to outage rates based on timestamp and region
+    forced_outage_rate_top_hours = top_net_load_hours.merge(
```
patrickbrown4: Could you explain the reasoning for using net load instead of load (since net load is a VRE concept but here we're looking at non-VRE)?

For context, the RHS of `eq_reserve_margin` is peak rather than net peak. (Of course, that's one of the weird things about capacity credit, that we use peak for some things and net peak for others.)

reeds/resource_adequacy/ra_calcs.py
```diff
@@ -165,8 +166,18 @@ def main(t, tnext, casedir, iteration=0):
...
+        and ('user' not in sw['GSw_PRM_StressModel'].lower())
+        ):
+        print('identifying new stress periods...')
+        tic = datetime.datetime.now()
+    elif (
+        int(sw['GSw_PRM_CapCredit']) and (int(sw['GSw_PRM_StressIterateMax']) > 0)
+    ):
+        print('updating PRM values...')
+        tic = datetime.datetime.now()
+
```
patrickbrown4: Kind of a lot of logic just for a status message, and `stress_periods.py` prints its own messages, so personally I would just cut it:
```suggestion
    tic = datetime.datetime.now()
```

cases_test.csv
```diff
@@ -19,6 +19,7 @@ GSw_RetirePenalty,,,,,0,,,,,,,,,,,,,,,,,,,,,,,,,
 GSw_FakeData,,,,,,1,1,1,,,,,,,,,,,,,,,,,,,,,,
 GSw_PRM_CapCredit,,,,,,,,1,1,,,,,,,,1,,,,,,,,,,,,,
 GSw_PRM_scenario,,,,,,,,,static,,,,,,,,static,,,,,,,,,,,,,
+GSw_PRM_UpdateMethod,,,,,,,,,1,,,,,,,,,,,,,,,,,,,,,
```
patrickbrown4, line 22: This is for Pacific_CC, right? It will slow down the test but I agree it's good to have to demonstrate best practice with capacity credit

runreeds.py
```diff
@@ -386,6 +386,16 @@ def check_compatibility(sw):
...
                 raise ValueError(err)
 
+    if int(sw['GSw_PRM_UpdateMethod']) == 0 and int(sw['GSw_PRM_CapCredit']) == 1 and int(sw['GSw_PRM_StressIterateMax']) > 0:
+        raise ValueError(
+            "The combination of GSw_PRM_UpdateMethod=0, GSw_PRM_CapCredit=1, "
+            "and GSw_PRM_StressIterateMax>0 is not supported."
+            "To iteratively update the PRM, set GSw_PRM_UpdateMethod to one of: "
+            "(1) static update set by GSw_PRM_UpdateFraction; "
+            "(2) dynamic update informed by PRAS; "
+            "(3) dynamic update but only after all new stress periods have been added"
```
```suggestion
            "The combination of GSw_PRM_UpdateMethod=0, GSw_PRM_CapCredit=1, "
            "and GSw_PRM_StressIterateMax>0 is not supported.\n"
            "To iteratively update the PRM, set GSw_PRM_UpdateMethod to an integer between 1-3:"
            "\n1: static update set by GSw_PRM_UpdateFraction; "
            "\n2: dynamic update informed by PRAS; "
            "\n3: dynamic update but only after all new stress periods have been added"
```

docs/source/model_documentation.md
```diff
@@ -2460,11 +2460,12 @@ Adapted from {cite}`maiIncorporatingStressfulGrid2024`.
...
 The calculation of capacity credit for VRE is described in the [VRE capacity credit](#vre-capacity-credit) section;
 the method for storage is described in the [storage capacity credit](#storage-capacity-credit) section.
-Thermal generators are given a capacity credit of 100% in ReEDS;
-technology-specific [outage rates](#outage-rates) for thermal generators are not considered in this method
-and are instead assumed to be factored into the [planning reserve margin](#planning-reserve-margins).
 
 
+#### Thermal generation capacity credit
+
+For thermal generators (i.e. combined cycle, combustion turbine, nuclear (conventional and SMR), and steam (coal)), ReEDS estimates a seasonal capacity credit for each region/technology combination based on temperature-dependent forced outage rates (described in detail in [outage rates](#outage-rates)). To calculate each technology's contribution to the seasonal reserve margin, its nameplate capacity for each technology is multiplied by $(1 - \bar{FOR})$, where $\bar{FOR}$ is the mean forced outage rate during the top 20 net load hours cross all modeled [weather years](#weather-years) for each season.
```
patrickbrown4:
```suggestion
For thermal generators (i.e. combined cycle, combustion turbine, nuclear (conventional and SMR), and steam (coal)), ReEDS estimates a seasonal capacity credit for each region/technology combination based on temperature-dependent forced outage rates (described in detail in [outage rates](#outage-rates)). To calculate each technology's contribution to the seasonal reserve margin, its nameplate capacity is multiplied by $(1 - \bar{FOR})$, where $\bar{FOR}$ is the mean forced outage rate during the top 20 net load hours cross all modeled [weather years](#weather-years) for each season. 
```

patrickbrown4 (commented):
This is a nice addition, thanks!

My main comment is that I think the `capacity_credit.py` logic could be sped up by avoiding the reshaping of the hourly outage-rate dataframe. I added some snippets to get started but happy to chat if easier.

cases_test.csv
clairehalloran: Yes, this is for Pacific_CC to avoid raising the ValueError I added in `runreeds.py`. Alternatively, I could set `GSw_PRM_StressIterateMax` to 0 like in github_MA_county_CC and PJM_county_CC, but then there aren't any tests using iterative capacity credit approaches.

reeds/resource_adequacy/capacity_credit.py
clairehalloran: I initially used net load to be consistent with the VRE and storage capacity credit. Revisiting it, I think thermal capacity credit would be more closely aligned with ELCC if we use load. I'll plan to update.

cases_test.csv
patrickbrown4: Yeah I agree it's good to have at least one test case where we endogenously increase the PRM, so I like this addition.

reeds/resource_adequacy/capacity_credit.py
wesleyjcole: Given that we are using net load as a proxy for the highest risk hours under the capacity credit method, doesn't it make sense to stick with net load here?  Happy to talk through this if that would be helpful.

patrickbrown4: It's not obvious to me which is better.

- Using net load for everything is consistent in the sense that all the capacity credits use the same profile
- Subtracting the hourly availability of existing fossil from load is consistent in the sense that that's what we do for VRE (and with hourly outage rates, fossil is also variable)
- It seems hard to call net load, defined as load - (wind+solar), tech-neutral

The stress periods approach seems the most tech-neutral to me, while I think the point of continuing to support the capacity credit method is to be able to reproduce what some regions are currently doing in practice. If that's the goal, I think the most helpful thing would be to review how regions that currently use tech-dependent capacity credits / ELCCs do it, and try to reproduce that as closely as possible.

clairehalloran: Based on a quick review, it seems like using net load as a proxy for highest risk hours is most aligned with current practice in the largest markets:
- MISO is moving towards accrediting all resource classes based on their UCAP during high-risk periods (see "Transition" on p 28 and "Conclusion" on p 29 [here](https://cdn.misoenergy.org/Resource%20Accreditation%20White%20Paper%20Version%202.1630728.pdf))
- PJM accredits generation "based on expected contribution during periods of system reliability risk in each season" and uses temperature-dependent FOR for thermal resources (see p 21 [here](https://www.pjm.com/-/media/DotCom/committees-groups/cifp-ra/2023/20230614/20230614-item-02---pjm-cifp-stage-3-proposal.pdf))
- Based on a 1.5 year old PJM [review](https://www.pjm.com/-/media/DotCom/committees-groups/task-forces/elccstf/2024/20241205/20241205-item-03---elcc-accreditation-methodology.pdf), it looks like ISONE and NYISO derate thermal generation based on historical seasonal forced outage rates

reeds/resource_adequacy/capacity_credit.py
clairehalloran: Thanks for the tip! The runtime for `capacity_credit.py` is just 9 minutes longer after implementing these changes.

reeds/resource_adequacy/capacity_credit.py
patrickbrown4: Thanks Claire. From a quick scan, the references still leave it a little vague; the PJM links don't mention net load or net peak, and the MISO link mentions "peak or net-peak" hours.

To me it seems like the most tech-neutral approach would be to calculate net load by subtracting the hourly availability of all generator types from load.

- For the old method of modeling thermal without hourly availability, there'd be no difference from defining net load as (load - wind - solar)
- When using hourly availability, hours with high forced outage rates for the existing fleet would have a higher net load and would be more likely to show up as peak-net-load hours and be included in the capacity credit
- The peak net load would presumably be negative if the system meets the RA target, but that seems ok; all that really matters for the capacity credit is the timing of the peak net load hours, which in this case are the hours of minimum surplus (which seems like a good definition of system stress to me)
- Otherwise, it seems like the metric would artificially favor technologies whose hourly availabilities are not subtracted as part of the net load definition

Happy to discuss if helpful though. And since the current method is still an improvement as is, a change to the net load definition (if we decide to make it) could be saved for a second PR.

wesleyjcole: I like the construct of

net_load = load - hourly_availability

I think that would be an improvement.  That would be most general, so should easily adapt with other changes related to availability.

reeds/resource_adequacy/ra_calcs.py
```diff
@@ -138,6 +138,7 @@ def main(t, tnext, casedir, iteration=0):
             'cc_old': pd.DataFrame(columns=['i','r','ccreg','szn','t','Value']),
             'cc_evmc': pd.DataFrame(columns=['i','r','szn','t','Value']),
             'sdbin_size': pd.DataFrame(columns=['ccreg','szn','bin','t','Value']),
+            'mean_forced_outage_rate': pd.DataFrame(columns=['i','r','szn','t','Value']),
```
patrickbrown4: I'm actually a little surprised the others work with `szn` (I would have thought it'd be more strict) but it's loaded with `ccseason` as the index so let's use that
```suggestion
            'mean_forced_outage_rate': pd.DataFrame(columns=['i','r','ccseason','t','Value']),
```

reeds/resource_adequacy/capacity_credit.py
```diff
@@ -152,13 +152,25 @@ def reeds_cc(t, tnext, casedir):
...
+    forced_outage_rate = forced_outage_rate.drop(columns=techs_to_drop, level=0, errors='ignore')
+    
     # Remove the "8760" safety valve bin
     safety_bin = max(sdb['bin'].values)
     sdb = sdb[sdb['bin'] != max(sdb['bin'])]
     sdb = [int(x) for x in sdb['bin']]
 
     # Temporal definitions
     h_dt_szn = pd.read_csv(os.path.join('inputs_case', 'rep', 'h_dt_szn.csv'))
+    hmap_allyrs = pd.read_csv(os.path.join('inputs_case','rep','hmap_allyrs.csv'))
```
patrickbrown4: Unused, so can be deleted
```suggestion
```

reeds/resource_adequacy/capacity_credit.py
```diff
@@ -152,13 +152,25 @@ def reeds_cc(t, tnext, casedir):
     ### Get the non-duplicated profiles
     resource_profiles = resources.drop_duplicates('resource')
 
+    ### Get forced outage rates for thermal technologies
+    forced_outage_rate = reeds.io.get_outage_hourly(inputs_case,'forced')
+    
+    tech_subset_table = reeds.techs.expand_GAMS_tech_groups(
+    reeds.techs.get_tech_subset_table(casedir).reset_index()
```
patrickbrown4: (indentation nitpick)
```suggestion
        reeds.techs.get_tech_subset_table(casedir).reset_index()
```

reeds/resource_adequacy/capacity_credit.py
```diff
@@ -152,13 +152,25 @@ def reeds_cc(t, tnext, casedir):
...
+    ).set_index('tech_group').i
+    techs_to_keep = tech_subset_table.loc['THERMAL'].tolist()
+    techs_to_drop = [col for col in forced_outage_rate.columns.levels[0] if col not in techs_to_keep]
```
patrickbrown4: A tiny bit safer to get the values by name (`i`) instead of position in case the level ordering were to change
```suggestion
    techs_to_drop = [
        col for col in forced_outage_rate.columns.get_level_values('i').unique()
        if col not in techs_to_keep
    ]
```

inputs/tech-subset-table.csv
```diff
@@ -1,74 +1,74 @@
...
+Gas-CC_H_1x1-CCS_max,,,,YES,YES,YES,,YES,,,YES,YES,YES,,YES,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,YES,,,
+Gas-CC_H_2x1-CCS_mod,,,,YES,YES,YES,,YES,,,YES,YES,YES,YES,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,YES,,,
+Gas-CC_H_2x1-CCS_max,,,,YES,YES,YES,,YES,,,YES,YES,YES,,YES,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,YES,,,
+Gas-CC-CCS-F1,,,,YES,YES,YES,,YES,,,YES,YES,YES,YES,,YES,YES,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,YES,,,
+Gas-CC-CCS-F2,,,,YES,YES,YES,,YES,,,YES,YES,YES,YES,,YES,,YES,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,YES,,,
+Gas-CC-CCS-F3,,,,YES,YES,YES,,YES,,,YES,YES,YES,YES,,YES,,,YES,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,YES,,,
+Gas-CT,,,,YES,,,YES,,YES,,YES,YES,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,YES,,,
+Gas-CT_aero,,,,YES,,,YES,,YES,,YES,YES,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,YES,,,
+ng-fuel-cell,,,,YES,,,,,,YES,,YES,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,YES,,,
```
patrickbrown4, line 28: Are fuel cells usually classified as thermal? I don't actually know but I would have thought not

inputs/tech-subset-table.csv
```diff
@@ -1,74 +1,74 @@
...
+hydED,,,,,,,,,,,,,,,,,,,,,YES,,YES,,,,,,,,,,,,,,,,,,,,,,,,,,,,YES,YES,,,,,,,,,,,,,,,,,,,,,,,,,YES,,,,
+hydEND,,,,,,,,,,,,,,,,,,,,,YES,,YES,,,,,,,,,,,,,,,,,,,,,,,,,,,,YES,,YES,,,,,,,,,,,,,,,,,,,,,,,,YES,,,YES,
+electrolyzer,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,YES,YES,,,,,,,,,,,,
+smr,,,,,,,,,,,YES,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,YES,YES,YES,,,,,,,,,,,
+smr_ccs,,,,,,,,,,,YES,,YES,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,YES,YES,YES,,,,,,,,,,,
+dac,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,YES,,,YES,,,,,,,,,,
+dac_gas,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,YES,,,YES,,,,,,,,,,
+egs_allkm_1*egs_allkm_10,,,,,,,,,,,,,,,,,,,,,YES,,YES,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,YES,,YES,YES,,,YES,,,,,,,,,,,,,,,,,,,,
+geohydro_allkm_1*geohydro_allkm_10,,,,,,,,,,,,,,,,,,,,,YES,,YES,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,YES,YES,,,,YES,,,,,,,,,,,,,,,,,,,,,
+egs_nearfield_1*egs_nearfield_10,,,,,,,,,,,,,,,,,,,,,YES,,YES,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,YES,,YES,,YES,YES,,,,,,,,,,,,,,,,,,,,,
```
patrickbrown4, line 74: I think geothermal would be thermal. But I guess what really matters is whether they're in the tech groups listed in the values of this dictionary (minus hydro/PSH): https://github.com/ReEDS-Model/ReEDS/blob/8a9c2fe1aaef38b67952badbc6e9e091e39018a5/reeds/input_processing/outage_rates.py#L29-L36

With that in mind, maybe you could call it THERMALDERATE or FORTECH or TEMPDERATE or TEMPOUTAGE (like REFURBTECH and cf_tech)? (I don't know why cf_tech isn't capitalized)

patrickbrown4 (approved):
Let's discuss what to do about the net load change (I think I lean toward saving it for a followup PR, with an issue as placeholder in the meantime, since it would also affect VRE capacity credits and would get kind of deep into `capacity_credit.py`), but otherwise just small formatting/naming comments. Thanks!

reeds/resource_adequacy/capacity_credit.py
clairehalloran: Opened Issue https://github.com/ReEDS-Model/ReEDS/issues/118 as a TODO for this change.

PR #93, opened 2026-05-14
End ReEDS run early if no valid scenarios

kennedy-mindermann:
## Summary
This PR updates `runreeds.py` to end the run in a few different places if there are no valid scenarios left.

I also did a minor update to one of the existing error messages so it was easier to read.

I didn't do a full run since it doesn't change the model itself. I did check all combinations of getting a run started to make sure it behaved as expected.

### Details to double-check

  - [ ] Units are specified
- [ ] Code formatting standardized 
- [ ] Reusable functions used where possible instead of copy/pasted code

### DIFF
--- a/runreeds.py
+++ b/runreeds.py
@@ -854,6 +854,7 @@ def setupEnvironment(
         single=single,
         skip_checks=skip_checks,
     )
+    all_case_names = list(df_cases.columns)
     ## Propagate debug setting
     if debug:
         df_cases.loc['debug'] = str(debug)
@@ -885,8 +886,8 @@ def setupEnvironment(
         for s in single.split(','):
             if s not in df_cases:
                 err = (
-                    f'Specified single={single} but available cases are:\n'
-                    + '\n> '.join([c for c in df_cases.columns])
+                    f'Specified single={single} but available cases are: '
+                    + ', '.join([c for c in df_cases.columns])
                 )
                 raise KeyError(err)
         df_cases = df_cases[single.split(',')].copy()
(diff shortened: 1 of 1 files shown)


runreeds.py
```diff
@@ -885,8 +886,8 @@ def setupEnvironment(
         for s in single.split(','):
             if s not in df_cases:
                 err = (
-                    f'Specified single={single} but available cases are:\n'
-                    + '\n> '.join([c for c in df_cases.columns])
+                    f'Specified single={single} but available cases are: '
+                    + ', '.join([c for c in df_cases.columns])
```
patrickbrown4: Personally I would revert and keep these on their own lines as I think that's easier to read

runreeds.py
```diff
@@ -854,6 +854,7 @@ def setupEnvironment(
         single=single,
         skip_checks=skip_checks,
     )
+    all_case_names = list(df_cases.columns)
```
patrickbrown4: For readability, define further down, next to the one place it's used

runreeds.py
```diff
@@ -896,6 +897,9 @@ def setupEnvironment(
     outpaths = [os.path.join(reeds_path,'runs',f'{BatchName}_{case}') for case in casenames]
     existing_outpaths = [i for i in outpaths if os.path.isdir(i)]
     if len(existing_outpaths):
+        if len(existing_outpaths) == 0:
+            print('Existing output directories were not overwritten. Exiting without starting runs.')
+            quit()
```
patrickbrown4: This will never run since it's inside the `if len(existing_outpaths)` block, so it can be removed
```suggestion
```

runreeds.py
```diff
@@ -951,6 +963,9 @@ def setupEnvironment(
         WORKERS = simult_runs
     else:
         WORKERS = int(input('Number of simultaneous runs [integer]: '))
```
```suggestion
        WORKERS = int(input('Number of simultaneous runs [positive integer]: '))
```

runreeds.py
```diff
@@ -951,6 +963,9 @@ def setupEnvironment(
...
+        if WORKERS <=0:
+            print('Number of simultaneous runs must be at least 1. Exiting.')
+            quit()
```
patrickbrown4: I think this one at least should be a ValueError

```suggestion
        if WORKERS <= 0:
            raise ValueError(f'Provided {WORKERS} but must be a positive integer')
```

patrickbrown4 (approved):
Thanks Kennedy, some suggestions but otherwise looks good

PR #94, opened 2026-05-14
Fix eq_rsc_INVlim infeasibility for geohydro_allkm by not applying geo_discovery to exogenous capacity

### LINKED ISSUE #84: eq_rsc_INVlim infeasibility for geohydro_allkm when exogenous capacity exceeds m_rsc_dat * geo_discovery
### Summary

Infeasibility in `eq_rsc_INVlim` for `geohydro_allkm` for Western interconnect at county-level resolution. At p35023 in 2015, the right-hand side of `eq_rsc_INVlim` for geohydro_allkm_9 bin1 is negative by ~0.447 MW, so the model has no feasible value for `INV_RSC` (which is constrained to be ≥ 0).

**Branch / commit:** yc_pbitl (in ReEDS-2.0)

**Case / scenario:** cases_pbitl.csv → pbitl1_Western_Default_Stress

**Solve year where it fails:** 2015

**Region / tech:** county p35023, geohydro_allkm

**Steps to reproduce:**
1. Check out branch yc_pbitl (which I think is consistent with ReEDS main on this constraint)
2. Run county-level Western Interconnect
3. Model is infeasible at the 2015 solve

---

### Observation

`eq_rsc_INVlim` enforces:

```
m_rsc_dat * geo_discovery  >=  sum(INV_RSC) + capacity_exog_rsc
```

`geo_discovery(i,r,t)` is the fraction of the geothermal supply curve treated as "discovered" in year t (e.g., 0.10–0.15 in early years, monotonically increasing toward 1).

In p35023 in 2015, the left-hand side `m_rsc_dat * geo_discovery` evaluates to slightly less than the right-hand-side terms, producing a negative RHS of ~0.447 MW and making the constraint infeasible together with the prescribed builds requirement (`eq_forceprescription_power`).

---

### Solutions considered

A. Inflate `m_rsc_dat` so RHS becomes positive — e.g., extend the year-aware capacity top-up logic from commit 77fb3a6 in `b_inputs.gms` to add enough geothermal `m_rsc_dat` to bin1 so prescribed builds always fit, after accounting for `geo_discovery`.
   - Pros: confirmed to resolve the p35023 infeasibility; works within the same framework used for wind and solar; no equation changes needed
   - Cons: the "extra" MW added to bin1 are made-up numbers that show up later in capacity reports; piling everything into bin1 can mislead the optimizer's cost ranking across bins; treats the symptom rather than addressing why the constraint produced negative RHS in the first place

B. Disable `eq_rsc_INVlim` for offending bins via `flag_eq_rsc_INVlim`
   - Pros: targeted; preserves data
   - Cons: bin loses physical upper bound; may concentrate prescribed builds in one bin; conceptually a patch

C. Set `geo_discovery = 1` wherever a bin has negative RHS
   - Pros: trivially feasible
   - Cons: opens all bins in the region (geo_discovery is region-level, not bin-level); requires updating `rsc_capacity_scalar` to stay consistent

D. Reformulate `eq_rsc_INVlim` so `geo_discovery` is applied symmetrically to both sides of the supply-curve accounting, by also scaling the `capacity_exog_rsc` term by `rsc_capacity_scalar`. Equivalent form:

```
(m_rsc_dat - exog) * geo_discovery  >=  sum(INV_RSC)
```

   - Pros: mathematically clean; no parameter fabrication; no special-case logic; backward-compatible (non-geothermal techs have rsc_capacity_scalar = 1, so unchanged); auto-converges as geo_discovery → 1
   - Cons: touches the model equation itself, so future maintainers must understand the new convention; the conceptual interpretation of `geo_discovery` relative to existing capacity should be confirmed with the modeling team

---

### Suggested Solution: D

Two synchronized changes:

1. `c_supplymodel.gms` — `eq_rsc_INVlim`: multiply the `capacity_exog_rsc` term by `rsc_capacity_scalar`.
2. `d_solveoneyear.gms` — `rhs_eq_rsc_INVlim` pre-solve calculation: apply the same `geo_discovery` factor to the exogenous term to match the new equation.

---

### Validation

- [ ] Re-run pbitl1_Western_Default_Stress through 2015 — confirm infeasibility is resolved
- [ ] Spot-check INV_RSC.l for geohydro_allkm in p35023 across solve years to verify reasonable bin allocation
- [ ] Run a baseline (non-pbitl) Standard Scenarios case to confirm no regression for non-geothermal technologies
- [ ] Verify the geothermal branch of the 77fb3a6 pre-processing no longer adds capacity once D is in effect

---

Labels: bug

Yunzhi-Chen:
## Summary

`eq_rsc_INVlim` previously scaled the entire supply curve by `rsc_capacity_scalar` but subtracted exogenous capacity at full value:

```
m_rsc_dat * rsc_capacity_scalar
    >= sum(INV_RSC) + capacity_exog_rsc
```

This PR revises the constraint to:

```
(m_rsc_dat - capacity_exog_rsc) * rsc_capacity_scalar
    >= sum(INV_RSC)
```

This implements Solution D from #84. Existing exogenous capacity is treated as already discovered, so the discovery rate applies only to the remaining resource available for new investment. Technologies without a capacity scalar use a multiplier of 1, making the revised formulation algebraically equivalent for those technologies.

The current `main`-branch preprocessing added by @atpham88  addresses the prescribed-build infeasibility primarily through Solution C, setting `geo_discovery` to 1 when the available resource is insufficient, with Solution A as a fallback that adds resource to bin 1.

This PR instead resolves the underlying supply-curve accounting through Solution D while retaining a narrower Solution A-style safeguard for prescribed builds. The safeguard adds only the positive bin-1 resource shortfall needed to keep cumulative prescribed builds feasible.

## Technical details

Prescribed hydrothermal builds now retain their valid discovery trajectories rather than having `geo_discovery` overridden to 1 when the resource falls short.

The preprocessing in [`reeds/core/setup/b_inputs.gms`](reeds/core/setup/b_inputs.gms) calculates the resource required to support the year-specific cumulative prescriptions under the applicable discovery factor, net of existing capacity:

```
required resource in year t
    = cumulative prescribed builds through year t / geo_discovery(t)

resource shortfall
    = max(required resource across prescribed years)
      - (total supply-curve resource - existing capacity)
```

Only a positive shortfall is added to bin 1. This is an A-style feasibility safeguard, but it is limited to the amount needed for prescribed builds and does not replace the discovery trajectory.

## Implementation notes

- [`reeds/core/setup/c_model.gms`](reeds/core/setup/c_model.gms) — subtracts first-year exogenous capacity from `m_rsc_dat` before applying `rsc_capacity_scalar`, and removes the exogenous-capacity term from the investment side of `eq_rsc_INVlim`.
- [`reeds/core/solve/3_solve_oneyear.gms`](reeds/core/solve/3_solve_oneyear.gms) — updates the `rhs_eq_rsc_INVlim` presolve calculation to match the revised equation. It also uses the generic `rsc_capacity_scalar` instead of the geothermal-specific `geo_discovery`; previously, the presolve RHS did not match the model equation for technologies with a non-geothermal capacity scalar.
- [`reeds/core/setup/b_inputs.gms`](reeds/core/setup/b_inputs.gms) — preserves valid geothermal discovery trajectories and calculates any prescribed-build bin-1 top-up using year-specific cumulative prescriptions and discovery factors, net of existing capacity.
- [`docs/source/model_documentation.md`](docs/source/model_documentation.md) — documents the treatment of existing capacity and prescribed builds.

## Issues resolved

Closes #84.

Comparison reports against `main`:

- Reference case: 

image
image
image
[Reference.zip](https://github.com/user-attachments/files/31969765/Reference.zip)

- Western Interconnection case: 

image
image
image
[Western.zip](https://github.com/user-attachments/files/31969891/Western.zip)

The comparison results show increased geothermal capacity and generation under the revised formulation. System-cost NPV decreases slightly because the less restrictive geothermal constraint allows the model to select additional geothermal resources.

The Western Interconnection runs were capped at one iteration because a separate issue prevents the stress-period solve from converging.

### Details to double-check

- [x] Reran affected geothermal cases after the prescribed-build preprocessing update

- [ ] Zero impact on results of default case — hydrothermal geothermal results change by design. `exog_rsc` is otherwise limited to `onswind` and `upv`, which do not have a capacity scalar and are algebraically unaffected.
- [x] No change to process flow
- [x] No change to package requirements

GitHub Copilot implemented the initial fix and drafted the original description. Codex refined the prescribed-build logic and revised the PR description. Claude Code reviewed the diff against `main` and provided feedback on the description.


reeds/core/setup/c_model.gms
```diff
@@ -1025,9 +1025,13 @@ eq_rsc_INVlim(r,i,rscbin,t)$[tmodel(t)
 *but the combination of m_rsc_con and rsc_agg allows for those investments
 *to be limited by the numeraire techs' m_rsc_dat
 
-*capacity indicated by the resource supply curve (scaled by rsc_capacity_scalar)
-    m_rsc_dat(r,i,rscbin,"cap")$[not evmc(i)] * (
-        1$[not rsc_capacity_scalar_i(i)] + rsc_capacity_scalar(i,r,t)$rsc_capacity_scalar_i(i))
+*capacity indicated by the resource supply curve minus exogenous (pre-start-year)
+*capacity (scaled by rsc_capacity_scalar)
+    (m_rsc_dat(r,i,rscbin,"cap")$[not evmc(i)]
```
wesleyjcole: @Max-Vanatta - should this `[not evmc(i)]` be here?  It is the only place where it is used in the entire model.

wesleyjcole (approved):
This looks good.  If Max says we can remove the `evmc(i)` piece, I think it can be removed entirely, either here or in a separate PR.

Because it has a pcat change, would you coordinate with An on her startyear PR (#12) before merging this in?

docs/source/model_documentation.md
```diff
@@ -923,6 +923,8 @@ Resource availability for hydrothermal (left) and deep EGS (right) for the CONUS
 The default geothermal resource assumptions allow for hydrothermal sites.
 Identified hydrothermal resources are based on the U.S. Geological Survey's 2008 geothermal resource assessment.
 The undiscovered portion of the hydrothermal resource is limited by a discovery rate defined as part of the GeoVision Study {cite}`doeGeoVisionHarnessingHeat2019`.
+Existing exogenous hydrothermal capacity is treated as already discovered; the discovery rate applies to the remaining resource available for new investment.
+Prescribed builds retain this discovery treatment, with first-bin resource added only as needed to keep the prescriptions feasible.
```
wesleyjcole, line 927: I would remove this second line.  I think the first line is sufficient for documenting things.

```suggestion
```

wesleyjcole (approved):

Issue #37, opened 2026-04-20
Calculate greenfield transmission costs for all zone geometries directly


patrickbrown4:
## Summary

Our interregional transmission costs are currently out of sync with our interconnection costs (updated in [1836](https://github.nrel.gov/ReEDS/ReEDS-2.0/pull/1836) on the old repo to use MISO 2025 data); we plan to update them so all transmission costs use the same cost assumptions from MISO 2025.

In keeping with #16, we should also calculate costs for all region geometries directly instead of adding up the costs for smaller regions.

As part of this change, we will also switch from using 500 kV AC costs for all interfaces; instead, we'll use the highest existing voltage between each zone (with a floor of 138 kV). For the few interfaces that are connected by 765 kV, that will decrease the USD/MWmile cost; for most other interfaces (connected by <500 kV), it will increase the USD/MWmile cost.

## TODO

- [x] Update `inputs/transmission/transmission_cost_{polarity}*` to use the zonehashes in a single file (the same as `itl_NARIS.csv`), removing the ba/county distinction
- [x] Same for `inputs/transmission/transmission_distance_*`
  - [x] Need to adapt it to read all values, not just for the interfaces listed in `inputs/zones/GSw_ZoneSet/interfaces_r.csv`
  - [x] Need to adapt for land-to-offshore and offshore-to-offshore. Should have values in:
    - [x] transmission_miles
    - [x] tranloss
    - [x] transmission_line_fom
    - [x] trancap_fut
    - [x] transmission_cost_nonac
- [x] Add existing/planned HVDC lines individually:
  - [x] TransWest Express
  - [x] SunZia
    - https://www.caiso.com/documents/apr-13-2026-informational-filing-of-effective-date-transmission-control-agreement-regarding-subscriber-participating-transmission-owners-sunzia-er25-169.pdf
    - https://www.eenews.net/articles/largest-us-renewable-project-begins-generating-electricity/
    - https://www.openstreetmap.org/way/1420870282#map=8/33.351/-108.567
- [x] Remove the `transmission_capacity_future*` files for now, since we'll need to move from BAs to either 1) point-to-point HVDC lines as in `inputs/transmission/hvdc_lines.csv` or 2) new bottom-up ITLs for specific model years, calculated with planned branches added
[...]

Issue #16, opened 2026-04-15
Reformat spatial inputs pipeline for more flexibility


patrickbrown4:
## Overview

To support more spatial flexibility as we redraw the ReEDS zones, we're planning to make the following changes to the spatial data pipeline. We'll add to this issue (and/or create subsidiary issues) and discuss the proposed design as work progresses, but opening now for awareness and discussion.

## Goal
User provides `county2zone.csv` and `hierarchy.csv`; then everything else (aside from transmission) is handled automatically.

## Pieces
- [x] Create aggregation functions for special-case technologies whose inputs will still be at 134-zone resolution (until we eventually get raw data at site or county or state resolution), and run them in either a new input_processing script or as part of the existing scripts.
  - [x] Hydro
  - [x] Geothermal
  - [x] DR
  - [x] Water cooling
  - [x] ~~[inputs/capacity_exogenous/cappayments_ba.csv](https://github.com/ReEDS-Model/ReEDS/blob/cd0ee68f684ab61e107e07da6dbe4c8bf27c1389/inputs/capacity_exogenous/cappayments_ba.csv) (moved from [779](https://github.nrel.gov/ReEDS/ReEDS-2.0/issues/779) on the old repo)~~ not a blocker for this issue as it's out of date anyway
- [x] Transmission
  - [x] Interzonal costs and distances
  - [x] Planned transmission and HVDC scenarios
  - [ ] Once the rest of the pipeline works, add `GSw_ZoneSet` option combining low-data counties into neighboring counties
- [x] One-off files
  - [x] `inputs/fuelprices/cendivweights.csv`: 134-zone boundaries are currently baked into county-level values. Rerun the census division weights using the updated census division boundaries in [state_groups.csv](https://github.com/ReEDS-Model/ReEDS/blob/2cfcb1272eab66de4961a111b78904d5495b1c27/inputs/zones/state_groups.csv#L1)
- [ ] Overall structure
  - [x] Remove `aggregate_regions.py` (or refactor it to only handle the special-case technologies that still have 134-zone inputs) and do all aggregation using `county2zone.csv`
  - [ ] Consider going back to reading directly from the repo in most `inputs_processing` scripts instead of relying on intermediate files from `copy_files.py` (discuss)
- [ ] Documentation and ease of use
[...]

PR #95, opened 2026-05-15
Update interzonal transmission costs/distances and use directly calculated values

patrickbrown4:
## Summary

In the same vein as [2003](https://github.nrel.gov/ReEDS/ReEDS-2.0/pull/2003), this PR:
- Adapts the interzonal transmission costs, distances, and planned/optional additions for spatial flexibility as laid out in #16
- Uses the latest reV methods and assumptions to calculate the interzonal least-cost paths and costs

#### Least-cost paths

There are a few updates to the interzonal least-cost path procedure:
- The interzonal transmission costs have been out of sync with interconnection costs since [1836](https://github.nrel.gov/ReEDS/ReEDS-2.0/pull/1836). This PR brings them in sync, so both are now using costs from the MISO 2025 Transmission Cost Estimation Guide ([parent](https://www.misoenergy.org/planning/transmission-planning/mtep), [description](https://cdn.misoenergy.org/MISO%20Transmission%20Cost%20Estimation%20Guide%20for%20MTEP25337433.pdf), [data](https://cdn.misoenergy.org/MISO%20Transmission%20Cost%20Estimate%20Workbook%20for%20MTEP25547535.xlsx))
  - A new pair of data files from MISO, `inputs/transmission/conductor_{ac or dc}.csv`, provide lookup tables for converting from the assumed kV to MW capacity
- We used to assume 500 kV costs for all interzonal greenfield lines, regardless of location or scale (so tiny counties with low-voltage lines got to use 500 kV costs, which are cheaper on a USD/MW basis than lower-voltage lines). Here, we instead use the maximum AC voltage currently connecting each pair of zones, with a 138 kV floor. So interfaces that are crossed by higher-voltage lines get lower USD/MW expansion than interfaces crossed by lower-voltage lines.

#### Spatial flexibility

- The old future-capacity scenarios (which were specific to z134) have been removed. The scenarios used in NTP have been maintained (`planned_lines-NTP_{MT or P2P}.csv`), but now specify individual from/to endpoints
- Similarly, planned additions are now specified as individual endpoints
  - TransWest Express and SunZia have been added; see the descriptions in `inputs/transmission/README.md`
  - The offshore zone backbones are in `inputs/transmission/newlinks_offshore_backbone.csv`
  - The offshore zone radial connections are in `inputs/zones/{GSw_ZoneSet}/newlinks_offshore_radial.csv`
    - These need to be specified for each `GSw_ZoneSet` because the user might want to control which zone they make landfall in, particularly at county resolution
    - Offshore zones and county resolution are now compatible (they were not before)
- We keep the old single-bin transmission upgrade supply curve (TSC) file for greenfield 500 kV with GSw_ZoneSet=z134 as an example of the required file format when using TSC, but will eventually remove it when we add a full supply curve
  - The greenfield costs are put into this format automatically in `transmission.py`

### Additional changes 

- Features that we no longer need with the new format have been removed:
  - `inputs/shapefiles/transmission_endpoints` (now in `inputs/zones/{GSw_ZoneSet}/zonehash.csv`)
  - Transmission aggregation functionality in `copy_files.py`
  - `aggregate_regions.py` (#102 removed everything except for transmission aggregation from `aggregate_regions.py`; given the removal of transmission aggregation here, we no longer need the script)
  - Zone aggregation functionality in `reeds.io.get_zonemap()`
- `transmission.py` has been reorganized to make it importable
- `min_co2_spurline_miles` (= 20): moved from hard-coded in `transmission.py` to `scalars.csv`

- `GSw_TranScen`: Allowable options changed to `none` (default, meaning only interfaces with existing capacity can be expanded), `NTP_MT`, or `NTP_P2P`

### Issues resolved 

- #37
- Part of #16

I ran the full `cases_test.csv` suite and everything worked except for WECC_county, which failed in 2015, the same year in which it fails on the main branch.

Here's some background info and plots of the new costs: [20260515 - transmission costs distances.pptx](https://nrel.sharepoint.com/:p:/s/ReEDS/IQAZ3vdpteS_SpMibtiadIrWAUj4SKZAyAPSlHQE9t5_b80?e=hrxhIQ)

Compare reports are available in the same deck.
- There's more LCC and VSC in the default solution since we've added existing/planned lines that were originally left out
  - image
  - image
- The big line from Chicago to Michigan now goes to Indiana since we only include directly connected AC interfaces
  - image
- I made sure the WECC_county case runs through 2015

### Details to double-check

- [x] Make sure SunZia and TransWestExpress show up in the solution
- [x] Add the SunZia cost/distance to keep the WECC_county case no more broken than on main
- [x] Update documentation text and figures as laid out in #37
- [x] Decide on MISO default (rename from 'default' to 'miso') or ACSR by default for new conductors
- [x] Open issue on updating environment
  
  - [x] Dollar year recorded and converted to 2004$ for GAMS
  - [x] Units are specified
  - [x] Preprocessing steps have been documented and committed to [ReEDS_Input_Processing](https://github.com/ReEDS-Model/ReEDS_Input_Processing)
  - [ ] ~~New large data files handled with .h5 instead of .csv~~ <- No, because we will add to the costs/distances as new zone geometries are added

No


inputs/shapefiles/transmission_endpoints/transmission_endpoints.cpg
patrickbrown4, line 1: Removed because the versions that match the interzonal paths are in the `zonehash.csv` files; these no longer match

inputs/transmission/conductor_ac_acss.csv
patrickbrown4, line 1: Included so we can avoid adding more 'MW' or '$/MWmile' columns to the already-large `transmission_cost_distance.csv`; instead we just look it up from here using the `voltage` column

inputs/transmission/planned_lines-NTP_MT.csv
patrickbrown4, line 1: The "name" column doesn't actually matter but including it here in case it's useful for backwards compatibility

docs/source/figs/docs/supplycurve-cost.png
patrickbrown4, line 1: not directly related but was deferred earlier and it's on the list for #37

inputs/zones/z69/b2b.csv
patrickbrown4, line 1: bugfix

reeds/core/setup/b_inputs.gms
```diff
@@ -3918,7 +3918,7 @@ parameter cost_prod(i,v,r,t)                  "--$/metric ton/hr-- cost or benef
 scalar    h2_combustion_intensity              "--metric tons/MMBtu-- amount of hydrogen consumed per MMBtu of H2-Combustion fuel consumption" ;
 
 parameter pipeline_distance(r,rr) "--miles-- distance between all adjacent BA centroids for pipeline investments" ;
-pipeline_distance(r,rr) = distance(r,rr,"AC") ;
+pipeline_distance(r,rr) = distance(r,rr,"VSC") ;
```
patrickbrown4, line 3450: VSC routes are defined for all adjacent zones (which is what we want); AC are not

reeds/inputs.py
```diff
@@ -11,52 +11,41 @@
...
 from pathlib import Path
 from warnings import warn
+from typing import Literal
 sys.path.append(str(Path(__file__).parent.parent))
 import reeds
 from reeds.input_processing import mcs_sampler
 
 ### Functions
-def parse_regions(case_or_string, case=None):
+def parse_regions(case:str|Path|None=None, **kwargs) -> list:
```
patrickbrown4, line 20: This more direct version is needed in updated `reeds.io.get_zonemap()`

reeds/io.py
```diff
@@ -189,193 +251,52 @@ def get_countymap(select_counties=None, exclude_water_areas=False):
...
+    offshore_zones = (
+        gpd.read_file(
+            os.path.join(reeds_path, 'inputs', 'shapefiles', 'offshore_zones.gpkg')
+        ).set_index('zone').to_crs(crs).drop(columns=['zone_old'], errors='ignore')
+        .rename(columns={'node_latitude':'node_lat', 'node_longitude':'node_lon'})
+    )
+    return offshore_zones
+
+
+def get_zonemap(case=None, exclude_water_areas=False, crs='ESRI:102008', **kwargs):
```
patrickbrown4, line 266: Updated to use the new zone node lat/lons and avoid need for county/mixed-specific processing

cases.csv
```diff
@@ -329,9 +330,9 @@ GSw_TransHurdleLevel2,Apply hurdle rates from inputs/transmission/cost_hurdle_in
 GSw_TransHurdleRate,Turn on (1) or off (0) hurdle rates between hierarchy regions as defined by GSw_TransHurdleLevel1/GSw_TransHurdleLevel2 and inputs/transmission/cost_hurdle_intra.csv,0; 1,0,
 GSw_TransNetworkSource,Network source for the initial transmission network used in the model,NARIS2024,NARIS2024,
 GSw_TransRestrict,hierarchy level within which to allow new transmission lines,r; nercr; transreg; transgrp; cendiv; st; interconnect; country; usda_region,country,
-GSw_TransScen,Transmission expansion scenario (options in inputs/transmission),N/A,default,
-GSw_TransSquiggliness,Transmission distance multiplier to approximate the squigglier paths followed by actual lines (only applied to NEW lines),float,1,
-GSw_TransUpgradeMethod,Method for calculating AC transmission upgrade costs,500kv,500kv,
+GSw_TransScen,Additional candidate interfaces for transmission expansion between zones,^(none|NTP_(MT|P2P))$,none,
+GSw_TransSquigglinessMin,Minimum squiggliness (straight-line length multiplier) to apply for interzonal transmission; the default value is from the MISO Transmission Cost Estimation Guide (https://www.misoenergy.org/planning/transmission-planning/mtep),float,1.3,
```
patrickbrown4, line 335: The new routes are even straighter than the old, so I think now is a good time for this change (and it keeps us more aligned with MISO assumptions)

reeds/input_processing/transmission.py
```diff
@@ -1,65 +1,281 @@
...
+    return shapely.LineString([[row.start_lon, row.start_lat], [row.end_lon, row.end_lat]])
+
+
+def get_interface_params(case, **kwargs):
+    """Get cost and distance for every interface that might be used in this run"""
+    sw = reeds.io.get_switches(case, **kwargs)
+    scalars = reeds.io.get_scalars(case)
+
+    interface_params = reeds.inputs.get_distances(case)
+    ## Apply the minimum-squiggliness factor
```
patrickbrown4, line 74: This part is new; most of the rest is just reorganized

reeds/io.py
```diff
@@ -189,193 +251,52 @@ def get_countymap(select_counties=None, exclude_water_areas=False):
...
         )
-    else:
-        agglevel_variables = {'lvl': 'ba',
-                              'agglevel': 'ba',
-                              }
-
-    # Mixed resolution procedure
-    if agglevel_variables['lvl'] == 'mult':
-        ### Model zones
-        dfba = gpd.read_file(os.path.join(reeds_path, 'inputs', 'shapefiles', 'US_PCA'))
```
patrickbrown4, line 216: With this change, we're no longer using this 134-zone shapefile directly. We can wait and remove it in a future PR though, since there might be upstream or downstream processes relying on it.

reeds/input_processing/transmission.py
```diff
@@ -70,28 +286,18 @@ def get_trancap_init(case, networksource='NARIS2024', level='r'):
         [['r', 'rr', 'MW_forward', 'MW_reverse']]
         .assign(trtype='AC')
     )
-    ### DEPRECATED: p19 is islanded with NARIS transmission data, so connect it manually
```
patrickbrown4, line 72: Dropped because we only calculate AC distances/costs for interfaces crossed by existing AC lines and this one is not crossed by any lines. So I think p19 is now islanded, but I didn't have any issues with extra ReEDS/PRAS iterations there (and we're now only ~2 PRs away from switching to 90 zones).

reeds/io.py
```diff
@@ -164,6 +184,48 @@ def get_county2zone(
...
+    Args:
+        case (str, Path, or None): Path to a ReEDS case.
+            If None, uses the default GSw_ZoneSet from cases.csv.
+        crs (str): Coordinate reference system
+        **kwargs: ReEDS switch:value pairs (overrides case argument)
+    """
+    dfcounty = reeds.spatial.get_map('county', source='tiger', crs=crs)
+    county2zone = reeds.io.get_county2zone(case, **kwargs)
+    dfcounty['r'] = county2zone
+    dfzones = dfcounty.dissolve('r')
```
patrickbrown4, line 221: Because this zone map is built from the county map, which is more detailed than the old US_PCA map, all the downstream calculations and plots that use the maps (accessed via `reeds.io.get_dfmap()` or `maps.gpkg`) are slowed down. There is a [`.simplify_coverage()`](https://geopandas.org/en/latest/docs/reference/api/geopandas.GeoSeries.simplify_coverage.html) function in geopandas 1.1 that we can use to simplify the geometry and speed them all up again, but it requires an update to the conda environment.

Elaine is also working on a `gdxpds` update; my plan is to go ahead with the current implementation for now, and then once the new `gdxpds` version is ready, do a holistic environment update to geopandas 1.1 / pandas 3 / python 3.14 / etc., so everyone only has to update their environment once.

wesleyjcole: One other package that might be worth adding in an update is the `highs` package to give access to the HiGHS solver for things like representative period selection.

docs/source/model_documentation.md
```diff
@@ -1812,32 +1814,18 @@ Substation POIs incur a substation upgrade cost of \$15/kW; transmission line PO
 
 Network reinforcement represents upgrades to the existing transmission network required to avoid congestion when moving power from the POI for a new generator to load centers.
 It is intended to represent the costs associated with interconnection queues, which represent a major bottleneck for the deployment of new wind and solar in the United States. {cite}`gormanGridConnectionBarriers2025`.
-Network reinforcement costs are approximated by tracing a path along existing transmission lines from each wind/solar POI to each zone "center" within the same state;
-the zone center is usually taken as the largest population center in the model zone but is sometimes (for zones without large urban centers) assigned to a high-voltage substation within the zone.[^ref35]
-A cost for each reinforcement route is calculated using the cost surface described above, with capital expenditure (CAPEX) costs multiplied by 50% to approximate the lower cost for reconductoring compared to greenfield transmission construction.
+Network reinforcement costs are approximated by tracing a path along existing transmission lines from each wind/solar POI to a nearby urban center.
```
wesleyjcole, line 1822: Can you quantify "nearby" and "urban center"?  I can't remember the specifics, but this seems like a good place to document those.

inputs/transmission/conductor_ac_acsr.csv
```diff
@@ -0,0 +1,9 @@
+voltage_kv,kcmil,conductor_type,conductor_quantity,amp
```
wesleyjcole, line 1: What is kcmil?

inputs/transmission/hvdc_lines.csv
```diff
@@ -1,8 +1,8 @@
-name,MW,from_lon,from_lat,to_lon,to_lat
-Pacific DC Intertie,3100,-121.11416667,45.59416667,-118.48916667,34.31083333
-Square Butte,500,-101.19583333,47.07166667,-92.29416667,46.77361111
-CU,1000,-101.15638889,47.37333333,-93.81,45.11111111
-Path 27,2400,-112.58083333,39.50055556,-117.43722222,34.55111111
-Cross Sound Cable,330,-72.90222222,41.28666667,-72.8675,40.95916667
-Neptune Cable,660,-73.55111111,40.76055556,-74.35308333,40.47371667
-Trans Bay Cable,400,-121.89666667,38.03083333,-122.38583333,37.75472222
+name,MW,year_online,trtype,certain,from_lon,from_lat,to_lon,to_lat
```
wesleyjcole: Is `year_online` the number of years until it is online?  If so, it might be worth updating the name, as I was expecting something like "2028" for the value rather than "0".

inputs/transmission/planned_lines-baseline.csv
```diff
@@ -0,0 +1,3 @@
+name,MW,year_online,trtype,certain,from_lon,from_lat,to_lon,to_lat
+TransWestExpress,3000,2032,VSC,1,-107.176147,41.747242,-112.590781,39.541825
```
wesleyjcole: This one has `year_online` as what I would expect, so I think I'm not understanding this column correcting, or what a zero value for this column means.

inputs/transmission/README.md
wesleyjcole, line 1: This could be a good place to document what year_online = 0 means.

inputs/transmission/README.md
```diff
@@ -6,6 +6,12 @@
   The `osm_id` column gives the OpenStreetMap ID of the converter; for example, the "Miles City, MT" converter (with `osm_id = 137835349`) can be found at <https://www.openstreetmap.org/way/137835349>.
   This file is used to validate the interface-level B2B capacity for different spatial resolutions stored at `inputs/zones/{GSw_ZoneSet}/b2b.csv`.
 
+- `conductor_(ac|dc)*.csv`: Conductor and power rating assumptions for AC/DC transmission lines as a function of voltage from the MISO 2025 Transmission Cost Estimation Guide ([parent page](https://www.misoenergy.org/planning/transmission-planning/mtep), [description](https://cdn.misoenergy.org/MISO%20Transmission%20Cost%20Estimation%20Guide%20for%20MTEP25337433.pdf), [data workbook](https://cdn.misoenergy.org/MISO%20Transmission%20Cost%20Estimate%20Workbook%20for%20MTEP25547535.xlsx))
+  - `conductor_ac_default.csv` and `conductor_dc.csv` use the conductors specified by the MISO guide.
```
wesleyjcole: There is no `conductor_ac_default.csv`.

inputs/scalars.csv
```diff
@@ -53,6 +53,7 @@ interconnection_start,2025,"year that interconnection queue data starts"
 inverter_efficiency,0.98,"--fraction-- one-way efficiency of AC-DC or DC-AC conversion"
 lb_per_tonne,2204.62,"--lb/metric ton-- pound per metric ton"
 maxage_default,100,"--years-- maximum age for technogies (default if no entry in maxage.csv)"
+min_co2_spurline_miles,20,"--miles-- minimum distance for a CO2 spur line (used to provide a floor for pipeline distances in r_cs_distance)"
```
wesleyjcole, line 56: Would you note this change in the PR text?

reeds/input_processing/copy_files.py
```diff
@@ -645,67 +636,7 @@ def subset_to_valid_regions(
...
             filename=filename
         )
-
-
-        # Transmission files need to be filtered differently to allow interfaces between BA and county resolution regions
-        transmission_files = [
-            'transmission_cost_ac.csv',
-            'transmission_cost_dc.csv',
-            'transmission_distance.csv',
-        ]
```
wesleyjcole, line 655: I love that this special handling is removed.

reeds/input_processing/transmission.py
```diff
@@ -1,64 +1,281 @@
...
+    sw = reeds.io.get_switches(case, **kwargs)
+    scalars = reeds.io.get_scalars(case)
+
+    interface_params = reeds.inputs.get_distances(case)
+    ## Apply the minimum-squiggliness factor
+    interface_params['geometry'] = interface_params.apply(_make_line, axis=1)
+    interface_params = gpd.GeoDataFrame(interface_params, crs='EPSG:4326')
+    interface_params['straight_miles'] = (
+        interface_params.geometry.to_crs('EPSG:5070').length
+        / 1609.34
```
wesleyjcole: Is this a units conversion?  Would you add a comment specifying?

reeds/input_processing/transmission.py
```diff
@@ -129,13 +335,152 @@ def get_trancap_init(case, networksource='NARIS2024', level='r'):
...
+    ### TEMPORARY 20260402: Drop county interfaces with no distance/cost
+    if (level == 'r') and (sw.GSw_RegionResolution in ['county', 'mixed']):
+        transmission_line_fom = get_transmission_fom(case, interface_params)
+        indices = ['r', 'rr', 'trtype']
+        drop = (
+            dfout
+            .merge(transmission_line_fom.reset_index(), on=indices, how='left')
+        )
+        drop = list(drop.loc[drop.USDperMWyear.isnull(), indices].itertuples(index=False))
+        dfout = dfout.set_index(indices).drop(drop).reset_index()
```
wesleyjcole, line 370: Is this what is making WECC county runs fail (mentioned in your PR text)?  Or does this make it not fail?

Yunzhi is nearly done with the WECC county bug fix, so it would be nice to not have it immediately broken again if that can be avoided, even if that meant a temporary patched solution until new reV data is available.

runreeds.py
```diff
@@ -293,8 +293,8 @@ def check_compatibility(sw):
 
     if sw['GSw_RegionResolution'] in ['county','mixed']:
         err_switch_configs = []
-        if int(sw['GSw_OffshoreZones']):
-            err_switch_configs.append('GSw_OffshoreZones=1')
+        if int(sw['GSw_CO2_Detail']):
+            err_switch_configs.append('GSw_CO2_Detail=1')
```
wesleyjcole: What is this change?

patrickbrown4: Offshore zones now work with county resolution, but I wasn't sure if CO2 pipelines would:

- The CO2 pipeline cost piggybacks off the transmission cost
- We used to calculate inter-county costs/distances for all counties within 50 miles of each other, but now we only do it for counties connected by transmission; that's fine for transmission but I wasn't sure about pipelines

I went ahead and tried the "WY_county" case with `GSw_CO2_Detail = 1` and it worked, so I think we're somewhat safe. So I've now removed this check.

inputs/transmission/conductor_ac_acsr.csv
patrickbrown4: It's a silly U.S. unit but it's the standard measurement of wire gauge: <https://en.wikipedia.org/wiki/Circular_mil>

reeds/input_processing/transmission.py
patrickbrown4: No, it's a bit different; this is to keep `GSw_ZoneSet = z3109` (county-level, which includes counties with no transmission links to neighbors) working until we finish reformatting the spatial pipeline, which will allow use of `GSw_ZoneSet = z2972` (which aggregates those counties into their neighbors and avoids making islanded zones).

The missing SunZia line at county resolution is caught in `check_inputs.py`. But I'll add it before merging.

inputs/transmission/hvdc_lines.csv
patrickbrown4: Added a description to `inputs/transmission/README.md` and changed these to their actual historical online year

docs/source/model_documentation.md
patrickbrown4: Added some lines to describe the current approach.

reeds/input_processing/transmission.py
patrickbrown4: The SunZia cost/distance has been added, so the WECC_county case now runs through 2015, the same as on main.

wesleyjcole (approved):
Thanks for the updated explanations and documentation.

reeds/inputs.py
```diff
@@ -592,18 +593,46 @@ def get_itls(case=None, level:str='r', errors='raise', **kwargs) -> pd.DataFrame
...
+    else:
+        ## Include any connections to offshore zones for now
+        dfoffshore = reeds.io.get_offshore_zones()
+        offshore_zones = dfoffshore.index.values.tolist()
+        dfout = dfin.loc[
+            (
+                dfin.start.isin(zonehash.tolist()+offshore_zones)
+                & dfin.end.isin(zonehash.tolist()+offshore_zones)
+            )
+            | (dfin.start.isin(offshore_zones) & (dfin.end.isin(offshore_zones)))
```
kodiobika: Unless I'm looking at this incorrectly, wouldn't the second clause of the 'or' statement being true necessitate the first clause being true (so we can get rid of the second clause)?

reeds/inputs.py
```diff
@@ -613,34 +642,55 @@ def get_itls(case=None, level:str='r', errors='raise', **kwargs) -> pd.DataFrame
...
+        case = None
+        level = 'r'
+        kwargs = {}
+        errors = 'raise'
+    """
+    return get_interface_data(
+        case=case,
+        datafile='itl_NARIS.csv',
+        level=level,
+        interaces=True,
```
kodiobika:
```suggestion
        interfaces=True,
```

reeds/inputs.py
```diff
@@ -613,34 +642,55 @@ def get_itls(case=None, level:str='r', errors='raise', **kwargs) -> pd.DataFrame
...
+    )
+    return distances
 
 
 def _make_line(row):
     return shapely.LineString([[row.from_lon, row.from_lat], [row.to_lon, row.to_lat]])
 
 
-def get_hvdc_lines():
+def get_hvdc_lines(filename='hvdc_existing.csv'):
```
kodiobika, line 688: No action required here, but this got me wondering - is there a well-defined difference between `reeds/inputs.py` and `reeds/io.py`? My instinct would've been to put something like this in the latter but I'm not actually sure which, if any, makes more sense in this case

reeds/io.py
```diff
@@ -191,193 +253,52 @@ def get_countymap(select_counties=None, exclude_water_areas=False):
...
+    dfba = dfzones.merge(hierarchy, left_index=True, right_on='r').set_index('r')
+    ## Record centroid locations for plot labels
+    dfba['centroid_x'] = dfba.geometry.centroid.x
+    dfba['centroid_y'] = dfba.geometry.centroid.y
 
     return dfba
 
 
-def get_dfmap(case=None, levels=None, exclude_water_areas=False):
+def get_dfmap(case=None, levels=None, exclude_water_areas=True):
```
kodiobika, line 301: Just checking - the fact that all the instances throughout the code where `get_dfmap` is currently being used and `exclude_water_areas` isn't specified will now exclude water areas (where they were previously including them) is intentional right?

reeds/input_processing/copy_files.py
```diff
@@ -645,67 +636,7 @@ def subset_to_valid_regions(
...
-                df_county = df_county.drop_duplicates(subset=['region_pair','tscbin'], keep='first')
-                df_county = df_county.drop(columns=['region_pair'])
-            else:
-                df_county = df_county.groupby([tx_region_col,'rr'] + fix_cols).sum().reset_index()
-
-            df_county['interface'] = df_county[tx_region_col] + '||'+df_county['rr']
-            df_county.reset_index(drop=True,inplace=True)
-
-        else:
+        if True:
```
kodiobika: Can we remove this (and outdent the code below)?

reeds/input_processing/copy_files.py
kodiobika: Can you remove the special case here too?

inputs/transmission/transmission_cost_distance.csv
```diff
@@ -0,0 +1,9313 @@
+start,end,polarity,voltage,cost_MUSD,length_miles
```
kodiobika, line 1: nit: It'd be nice for consistency to specify the voltage units too, but no need if it's too much of a hassle at this point

inputs/transmission/transmission_cost_distance_lines.csv
```diff
@@ -0,0 +1,499 @@
+name,from_lon,from_lat,to_lon,to_lat,polarity,voltage,cost_MUSD,length_miles
```
kodiobika, line 1: nit: Ditto on voltage units ('voltage' -> 'voltage_kv')

reeds/input_processing/transmission.py
```diff
@@ -1,64 +1,282 @@
...
+    """Get cost and distance for every interface that might be used in this run"""
+    sw = reeds.io.get_switches(case, **kwargs)
+    scalars = reeds.io.get_scalars(case)
+
+    interface_params = reeds.inputs.get_distances(case)
+    ## Apply the minimum-squiggliness factor
+    interface_params['geometry'] = interface_params.apply(_make_line, axis=1)
+    interface_params = gpd.GeoDataFrame(interface_params, crs='EPSG:4326')
+    interface_params['straight_miles'] = (
+        interface_params.geometry.to_crs('EPSG:5070').length
```
kodiobika, line 78: Just curious, why 'EPSG:5070' here instead of the usual 'ESRI:102008'?

reeds/input_processing/transmission.py
```diff
@@ -129,13 +336,152 @@ def get_trancap_init(case, networksource='NARIS2024', level='r'):
...
+        interface_params.trtype=='AC',
+        ['r', 'rr', f'USD{dollar_year}perMW']
+    ].copy()
+    transmission_cost_ac = (
+        transmission_cost_ac
+        .rename(columns={f'USD{dollar_year}perMW':f'USD{dollar_year}perMW_forward'})
+        .assign(**{f'USD{dollar_year}perMW_reverse':transmission_cost_ac[f'USD{dollar_year}perMW']})
+        .assign(tscbin='t0')
+        .assign(**{f'binwidth_USD{dollar_year}':1e12})
+        .rename(columns={})
```
kodiobika: Can this line be deleted?
```suggestion
```

kodiobika (approved):
Thanks for doing this! LGTM

reeds/inputs.py
patrickbrown4: ugh I forgot **kwargs allows this kind of thing, thanks

reeds/input_processing/copy_files.py
patrickbrown4: Yeah I just left it to make the review marginally easier but have removed now, thanks for the reminder

reeds/input_processing/transmission.py
patrickbrown4: They're both equal-area projections, but EPSG:5070 is zoomed in more on the contiguous US, while ESRI:102008 is for all of North America (there are examples of the two at https://gis.stackexchange.com/a/457667). That means that within the CONUS, there's a bit less distortion in distances and areas using EPSG:5070 than ESRI:102008. (Eventually it'd be nice to change everything to EPSG:5070 and make it controlled by a global switch.)

inputs/transmission/transmission_cost_distance.csv
patrickbrown4: Yeah fair; I kept these because they're the column names for inputs to and outputs from the [reV Routing (reVRt) model](https://github.com/NatLabRockies/reVRt). I've added notes on the units to `inputs/transmission/README.md`; think that's enough for now?

reeds/inputs.py
patrickbrown4: My very rough notion has been to use `inputs.py` for things that are "further up" in the pipeline (closer to the direct `inputs/` data) or that require more processing, and `io.py` for interacting with intermediate files from the `{case}/inputs_case` folder. I'm sure I haven't been super consistent though (`inputs.py` is newer and I didn't port everything to it from `io.py` that would fit those guidelines.) I was also worried about `io.py` becoming too huge if we use it as a catch-all for everything, but I'm not sure how important that is.

Maybe we could do a standalone cleanup/organizational PR after we're done with all the spatial stuff? We'll need some more layers of organization anyway as we start moving `b_inputs.gms` into Python.

reeds/io.py
patrickbrown4: `get_dfmap()` until now has used the 134-zone shapefile for everything except county zones, and that shapefile excludes water areas. So I think excluding them by default is most consistent, but I didn't actually go through and check all the areas where water areas might matter.

Do you remember offhand the places where including water areas is intended and important?

kodiobika: Ah got it, thanks that SGTM. No instances come to mind (and from a cursory scan it looks like there are none) so I think we're good

inputs/transmission/transmission_cost_distance.csv
kodiobika: Makes sense, yup that SGTM

PR #99, opened 2026-05-20
Add multi-metrics for RA

abdelrahman-ayad:
## Summary

- This PR expands the use of Resource Adequacy metrics to use a multi-metrics approach. Users can select one or more metric (EUE, LOLH, NEUE, or a combination) with a corresponding threshold to be used in stress periods selection. 
- An upcoming PR will extend the available stress metrics to assess duration, frequency, and magnitude of shortfalls. 

### Implementation notes 
- `EUE` is no longer used for determining the shoulder periods in function `get_houlder_periods()`, in case `EUE` is not defined as a stress metric
- If `EUE` is not included as a stress metric, PRAS-informed PRM update is blocked, and fixed-increment update is used, similar to setting `GSW_PRM_UpdateMethod=1`

- `GSw_PRM_StressThreshold` -> `GSw_PRM_StressThresholdMetrics`: This switch now defines the `/` delimited stress metric switches to be used for evaluation. For each `Metric`, a dedicated switch is added to define the `HierarchyLevel`, `Criterion`, and the `PeriodAggMethod`. Current tested metrics are `EUE`, `LOLH`, and `NEUE`. 

- `GSw_PRM_StressThresholdEUE`: Switch to define the `EUE` stress metric hierarchy, threshold, and PeriodAggMethod.
- `GSw_PRM_StressThresholdLOLH`: Switch to define the `LOLH` stress metric hierarchy, threshold, and PeriodAggMethod.
- `GSw_PRM_StressThresholdNEUE`: Switch to define the `NEUE` stress metric hierarchy, threshold, and PeriodAggMethod.

| Metric | Switch | Definition | RA dimension | Example |
|--------|--------|--------|--------|--------|
|`LOLH`|`GSw_PRM_StressThresholdLOLH`|Expected (average) count of event-hours/year experiencing shortfall. It is expressed in terms of event-hours|Frequency|`transgrp_2.4_sum`: The `LOLH` threshold is set at 2.4 events-hour/year, aggregated using `sum` over `transgrp`|
|`EUE`|`GSw_PRM_StressThresholdEUE`|Expected (average) total energy shortfall over the study period, expressed in energy units (MWh per year)|Energy|`transgrp_10000_sum`: The `EUE` threshold is set at 10000 MWh over the study period, aggregated using `sum` over `transgrp`|
|`NEUE`|`GSw_PRM_StressThresholdNEUE`| Normalized expected (average) total energy shortfall over the study period expressed in parts-per-million (ppm)|Energy|`transgrp_1_sum`: The `NEUE` threshold is set at 1 ppm aggregated using `sum` over `transgrp`|

Slides deck reviewing Resource Adequacy multi-metrics: 
[RA multi metrics slides deck](https://nrel.sharepoint.com/:p:/r/sites/AdvancedTransmissionModelingATB/Shared%20Documents/General/RA/MultiMetricRA/[20260331_multi_metric_deliverable.pptx](https://nrel.sharepoint.com/:p:/r/sites/AdvancedTransmissionModelingATB/Shared%20Documents/General/RA/MultiMetricRA/20260331_multi_metric_deliverable.pptx?d=w4164c38e4e0c4179939e94ee2a1106af&csf=1&web=1&e=KLM4PX&isSPOFile=1&xsdata=MDV8MDJ8fGExYTE4OGQ4ZjdlYjQ5MDUzM2RhMDhkZTdmYTMyYWE1fGEwZjI5ZDdlMjhjZDRmNTQ4NDQyNzg4NWFlZTdjMDgwfDB8MHw2MzkwODg1MzU5MDUwNTU3MDZ8VW5rbm93bnxWR1ZoYlhOVFpXTjFjbWwwZVZObGNuWnBZMlY4ZXlKRFFTSTZJbFJsWVcxelgwRlVVRk5sY25acFkyVmZVMUJQVEU5R0lpd2lWaUk2SWpBdU1DNHdNREF3SWl3aVVDSTZJbGRwYmpNeUlpd2lRVTRpT2lKUGRHaGxjaUlzSWxkVUlqb3hNWDA9fDF8TDJOb1lYUnpMekU1T20xbFpYUnBibWRmVG1wbk1WcEVWWGROVjBsMFdrUk5NMDU1TURCUFJHZDRURlJyTVU1NlZYUlBWRnBvV1RKUk0xbHRVVFZQUjA1cVFIUm9jbVZoWkM1Mk1pOXRaWE56WVdkbGN5OHhOemN6TWpVMk56ZzRPRGcwfGVjZWU1MjFmYzhiZDRiYzYzM2RhMDhkZTdmYTMyYWE1fDA3ZjFiODA3ZjUyYzQ2ODU5OWJkZDgwMjY4YjlhM2Q2&sdata=b0x3WUhEZmZrTUlBK2IrQTJpL2VtUnBUVDVOWlU3cVJKVGdjemZ6MDU5UT0%3D&ovuser=a0f29d7e-28cd-4f54-8442-7885aee7c080%2Caayad%40nrel.gov)?d=w4164c38e4e0c4179939e94ee2a1106af&csf=1&web=1&e=KLM4PX)

[Comparison report](https://nrel-my.sharepoint.com/:p:/g/personal/aayad_nrel_gov/IQC24ydLSsrsSLCivJ6SHd57AajL6by867MiV3RnPtNWqHQ?e=AaJa9L)

# To check
- [x] Pacific (EUE, LOLH, NEUE, EUE/NEUE/LOLH)
- [x] USA_defaults (EUE, LOLH, NEUE, EUE/NEUE/LOLH)

### Details to double-check

- I used Claude to plan the changes in `runreeds.py` and `stress_periods.py`


cases.csv
```diff
@@ -261,14 +261,17 @@ GSw_PRM_NetImportLimit,Turn on (1) or off (0) the eq_firm_transfer_limit constra
...
 GSw_PRM_StressStorageCutoff,"How to select ""shoulder"" stress periods giving storage time to recharge before/after high-unserved-energy periods. Two-part switch separated by _. The first argument is ""EUE"" or ""capacity"" or ""absolute"". If ""EUE"" the second argument specifies a PRAS storage headspace / EUE threshold [fraction]; if ""cap"" it specifies a headspace / storage capacity threshold [fraction]; if ""abs"" it specifies absolute number of periods before/after [integer]. Turned off if set to ""off"".",N/A,EUE_0.1,
-GSw_PRM_StressThreshold,/-delimited list of annual NEUE level [ppm] above which to re-solve the latest model year with new stress periods; formulated as HierarchyLevel_NEUEppm_StressMetric_PeriodAggMethod where HierarchyLevel is a column in hierarchy.csv; NEUEppm is normalized expected unserved energy in parts per million; StressMetric is EUE or NEUE (only used in period selection); PeriodAggMethod is 'sum' or 'max' over the hours in each period (only used in period selection) (see README.md for detailed notes),N/A,transgrp_1_EUE_sum,
+GSw_PRM_StressThresholdMetrics,/-delimited list of metrics for identifying stress periods,N/A,EUE,
```
patrickbrown4: 1 ppm NEUE is the current default on the main branch, so we should stick with that here.

It would also be helpful to list the supported options in the comment.
```suggestion
GSw_PRM_StressThresholdMetrics,/-delimited list of metrics for identifying stress periods (supported options are: NEUE EUE LOLE),N/A,NEUE,
```

cases.csv
```diff
@@ -261,14 +261,17 @@ GSw_PRM_NetImportLimit,Turn on (1) or off (0) the eq_firm_transfer_limit constra
...
+GSw_PRM_StressThresholdLOLE,LOLE threshold above which to re-solve the latest model year with new stress periods; formulated as HierarchyLevel_nLOLE_StressMetric_PeriodAggMethod where HierarchyLevel is a column in hierarchy.csv; nLOLE is number of loss of load events; StressMetric is LOLE; PeriodAggMethod is 'sum' or 'max' over the hours in each period (only used in period selection) (see README.md for detailed notes),N/A,transgrp_36_LOLE_sum,
```
patrickbrown4: The comment says "events", but the PR text and `run_pras.jl` say "event-hours". Assuming it is actually event-hours, can you:
- [ ] Change the name here and throughout the code to LOLH?

It's also not ideal to bake the number of weather years into the threshold, because then users would need to remember to adjust two things if they change the number of weather years.
- [ ] Can you instead formulate it as event-hours per year, with a default of 2.4? (Or if it were LOLE and not LOLH, 0.1?)
- [ ] Add the units to the comment: [hours/year] ([events/year] if it were LOLE)

patrickbrown4: Also, is the third element (StressMetric) required? If only one option is supported, it can be dropped. (As noted below, I think you can drop the third element for all of the new switches.)

cases.csv
```diff
@@ -261,14 +261,17 @@ GSw_PRM_NetImportLimit,Turn on (1) or off (0) the eq_firm_transfer_limit constra
...
+GSw_PRM_StressThresholdEUE,/Annual EUE level [MWh] threshold above which to re-solve the latest model year with new stress periods; formulated as HierarchyLevel_EUE_StressMetric_PeriodAggMethod where HierarchyLevel is a column in hierarchy.csv; EUE is expected unserved energy; StressMetric EUE (only used in period selection); PeriodAggMethod is 'sum' or 'max' over the hours in each period (only used in period selection) (see README.md for detailed notes),N/A,transgrp_10000_EUE_sum,
```
patrickbrown4: Remove leading `/` in the comment (here and in the next line)

cases.csv
```diff
@@ -261,14 +261,17 @@ GSw_PRM_NetImportLimit,Turn on (1) or off (0) the eq_firm_transfer_limit constra
...
+GSw_PRM_StressThresholdNEUE,/Annual NEUE level [ppm] threshold above which to re-solve the latest model year with new stress periods; formulated as HierarchyLevel_NEUEppm_StressMetric_PeriodAggMethod where HierarchyLevel is a column in hierarchy.csv; NEUEppm is normalized expected unserved energy in parts per million; StressMetric is EUE or NEUE (only used in period selection); PeriodAggMethod is 'sum' or 'max' over the hours in each period (only used in period selection) (see README.md for detailed notes),N/A,transgrp_1_NEUE_sum,
```
patrickbrown4: To simplify things for the user, it seems like we could just drop the StressMetric part of the input and define it in `stress_periods.py`. It seems like you'd always use EUE for stress period selection when using an EUE- or NEUE-based threshold and LOLH when using an LOLH-based threshold. (We've never actually used StressMetric=NEUE so I think it's fine to drop support for that.)

runreeds.py
```diff
@@ -319,33 +319,53 @@ def check_compatibility(sw):
...
-                f"The combination of GSw_PRM_StressModel={sw['GSw_PRM_StressModel']} and "
-                f"stress_metric={stress_metric} is not supported."
-            )
-            raise NotImplementedError(err)
+    ### Check that the stress metrics specified in GSw_PRM_StressThresholdMetrics
+    # have corresponding GSw_PRM_StressThreshold{metric} switches
+    stressThresholdMetricSwitches = [s.split('GSw_PRM_StressThreshold')[1]
+                             for s in sw.keys() if s.startswith('GSw_PRM_StressThreshold')
+                             and not s.endswith('Metrics')
+                             ]
```
patrickbrown4: Parsing all switch names makes me nervous (it's not something we do elsewhere) so I'd just be explicit (we're in the `check_compatibility()` function so I think it's ok to be explicit).

(Normally we'd put the check in the `Choices` column but it would be a pain to write a regex that handles all orders, so it's ok to just do it here instead.)

```suggestion
    stressThresholdMetricSwitches = ['NEUE', 'EUE', 'LOLH']
```

runreeds.py
```diff
@@ -319,33 +319,53 @@ def check_compatibility(sw):
...
+    for metric in stressTresholdMetrics:
+        if metric.upper() not in stressThresholdMetricSwitches:
+            raise NotImplementedError(f"GSw_PRM_StressThreshold{metric} is not specified as a switch.")
+    
+    # # Threshold metric is not added but specified as a switch
+    # for metric in stressThresholdMetricSwitches:
+    #     if metric.upper() not in stressTresholdMetrics:
+    #         raise NotImplementedError(f"GSw_PRM_StressThreshold{metric} is specified as a switch,"
+    #             f"but not added to GSw_PRM_StressThresholdMetrics list {stressTresholdMetrics}")
+
```
```suggestion
```

runreeds.py
```diff
@@ -319,33 +319,53 @@ def check_compatibility(sw):
...
+    # for metric in stressThresholdMetricSwitches:
+    #     if metric.upper() not in stressTresholdMetrics:
+    #         raise NotImplementedError(f"GSw_PRM_StressThreshold{metric} is specified as a switch,"
+    #             f"but not added to GSw_PRM_StressThresholdMetrics list {stressTresholdMetrics}")
+        
+    for metric in stressThresholdMetricSwitches:
+        for threshold in sw[f'GSw_PRM_StressThreshold{metric}'].split('/'):
+            ## Example: threshold = 'transgrp_10_EUE_sum'
+            allowed_levels = ['country','interconnect','nercr','transreg','transgrp','st','r']
+            (hierarchy_level, ppm, stress_metric, period_agg_method) = threshold.split('_')
```
patrickbrown4: It's no longer just `ppm`
```suggestion
            (hierarchy_level, value, stress_metric, period_agg_method) = threshold.split('_')
```

runreeds.py
```diff
@@ -319,33 +319,53 @@ def check_compatibility(sw):
...
+                    f"GSw_PRM_StressThreshold{stress_metric}: level={hierarchy_level} but must be in:\n"
+                    + '\n'.join(allowed_levels)
+                )
+            if period_agg_method.lower() not in ['sum','max']:
+                raise ValueError(f"Fix period agg method in GSw_PRM_StressThreshold{stress_metric}")
+            if not (float(ppm) >= 0):
+                raise ValueError(
+                    f"ppm in GSw_PRM_StressThreshold{stress_metric} must be a positive number "
+                    f"but '{ppm}' was provided"
+                )
```
```suggestion
            if not (float(value) >= 0):
                raise ValueError(
                    f"value in GSw_PRM_StressThreshold{stress_metric} must be a positive number "
                    f"but '{value}' was provided"
                )
```

runreeds.py
```diff
@@ -319,33 +319,53 @@ def check_compatibility(sw):
...
+                raise ValueError(
+                    f"stress metric in GSw_PRM_StressThreshold{stress_metric} must be 'EUE', 'NEUE', or 'LOLE' "
+                    f"but '{stress_metric}' was provided"
+                )
+            if (sw['GSw_PRM_StressModel'].lower() != 'pras') and (stress_metric.upper() != 'EUE'):
+                err = (
+                    f"The combination of GSw_PRM_StressModel={sw['GSw_PRM_StressModel']} and "
+                    f"stress_metric={stress_metric} is not supported."
+                )
+                raise NotImplementedError(err)
```
patrickbrown4: This one's obsolete so can be removed (we used to have a different model, `osprey`, that could be used in place of PRAS, but it's been gone for a while now)
```suggestion
```

reeds/resource_adequacy/stress_periods.py
```diff
@@ -616,7 +655,13 @@ def update_prm(sw, t, iteration, failed, combined_periods_write):
...
 
     ## Fixed-increment update
-    if int(sw.GSw_PRM_UpdateMethod) == 1:
+    if int(sw.GSw_PRM_UpdateMethod) == 1 or 'EUE' not in stress_metrics:
+        if int(sw.GSw_PRM_UpdateMethod) > 1:
+            print(
+                f"Warning: GSw_PRM_UpdateMethod is set to {sw.GSw_PRM_UpdateMethod}, "
+                "but EUE is not included in GSw_PRM_StressThresholdMetrics, so defaulting to fixed increment. " 
+                "Add EUE to GSw_PRM_StressThresholdMetrics to use PRAS-informed update method."
+            )
```
patrickbrown4: Can you move this to `runreeds.check_compatibility()` and raise it as a `NotImplementedError` instead?

reeds/resource_adequacy/stress_periods.py
```diff
@@ -647,13 +692,22 @@ def update_prm(sw, t, iteration, failed, combined_periods_write):
 def main(sw, t, iteration=0, logging=True):
     """
     """
-    #%% Write consolidated NEUE so far
+    #%% Write consolidated stress metrics so far
     try:
+        ## TODO: Check if get_and_write_neue() is still needed
```
patrickbrown4, line 700: We don't use the national NEUE within the model so it can be removed (sorry for all the cruft). You would also remove the following blocks. It's kind of a lot though, and it's not directly related to this PR, so it might be best to save for a followup cleanup PR.

https://github.com/ReEDS-Model/ReEDS/blob/ce7c0d5a46bacc05bf4681f1220e7fb90931878a/postprocessing/bokehpivot/reeds2.py#L2764-L2775

https://github.com/ReEDS-Model/ReEDS/blob/ce7c0d5a46bacc05bf4681f1220e7fb90931878a/postprocessing/bokehpivot/reports/templates/reeds2/standard_report_expanded.py#L57

https://github.com/ReEDS-Model/ReEDS/blob/ce7c0d5a46bacc05bf4681f1220e7fb90931878a/postprocessing/compare_cases.py#L196

All the `'NEUE (ppm)'` keys in `reedsplots.plot_diff()` like https://github.com/ReEDS-Model/ReEDS/blob/ce7c0d5a46bacc05bf4681f1220e7fb90931878a/reeds/reedsplots.py#L117

https://github.com/ReEDS-Model/ReEDS/blob/ce7c0d5a46bacc05bf4681f1220e7fb90931878a/reeds/reedsplots.py#L486-L490

postprocessing/compare_cases.py
```diff
@@ -473,12 +473,12 @@ def plot_bars_abs_stacked(
 dictin_neue = {}
 dictin_neue_all = {}
 for case in tqdm(cases, desc='NEUE'):
-    infiles = sorted(glob(os.path.join(cases[case],'outputs','neue_*.csv')))
+    infiles = sorted(glob(os.path.join(cases[case],'outputs','NEUE_*.csv')))
```
patrickbrown4: Personally I would default to keeping most filenames lower case. Changing the case would also break backwards-compatibility (which we normally don't worry about, except for in `compare_cases.py` to avoid headaches running for older runs) on linux. Would it complicate things to revert?

(I agree the switch names are more readable with upper case NEUE/EUE/LOLH; here I'm just concerned about the filenames.)

reeds/resource_adequacy/stress_periods.py
```diff
@@ -18,31 +18,35 @@
...
-def get_pras_eue(case, t, iteration=0):
+def get_pras_stress_metric(case, t, iteration=0, stress_metric='EUE'):
     """
     """
     ### Get PRAS outputs
+    # NEUE load division takes place in the caller function
+    # get_stress_metric_periods() or get_annual_stress_metric() based on agg_period
+    if stress_metric.upper() == 'NEUE':
+        stress_metric = 'EUE'
+
```
patrickbrown4: Coercing arguments makes me nervous; my suggestion would just be to call it with the right argument. You could use a mapper like:
```python
use_metric = {'NEUE':'EUE'}
get_pras_stress_metric(stress_metric=use_metric.get(stress_metric, stress_metric))
```

reeds/resource_adequacy/stress_periods.py
```diff
@@ -70,37 +74,37 @@ def get_eue_periods(
...
         dfmetric_period = (
-            dfeue
-            .groupby([dfeue.index.year, dfeue.index.month, dfeue.index.day])
+            dfmetric
+            .groupby([dfmetric.index.year, dfmetric.index.month, dfmetric.index.day])
             .agg(period_agg_method)
             .rename_axis(['y','m','d'])
         )
-    elif stress_metric.upper() == 'NEUE':
+    else:
```
patrickbrown4: I think it'd be clearest to drop support for `stress_metric = 'NEUE'` and just keep the first block.

To clarify, we'd keep `GSw_PRM_StressThresholdNEUE`, and the default `GSw_PRM_StressThresholdMetrics` would still be `NEUE`; we'd just drop the third argument of the compound switches and use EUE for stress period selection for both the `GSw_PRM_StressThresholdEUE` and `GSw_PRM_StressThresholdNEUE` metrics

patrickbrown4 (commented):
This is cool, thanks! I haven't gone all the way through `stress_periods.py` yet, but adding some interim comments. My main thoughts are:
1. I would call it LOLH instead of LOLE if the units are event-hours
2. I think it would simplify the processing to drop the third argument of the compound `GSw_PRM_StressThreshold{LOLE|EUE|NEUE}` switches; instead of leaving it up to the user, NEUE- and EUE-based thresholds would always add stress periods with the highest EUE, and the LOLH-based threshold would always add stress periods with the highest LOLH. (When we add the depth-based metric, it would always add stress periods with the highest peak outage MW.)

postprocessing/compare_cases.py
abdelrahman-ayad: Ok, I'll revert the filename changes. 
Switches in uppercase -> writing filenames in lowercase.

reeds/resource_adequacy/stress_periods.py
abdelrahman-ayad: Fixed in the PR revision.

reeds/resource_adequacy/stress_periods.py
abdelrahman-ayad: Addressed in PR revision. dfmetric is now determined using EUE for the EUE and NEUE cases, so dfmetric_period is calculated accordingly

runreeds.py
abdelrahman-ayad: Addressed in PR revision.

runreeds.py
abdelrahman-ayad: Removed.

runreeds.py
abdelrahman-ayad: Fixed in PR revision.

cases.csv
```diff
@@ -262,14 +262,17 @@ GSw_PRM_NetImportLimit,Turn on (1) or off (0) the eq_firm_transfer_limit constra
...
+GSw_PRM_StressThresholdEUE,Annual EUE level [MWh] threshold above which to re-solve the latest model year with new stress periods; formulated as HierarchyLevel_EUE_StressMetric_PeriodAggMethod where HierarchyLevel is a column in hierarchy.csv; EUE is expected unserved energy; StressMetric EUE (only used in period selection); PeriodAggMethod is 'sum' or 'max' over the hours in each period (only used in period selection) (see README.md for detailed notes),N/A,transgrp_10000_sum,
+GSw_PRM_StressThresholdNEUE,Annual NEUE level [ppm] threshold above which to re-solve the latest model year with new stress periods; formulated as HierarchyLevel_NEUEppm_StressMetric_PeriodAggMethod where HierarchyLevel is a column in hierarchy.csv; NEUEppm is normalized expected unserved energy in parts per million; StressMetric is EUE or NEUE (only used in period selection); PeriodAggMethod is 'sum' or 'max' over the hours in each period (only used in period selection) (see README.md for detailed notes),N/A,transgrp_1_sum,
```
patrickbrown4: - [ ] Change default LOLH from 36 to 2.4
- [ ] Tweak description to match new 3-element structure
- [ ] If the EUE option is kept (see note above), change the input for `GSw_PRM_StressThresholdEUE` to per-year as well
  - [ ] including downstream in `stress_periods.py` if it is not already

```suggestion
GSw_PRM_StressThresholdLOLH,LOLH threshold [hours/year] above which to re-solve the latest model year with new stress periods; formulated as HierarchyLevel_LOLH_PeriodAggMethod where HierarchyLevel is a column in hierarchy.csv; LOLH is loss of load event-hours per year [event-h/year]; PeriodAggMethod is 'sum' or 'max' over the hours in each period (only used in period selection) (see README.md for detailed notes),N/A,transgrp_2.4_sum,
GSw_PRM_StressThresholdEUE,Annual EUE level [MWh] threshold above which to re-solve the latest model year with new stress periods; formulated as HierarchyLevel_EUE_PeriodAggMethod where HierarchyLevel is a column in hierarchy.csv; EUE is expected unserved energy per year [MWh/year]; PeriodAggMethod is 'sum' or 'max' over the hours in each period (only used in period selection) (see README.md for detailed notes),N/A,transgrp_10000_sum,
GSw_PRM_StressThresholdNEUE,Annual NEUE level [ppm] threshold above which to re-solve the latest model year with new stress periods; formulated as HierarchyLevel_NEUEppm_PeriodAggMethod where HierarchyLevel is a column in hierarchy.csv; NEUEppm is normalized expected unserved energy in parts per million [ppm]; PeriodAggMethod is 'sum' or 'max' over the hours in each period (only used in period selection) (see README.md for detailed notes),N/A,transgrp_1_sum,
```

reeds/resource_adequacy/stress_periods.py
```diff
@@ -200,43 +173,49 @@ def get_and_write_neue(sw, write=True):
         eue.to_csv(os.path.join(sw['casedir'],'outputs','eue.csv'))
     return neue
 
-
```
patrickbrown4, line 203: (nitpick) standard is 2 lines between function definitions

reeds/resource_adequacy/stress_periods.py
```diff
@@ -200,43 +173,49 @@ def get_and_write_neue(sw, write=True):
...
+        ### Get stress metric summed over year
+        _metric[hierarchy_level,'sum'] = dfmetric.rename(columns=rmap).groupby(axis=1, level=0).sum().sum()
+
+        ### Get max stress metric hour
+        _metric[hierarchy_level,'max'] = dfmetric.rename(columns=rmap).groupby(axis=1, level=0).sum().max()
+
+        if stress_metric.upper() == 'NEUE':
+            _metric[hierarchy_level,'sum'] = (
+            dfmetric.rename(columns=rmap).groupby(axis=1, level=0).sum().sum()
             / dfload.rename(columns=rmap).groupby(axis=1, level=0).sum().sum()
```
patrickbrown4:
```suggestion
                dfmetric.rename(columns=rmap).groupby(axis=1, level=0).sum().sum()
                / dfload.rename(columns=rmap).groupby(axis=1, level=0).sum().sum()
```

reeds/resource_adequacy/stress_periods.py
```diff
@@ -200,43 +173,49 @@ def get_and_write_neue(sw, write=True):
...
-        ) * 1e6
-        ### Get max NEUE hour
-        _neue[hierarchy_level,'max'] = (
-            dfeue.rename(columns=rmap).groupby(axis=1, level=0).sum()
+            ) * 1e6
+            ### Get max NEUE hour
+            _metric[hierarchy_level,'max'] = (
+            dfmetric.rename(columns=rmap).groupby(axis=1, level=0).sum()
             / dfload.rename(columns=rmap).groupby(axis=1, level=0).sum()
```
patrickbrown4:
```suggestion
                dfmetric.rename(columns=rmap).groupby(axis=1, level=0).sum()
                / dfload.rename(columns=rmap).groupby(axis=1, level=0).sum()
```

reeds/resource_adequacy/stress_periods.py
```diff
@@ -288,8 +263,7 @@ def get_shoulder_periods(sw, criterion, dfenergy_r, high_eue_periods):
         day_after = timeindex[(day_index + periodhours) % len(timeindex)]
 
         if (
-            ((cutofftype == 'eue') and (end_headspace_MWh / day_eue >= float(cutoff)))
-            or ((cutofftype[:3] == 'cap') and (end_headspace_frac  >= float(cutoff)))
```
patrickbrown4, line 292: The default `GSw_PRM_StressStorageCutoff` is still `EUE_0.1`, which looks like it means the default shoulder-period functionality would be removed with this change.

I think we should keep the original EUE formulation in `get_shoulder_periods()`. It should apply for both `GSw_PRM_StressThresholdEUE` and `GSw_PRM_StressThresholdNEUE` (both of which use EUE to select stress periods).

I'm not sure if it should apply to `GSw_PRM_StressThresholdLOLH`-identified periods. If it does, I think it's fine to still use EUE to identify those shoulder periods. If that complicates things, you can just drop the shoulder functionality for LOLH-identified periods (but keep it for the NEUE-/EUE-identified periods).

reeds/resource_adequacy/stress_periods.py
```diff
@@ -311,8 +284,91 @@ def get_shoulder_periods(sw, criterion, dfenergy_r, high_eue_periods):
 
     return shoulder_periods
 
+def _evaluate_stress_threshold_criterion(stress_criteria, criterion, sw, t, iteration, dfenergy_r, stressperiods_this_iteration, stress_metric):
```
patrickbrown4: (nitpick; wrap long lines)
(also wouldn't hurt to add a brief docstring saying what the function does)
```suggestion
def _evaluate_stress_threshold_criterion(
    stress_criteria,
    criterion,
    sw,
    t,
    iteration,
    dfenergy_r,
    stressperiods_this_iteration,
    stress_metric,
):
```

reeds/resource_adequacy/stress_periods.py
```diff
@@ -647,13 +663,22 @@ def update_prm(sw, t, iteration, failed, combined_periods_write):
...
-        neue.round(2).to_csv(
-            os.path.join(sw.casedir, 'outputs', f"neue_{t}i{iteration}.csv")
-        )
+        stress_metrics = sw.GSw_PRM_StressThresholdMetrics.split('/')
+        ## TODO: check if need to refactor or remove
+        # Include NEUE if not already specified, used for plots - only used for writing files
+        if 'NEUE' not in stress_metrics:
+            stress_metrics.append('NEUE')
+
+        for stress_metric in stress_metrics:
```
patrickbrown4: It'd be nice to always write all the metrics even if they aren't used in the run
```suggestion
        for stress_metric in ['NEUE', 'EUE', 'LOLH']:
```

reeds/resource_adequacy/stress_periods.py
```diff
@@ -647,13 +663,22 @@ def update_prm(sw, t, iteration, failed, combined_periods_write):
...
-        neue = get_annual_neue(sw.casedir, t, iteration=iteration)
-        neue.round(2).to_csv(
-            os.path.join(sw.casedir, 'outputs', f"neue_{t}i{iteration}.csv")
-        )
+        stress_metrics = sw.GSw_PRM_StressThresholdMetrics.split('/')
+        ## TODO: check if need to refactor or remove
+        # Include NEUE if not already specified, used for plots - only used for writing files
+        if 'NEUE' not in stress_metrics:
+            stress_metrics.append('NEUE')
+
```
patrickbrown4, line 703: Seems like this could be removed (as noted below, let's just always write the results for all metrics)

reeds/resource_adequacy/stress_periods.py
```diff
@@ -18,31 +18,31 @@
 
 
 #%%### Functions
-def get_pras_eue(case, t, iteration=0):
+def get_pras_stress_metric(case, t, iteration=0, stress_metric='EUE'):
```
patrickbrown4: Add `from typing import Literal` to the import block above, then:
```suggestion
def get_pras_stress_metric(case, t, iteration=0, stress_metric:Literal['EUE','LOLE']='EUE'):
```
because the only columns saved in the `PRAS_{t}i{iteration}.h5` file are EUE and LOLE (which in this case is the same as LOLH since it's at hourly resolution).

Then below, read EUE for the EUE/NEUE metrics and LOLE for the LOLH metric.

Image

reeds/resource_adequacy/stress_periods.py
```diff
@@ -200,43 +173,49 @@ def get_and_write_neue(sw, write=True):
...
 
-
-def get_annual_neue(case, t, iteration=0):
+def get_annual_stress_metric(case, t, stress_metric, iteration=0):
     """
     """
-    ### Get EUE from PRAS
-    dfeue = get_pras_eue(case=case, t=t, iteration=iteration)
+    ### Get values from PRAS
+    dfmetric = get_pras_stress_metric(case=case, t=t, iteration=iteration, stress_metric=stress_metric)
```
patrickbrown4: Add the `use_metric_for_pras = {'EUE':'EUE', 'NEUE':'EUE', 'LOLH':'LOLE'}` dictionary here as well and use it in the function call, since only EUE and LOLE work:

```suggestion
    use_metric_for_pras = {'EUE':'EUE', 'NEUE':'EUE', 'LOLH':'LOLE'}
    dfmetric = get_pras_stress_metric(
        case=case,
        t=t,
        iteration=iteration,
        stress_metric=use_metric_for_pras[stress_metric],
    )
```

(I tried running and the `neue_{t}i{iteration}.csv` files were coming out empty)

reeds/resource_adequacy/stress_periods.py
```diff
@@ -70,52 +70,25 @@ def get_eue_periods(
     ### Get the region aggregator
     rmap = reeds.io.get_rmap(case=case, hierarchy_level=hierarchy_level)
 
-    ### Get EUE from PRAS
-    dfeue = get_pras_eue(case=case, t=t, iteration=iteration)
+    ### Get stress metric from PRAS
+    # Use EUE metric for both EUE and NEUE calculations, since the load division to get NEUE is
+    # peformed afterwards based on agg_period. 
+    use_metric_for_pras = {'EUE':'EUE', 'NEUE':'EUE', 'LOLE':'LOLE'}
```
patrickbrown4:
```suggestion
    use_metric_for_pras = {'EUE':'EUE', 'NEUE':'EUE', 'LOLH':'LOLE'}
```

reeds/resource_adequacy/stress_periods.py
```diff
@@ -326,91 +382,50 @@ def get_eue_sorted_periods(sw, t, iteration):
...
-
-    eue_sorted_periods = pd.concat(_eue_sorted_periods, names=['criterion'])
+    stress_criteria = {'stress_sorted_periods': {}, 'failed': {}, 'high_stress_periods': {}, 'shoulder_periods': {}}
+
+    # Validation check: Display any GSw_PRM_StressThreshold{metric}
+    # that is not specified in GSw_PRM_StressThresholdMetrics
+    stress_metric_switches = sw.GSw_PRM_StressThresholdMetrics.split('/')
+    
+    # stress periods column names for writing outputs
+    stress_metrics_units = {'EUE':'MWh', 'NEUE':'ppm', 'LOLE':'days'}
```
patrickbrown4: I think LOLE would be "events/year", but since it's not supported yet, it should be "event-hours/year" for LOLH:
```suggestion
    stress_metrics_units = {'EUE':'MWh', 'NEUE':'ppm', 'LOLH':'event-h/year'}
```

reeds/resource_adequacy/stress_periods.py
```diff
@@ -326,91 +382,50 @@ def get_eue_sorted_periods(sw, t, iteration):
...
+    stress_sorted_periods = stress_criteria['stress_sorted_periods']
+    failed = stress_criteria['failed']
+    high_stress_periods = stress_criteria['high_stress_periods']
+    shoulder_periods = stress_criteria['shoulder_periods']
+
+    stress_sorted_periods = pd.concat(stress_sorted_periods, names=['criterion'])
 
     ### Get lists of stress periods: new (added this iteration) and all
+    ##TODO: What if same high stress period is added for multiple criteria?
+    # Currently the duplicates are dropped.
```
patrickbrown4, line 454: Yes, duplicate stress periods should be dropped

reeds/resource_adequacy/stress_periods.py
```diff
@@ -326,91 +382,50 @@ def get_eue_sorted_periods(sw, t, iteration):
...
+    stress_criteria = {'stress_sorted_periods': {}, 'failed': {}, 'high_stress_periods': {}, 'shoulder_periods': {}}
+
+    # Validation check: Display any GSw_PRM_StressThreshold{metric}
+    # that is not specified in GSw_PRM_StressThresholdMetrics
+    stress_metric_switches = sw.GSw_PRM_StressThresholdMetrics.split('/')
+    
+    # stress periods column names for writing outputs
+    stress_metrics_units = {'EUE':'MWh', 'NEUE':'ppm', 'LOLE':'days'}
+    stress_metrics_col_names = {m:f'{m}_{stress_metrics_units[m]}' for m in 
+                                stress_metric_switches}
```
patrickbrown4: For readability, move this definition lower, just before the one place it's used

reeds/resource_adequacy/stress_periods.py
```diff
@@ -326,91 +382,50 @@ def get_eue_sorted_periods(sw, t, iteration):
...
+            print(f"Evaluating GSw_PRM_StressThreshold {stress_metric.upper()} with criterion: {criterion}")
+            stress_criteria = _evaluate_stress_threshold_criterion(stress_criteria,
+                                                                   criterion,
+                                                                   sw,
+                                                                   t,
+                                                                   iteration,
+                                                                   dfenergy_r,
+                                                                   stressperiods_this_iteration,
+                                                                   stress_metric
+                                                                   )
```
patrickbrown4: (nitpick)
```suggestion
            stress_criteria = _evaluate_stress_threshold_criterion(
                stress_criteria,
                criterion,
                sw,
                t,
                iteration,
                dfenergy_r,
                stressperiods_this_iteration,
                stress_metric,
            )
```

reeds/resource_adequacy/stress_periods.py
```diff
@@ -311,8 +284,91 @@ def get_shoulder_periods(sw, criterion, dfenergy_r, high_eue_periods):
...
+                                                                                                        
+    ## NEUE Example: criterion = 'transgrp_1_sum' 
+    (hierarchy_level, stress_value, period_agg_method) = criterion.split('_')
+
+    # LOLH and LOLE values are defined on a yearly basis, 
+    # PRAS reports the total LOLH/LOLE over the entire evaluation periods
+    # Multiply by the number of resource adequacy years to compare
+
+    if stress_metric.upper() in ['LOLE', 'LOLH']:
+        stress_value = float(stress_value) * len(sw['resource_adequacy_years'])
```
patrickbrown4: I think it'd be clearest if we report the results in MWh/year for EUE and event-hours/year for LOLH. (NEUE is already normalized so doesn't need to be changed.) Otherwise users will have to remember to repeat this step in all of their downstream processing. (I'm hoping we can eventually be more flexible on the number & identify of RA years considered, and we don't want the RA metrics to be jumping around just because different runs use different numbers of years.) Could you instead:
- [ ] Move the normalization by number of weather years for EUE and LOLH up to `get_annual_stress_metric()` (so it is applied in the `eue_*.csv` and `lolh_*.csv` outputs that get written out), where we would divide the annual metrics by `len(sw['resource_adequacy_years'])`
- [ ] Remove the multiplication here (so we are always comparing EUE/y and event-h/y within the code)

cases.csv
```diff
@@ -262,14 +262,17 @@ GSw_PRM_NetImportLimit,Turn on (1) or off (0) the eq_firm_transfer_limit constra
...
+GSw_PRM_StressThresholdMetrics,"/-delimited list of metrics for identifying stress periods (supported options are: NEUE, EUE, LOLH)",N/A,NEUE,
+GSw_PRM_StressThresholdLOLH,LOLH threshold [hours/year] above which to re-solve the latest model year with new stress periods; formulated as HierarchyLevel_LOLH_StressMetric_PeriodAggMethod where HierarchyLevel is a column in hierarchy.csv; LOLH is number of loss of load event-hours; StressMetric is LOLH; PeriodAggMethod is 'sum' or 'max' over the hours in each period (only used in period selection) (see README.md for detailed notes),N/A,transgrp_36_sum,
+GSw_PRM_StressThresholdEUE,Annual EUE level [MWh] threshold above which to re-solve the latest model year with new stress periods; formulated as HierarchyLevel_EUE_StressMetric_PeriodAggMethod where HierarchyLevel is a column in hierarchy.csv; EUE is expected unserved energy; StressMetric EUE (only used in period selection); PeriodAggMethod is 'sum' or 'max' over the hours in each period (only used in period selection) (see README.md for detailed notes),N/A,transgrp_10000_sum,
```
patrickbrown4: Can you remind me, do any of the regions you reviewed formulate their RA target as EUE [MWh] instead of NEUE [ppm]? EUE MWh would be a bit of a weird threshold in the context of load growth, and a single number wouldn't really make sense to apply across different `transgrp` regions with different loads (as is specified in the default value).

So I guess my feeling is that it might be simpler just to drop `GSw_PRM_StressThresholdEUE` and stick with NEUE and LOLH as the two options. What do you think?

reeds/resource_adequacy/stress_periods.py
```diff
@@ -590,33 +604,35 @@ def update_prm(sw, t, iteration, failed, combined_periods_write):
...
     failed_regions = (
         pd.concat(_failed_regions)
-        .sort_values(by=['ppm'])
+        .sort_values(by=['stress_value'])
         .drop_duplicates(subset='r', keep='first')
     )
 
     ## Fixed-increment update
-    if int(sw.GSw_PRM_UpdateMethod) == 1:
+    if int(sw.GSw_PRM_UpdateMethod) == 1 or 'EUE' not in stress_metrics:
```
patrickbrown4: `update_prm()` currently works with NEUE as the only stress metric, so we should maintain that functionality.

Instead of adding `stress_metrics` as an input and adding this conditional, could you instead just pass it the entries of `failed` that arise from the NEUE criterion?

postprocessing/single_case_plots.py
```diff
@@ -701,7 +701,9 @@
         print(traceback.format_exc())
 
     try:
-        level, threshold, _, metric = sw['GSw_PRM_StressThreshold'].split('/')[0].split('_')
+        _first_metric = sw['GSw_PRM_StressThresholdMetrics'].split('/')[0].upper()
+        _parts = sw[f'GSw_PRM_StressThreshold{_first_metric}'].split('_')
+        level, threshold, _stress_metric, metric = _parts[0], _parts[1], _parts[2], _parts[3]
```
patrickbrown4: These are now 3-part switches, so the old 3rd element should be removed

```suggestion
        level, threshold, metric = _parts
```

reeds/reedsplots.py
```diff
@@ -114,7 +114,7 @@ def plot_diff(
         'Present Value of System Cost': 'Discounted Cost (Bil $)',
         'Runtime (hours)': 'processtime',
         'Runtime by year (hours)': 'processtime',
-        'NEUE (ppm)': 'neue',
+        'NEUE (ppm)': 'NEUE',
```
patrickbrown4: This function uses the bokeh outputs, which I don't think were changed
```suggestion
        'NEUE (ppm)': 'neue',
```

reeds/reedsplots.py
```diff
@@ -4147,7 +4147,9 @@ def plot_stressperiod_evolution(
     """Plot NEUE by year and stress period iteration"""
     ### Parse inputs
     sw = reeds.io.get_switches(case)
-    _level, _threshold, _, _metric = sw['GSw_PRM_StressThreshold'].split('/')[0].split('_')
+    _first_metric = sw['GSw_PRM_StressThresholdMetrics'].split('/')[0].upper()
+    _parts = sw[f'GSw_PRM_StressThreshold{_first_metric}'].split('_')
+    _level, _threshold, _stress_metric, _metric = _parts[0], _parts[1], _parts[2], _parts[3]
```
patrickbrown4:
```suggestion
    _level, _threshold, _metric = _parts
```

patrickbrown4 (commented):
Thanks, happy to discuss any of these if helpful (particularly the one about dropping `GSw_PRM_StressThresholdEUE`). Could you re-request a review when I should take another look?

reeds/resource_adequacy/stress_periods.py
abdelrahman-ayad: Addressed in PR review.

postprocessing/single_case_plots.py
abdelrahman-ayad: Switches updated in the PR revision.

reeds/reedsplots.py
abdelrahman-ayad: Switches updated in the PR revision.

reeds/resource_adequacy/stress_periods.py
abdelrahman-ayad: Addressed.

cases.csv
abdelrahman-ayad: Updated switches.

cases.csv
abdelrahman-ayad: Some regions (e.g., SPP had EUE seasonal targets and The Tri-State Agency had <0.5 GWh, and ERCOT (magnitude) of 19 GW). Nevertheless, I agree that NEUE makes it easier to implement across multiple regions with different load profiles etc. I'll remove the EUE for now and we can always return it in a future update.

patrickbrown4:
Included in #124

Issue #23, opened 2026-04-16
State-level capacity mandates are applied in full when running with only a portion of a state


wesleyjcole:
### Description of Issue
In ReEDS runs where part of a state is included and that state happens to have a capacity mandate (e.g., for storage or wind), the entire mandated capacity is required to be built in just the part of the state that's modeled. For example, in cases with PJM counties as an island, where very little of Michigan is included, the entire 2.5 GW storage mandate is applied to the handful of Michigan counties.

<img width="1056" height="782" alt="Image" src="https://github.com/user-attachments/assets/78f0c299-9ff6-4a3e-8005-3e6a2d32784c" />

### Potential Solutions
- Zero out the capacity mandate if a state isn't fully modeled.
- Zero out the capacity mandate if a state is modeled to some extent less than a threshold value (e.g., 75%).
- Multiply the capacity mandate by the load share of the part of the state that's modeled (more work, unknown benefit).

wesleyjcole:
This chart shows the load share vs. battery capacity share in a national run for states that have a storage mandate and more than one model regions (uses the head of main from March 12, 2026).  The load and storage is approximately correlate.  Michigan and New York are the biggest outliers. MI puts most of the capacity in p104, which I expect is a ReEDS artifact associated with the transgrp regions. New York puts more capacity upstate even though a large amount of load is in Long Island. I expect that is related to the high capital cost multipiers in Long Island.

<img width="1687" height="1332" alt="Image" src="https://github.com/user-attachments/assets/cc8d39ea-2a2e-4a77-a219-c29eccc88f24" />

PR #102, opened 2026-05-20
Disaggregate and re-aggregate inputs with regional scope of legacy (134) zones

kodiobika:
## Summary
The two major goals of this PR are to:
- Allow the input files that we only have legacy zonal data for (i.e., data corresponding to the current default 134 zones) to be incorporated in ReEDS runs using new/custom zones by first disaggregating those inputs to the county level and then re-aggregating them to the zone level (using the zones for any given run).
- Mostly deprecate `reeds/input_processing/aggregate_regions.py`

## Technical details 
- Adds functions to `reeds/spatial.py` for disaggregating from the legacy 134 zones to counties and aggregating from counties to the zones corresponding to a given run.
- For all files except the transmission files (which will be addressed in a follow-up PR), aggregation and disaggregation now take place in `reeds/input_processing/copy_files.py`.
- Disaggregation and aggregation both take place in the same run now (rather than disaggregation only happening in sub-BA runs and aggregation only happening in super-BA runs)

### Additional changes 
- To clean up `reeds/input_processing/runfiles.csv` ~~and make it clearer which files are actually being aggregated and disaggregated, I set the `aggfunc` and `disaggfunc` for all files that are not read from the repo (i.e., files that are created after `reeds/input_processing/aggregate_regions.py`) to `ignore`.~~ , I deleted all rows that don't have a specified `filepath`.
- In `reeds/input_processing/runfiles.csv`, I updated the `disaggfunc` for `unapp_water_sea_distr.csv` ~~and `water_req_psh_10h_1_51.csv`~~ from `geosize` to `uniform`, the `aggfunc` for `water_req_psh_10h_1_51.csv` from `mean` to `sum`, and the 'aggfunc' for the prescribed wind files from `sum` to `ignore`.
- ~~As part of testing this change I realized that there are duplicate rows in `inputs/hydro/SeaCapAdj_hy.csv` so I deleted those.~~ (@jvcarag will address this in a follow-up PR)

### Issues resolved 
Part of #16 

- Essentially zero change for the `Pacific` case: 
[results-main_Pacific,test_Pacific.pptx](https://github.com/user-attachments/files/28074104/results-main_Pacific.test_Pacific.pptx)

- There are small changes for the `NYVT_mixed` case: 
[results-0526_Main_NYVT_mixed,0527_AggDisagg_NYVT_mixed.pptx](https://github.com/user-attachments/files/28370401/results-0526_Main_NYVT_mixed.0527_AggDisagg_NYVT_mixed.pptx)
  - Based on the `inputs.gdx` files (see the `gdxdiffs.ipynb` notebook below), the differences stem from the `cap_hyd_szn_adj`~~, `water_req_psh`,~~ and `watsa_temp` parameters. `cap_hyd_szn_adj` is different because doing the disaggregation and aggregation ends up removing the duplicate rows from `inputs/hydro/SeaCapAdj_hy.csv` (see "Additional changes"), so for the groups with duplicates, the values in the parameter go from 2 to 1. ~~`water_req_psh` and `watsa_temp` are different for only the county-level zones of this case because the `disaggfunc` for their corresponding files was changed from `geosize` to `uniform`.~~ `watsa_temp` is different for only the county-level zones of this case because the `disaggfunc` for the corresponding file was changed from `geosize` to `uniform`.

- There are small changes for the `USA_defaults` case: 
[results-0526_Main_USA_defaults,0527_AggDisagg_USA_defaults.pptx](https://github.com/user-attachments/files/28370400/results-0526_Main_USA_defaults.0527_AggDisagg_USA_defaults.pptx)
  - Based on the `inputs.gdx` files (see the `gdxdiffs.ipynb` notebook below), the differences almost all stem from the aggregated zones (`z28` and `z122`). Because we now disaggregate to counties before re-aggregating to these zones, inputs where the `aggfunc` is `mean` now represent a weighted average of legacy zonal values (where weights correspond to the number of counties in each legacy zone) rather than a simple average of legacy zonal values, which results in different values for these aggregated zones. The only difference not related to the aggregated zones is in the `hyd_add_upg_cap` parameter, where `p108` and `p69` no longer have values. This is because we now disaggregate the corresponding file to counties first according to the `hydroexist` disaggfunc, and these legacy zones have no existing hydro.

- There are small changes for the `USA_decarb` case: 
[results-0519_Main_USA_decarb,0519_AggDisagg_USA_decarb.pptx](https://github.com/user-attachments/files/28074585/results-0519_Main_USA_decarb.0519_AggDisagg_USA_decarb.pptx)

- This is a notebook looking at the differences in `inputs.gdx` for the `USA_defaults` and `NYVT_mixed` cases: 
[gdxdiffs.ipynb](https://github.com/user-attachments/files/28351118/gdxdiffs.ipynb). Aside from the differences explained above, there are only rounding-error differences.

### Details to double-check

No

- [x] Water cooling: @jcarag or @stuartcohen8

- [x] Hydropower or PSH: @stuartcohen8


stuartcohen8:
`unapp_water_sea_distr.csv` is ok to be uniform, but `water_req_psh_10h_1_51.csv` should be reverted to geosize. The latter are water volumes, so they should be split up when disaggregating.  We think there is a separate bug with `SeaCapAdj_hy.csv` because older files do not have duplicates and have values for the hydD technology. Somewhere in input processing for PR1692 in the internal repo, this occurred. @jvcarag will make an issue to correct this file so that it has 1 values for hydD and no duplicates. In practice, it will not affect most if not all solutions because hydD capacity is high cost and therefore rarely economic. But if there are prescribed hydD builds, they might have there capacity zeroed out.

kodiobika:
> `unapp_water_sea_distr.csv` is ok to be uniform, but `water_req_psh_10h_1_51.csv` should be reverted to geosize. The latter are water volumes, so they should be split up when disaggregating. We think there is a separate bug with `SeaCapAdj_hy.csv` because older files do not have duplicates and have values for the hydD technology. Somewhere in input processing for PR1692 in the internal repo, this occurred. @jvcarag will make an issue to correct this file so that it has 1 values for hydD and no duplicates. In practice, it will not affect most if not all solutions because hydD capacity is high cost and therefore rarely economic. But if there are prescribed hydD builds, they might have there capacity zeroed out.

Thanks @stuartcohen8. Just checking again on `water_req_psh_10h_1_51.csv`, I see in `b_inputs.gms` that the units for `water_req_psh` are `Mgal/MW/yr` (as opposed to `Mgal`). Is that not correct? And if it's actually `Mgal`, should the `aggfunc` for that file be updated from `mean` to `sum`? Everything else sounds good to me.

stuartcohen8:
@kodiobika yes you're correct that aggfunc should be sum for that parameter. The water volumes are normalized by capacity because when they are used in water availability constraints, they are multiplied by the amount of PSH capacity that gets built to get a total water volume. The 'per year' convention is because water availability is generally characterized as an annual quantity of water available. So the units are right, and just to confirm you should sum to aggregate and use geosize to disaggregate. In reality it's more complicated based on basin dynamics, but that's beyond the scope of what we've done here so far.

kodiobika:
> @kodiobika yes you're correct that aggfunc should be sum for that parameter. The water volumes are normalized by capacity because when they are used in water availability constraints, they are multiplied by the amount of PSH capacity that gets built to get a total water volume. The 'per year' convention is because water availability is generally characterized as an annual quantity of water available. So the units are right, and just to confirm you should sum to aggregate and use geosize to disaggregate. In reality it's more complicated based on basin dynamics, but that's beyond the scope of what we've done here so far.

Got it, thanks again!

reeds/input_processing/aggregate_regions.py
```diff
@@ -1165,7 +1165,14 @@ def agg_disagg(filepath, r2aggreg_glob, r_ba_glob, runfiles_row):
...
 # loop over inputfiles from runfiles and call aggregation/disaggregation function
-for filepath in inputfiles:
+transmission_files = [
+    'transmission_capacity_future.csv',
+    'transmission_capacity_future_baseline.csv',
+    'transmission_cost_ac.csv',
+    'transmission_cost_dc.csv',
+    'transmission_distance.csv',
+]
+for filepath in transmission_files:
```
patrickbrown4, line 1143: It seems like we should be able to drop every remaining row in runfiles.csv with an empty filepath. Is that right? If so, should we go ahead and do it for simplicity and clarity? Or would you rather save it for a following cleanup PR?
It would probably also require removing the missing-files block: https://github.com/ReEDS-Model/ReEDS/blob/e18f59aaee3c722527ccc43b3ab55017e4bba241/reeds/input_processing/aggregate_regions.py#L1082

reeds/input_processing/runfiles.csv
patrickbrown4, line 1: It seems to me like these changed rows should just be able to be dropped. But it'd be good to run the `MonteCarlo` case from `cases_test.csv` just to make sure, since it also uses `runfiles.csv`.

reeds/spatial.py
```diff
@@ -215,3 +215,356 @@ def get_map(resolution='county', source='tiger', crs='ESRI:102008'):
...
+
+def apply_uniform_disaggregation(
+    df: pd.DataFrame,
+    region_col: str
+):
+    """
+    Disaggregate a dataframe whose regional scope is the 134 legacy
+    zones (as specified in the dataframe's 'region_col' column) to
+    the county level by assigning zonal values to their corresponding
+    counties uniformly.
```
patrickbrown4: Maybe a bit more specific (since "uniformly" could also be interpreted to mean 10 MW in region p1 with 10 counties would turn into 1 MW per county instead of 10 MW per county)
```suggestion
    the county level by copying the zonal values to their constituent counties.
```

reeds/spatial.py
```diff
@@ -215,3 +215,356 @@ def get_map(resolution='county', source='tiger', crs='ESRI:102008'):
...
+
+    return df
+
+
+def apply_variable_disaggregation(
+    df: pd.DataFrame,
+    region_col: str,
+    fix_cols: list[str],
+    inputs_case: str,
+    disagg_variable: str
```
patrickbrown4: Maybe `Literal['population', 'hydroexist']` (+ whatever the other supported options are) (enabled by `from typing import Literal`)?

Or should that go here instead? https://github.com/ReEDS-Model/ReEDS/blob/e18f59aaee3c722527ccc43b3ab55017e4bba241/reeds/io.py#L414

reeds/spatial.py
```diff
@@ -215,3 +215,356 @@ def get_map(resolution='county', source='tiger', crs='ESRI:102008'):
...
+    """
+    # Save the dataframe's original columns
+    # (used later to put the output in the correct format)
+    original_columns = df.columns
+
+    # Get legacy zone-to-county allocation factors for disagg_variable
+    disagg_data = reeds.io.get_disagg_data(
+        os.path.dirname(inputs_case),
+        disagg_variable
+    )
```
patrickbrown4, line 268: In the next PR (or whatever one adds support for `GSw_ZoneSet = z90`), we should double check what happens if `GSw_ZoneSet = z90` and `GSw_Region = r/NJ.NY_NYC`. NY_NYC is a subset of p127, and it's not immediately clear from `copy_files.write_disagg_data_files()` what would happen if your'e disaggregating/reaggregating with a subset of a z134 zone. I think it depends on whether you're using the full county2zone or the run-specific county2zone that only includes the zones in that run. I'd just want to make sure that the entire capacity (or whatever) of p127 doesn't end up in NY_NYC, since the other parts of p127 (NY_E and NY_W) aren't in the run. (Same idea as #23, but here for sub-z134 runs.)

I don't think we support sub-z134 runs now, so we can wait to test it once that capability works; just flagging since it came to mind.

patrickbrown4 (approved):
This is great! I really like the new structure. I'll defer to Stuart and Vincent on the hydro/water parts, but the rest looks good to me.

I think your approach to `aggregate_regions.py` makes sense; rather than picking through and removing the outdated code blocks now, we can move ahead with this PR and #95 in parallel, and then once they're both in, just remove the whole script.

reeds/input_processing/aggregate_regions.py
kodiobika: Good point, I'll go ahead and do it in this PR

reeds/spatial.py
kodiobika: Good point, I ended up using it in both locations

kodiobika:
Confirmed with @jvcarag over Teams that water/hydro changes look good now, so going to go ahead and push. Thanks again for the reviews!

PR #103, opened 2026-05-20
Update conda environment

patrickbrown4:
## Summary

This PR updates the conda environment to use new features in `gdxpds` and `geopandas`. The main changes are:
- python: 3.11 -> 3.14
- pandas: 2.0 -> 3.0
- gdxpds: 1.4 -> 4.0
  - uses gams.transfer
- geopandas: 0.14 -> 1.1
  - Enables the use of `.simplify_coverage()`, which speeds up all the geospatial analysis and plotting (which has been slow since #95, especially in `compare_cases.py`)

Thanks a lot to @mmowers for taking care of the bokeh update!

### Implementation notes 

Almost all of the code changes are related to pandas 3. The main things to look out for in new code are:
- `.groupby()` no longer accepts an `axis` argument, so all the uses of `df.groupby(axis=1).something()` are now `df.T.groupby().something().T`
- `.fillna(method='ffill')` (and analogous) is now just `.ffill()`
- Type checking in series and dataframes is generally more strict

### Additional changes

While working on debugging `calc_financial_inputs.py`, I simplified the `inputs/financials/financials_tech_*.csv` files by removing the country column and non-USA entries. They weren't complete anyway, and Canada/Mexico haven't worked since I've been here, so since it simplifies the process for adding new techs, I've kept the change even though it might not actually be required.

Changes for USA_default are nonzero but quite small: [results-v20260708_mainK0_USA_defaults,v20260708_envK0_USA_defaults.pptx](https://github.com/user-attachments/files/29856417/results-v20260708_mainK0_USA_defaults.v20260708_envK0_USA_defaults.pptx)

### 20260722
Same story: [results-v20260722_mainK0_USA_defaults,v20260722_envK0_USA_defaults.pptx](https://github.com/user-attachments/files/30331080/results-v20260722_mainK0_USA_defaults.v20260722_envK0_USA_defaults.pptx)

I'll go through the rest of these before merging:
- [x] bokehpivot
- [x] helper scripts
- [x] postprocessing scripts
  - [x] compare_cases.py
  - [x] input_plots.py
  - [x] run_reeds2pras.py
  - [x] single_case_plots.py
  - (I didn't test the ones that are less commonly used)
- [ ] hourlize
- [x] ~~reeds_to_rev.py~~ (currently broken; will be addressed by #6)
- [x] preprocessing scripts
  - [x] casemaker.py
  - [x] get_case_periods.py
- [x] cases_test.csv
  - [x] Mid_Case
  - [x] USA_decarb
  - [x] USA_defaults
  - [x] Interday_storage
  - [x] MonteCarlo_LHS_MC0001
  - [x] MonteCarlo_LHS_MC0002
  - [x] MonteCarlo_Random_MC0001
  - [x] MonteCarlo_Random_MC0002
  - [x] OR_water
  - [x] Pacific
  - [x] Pacific_2020
  - [x] Pacific_CC
  - [x] Pacific_DR
  - [x] Pacific_LoadSite95
  - [x] Pacific_MGA
  - [x] Pacific_rep15
  - [x] Pacific_weks
  - [x] R2P
  - [x] USA_faster
  - [x] NYVT_mixed
  - [x] PJM_county_CC
  - [x] Pacific_full_year
  - [x] USA_fast
  - [x] Everything
  - [x] MARICTNYNJPAOH_Offshore
  - [x] Simple
  - [ ] WECC_county: Fails on both
  - [x] WY_county
- [x] github runner

### Details to double-check

No


patrickbrown4:
Still working on a few things, so closing for now.

inputs/financials/financials_tech_ATB2023.csv
patrickbrown4, line 1: In the end these changes probably weren't necessary, but they were helpful for debugging (and make future updates very slightly easier), so I left them - let me know if you think they should be reverted

cases_test.csv
```diff
@@ -3,7 +3,7 @@ ignore,1,0,,,,,,,,,,,,,,,,,,,,,,,,,,,,
 GSw_Region,cendiv/Pacific,,country/USA,country/USA,country/USA,,st/ID.WY.NE.IA.IL,st/MA,,,,,,,st/WY,interconnect/western,transreg/PJM,st/NY.VT,st/OR,st/NE.NY.PA,st/NE.NY.PA,st/ID.WY.NE.IA.IL,st/KS,country/USA,country/USA,,,,st/MA.RI.CT.NY.NJ.PA.OH,
 endyear,2032,,2050,2050,2050,2029,2060,2026,,,,,,,,,,,2035,2030,2030,2060,2035,2050,2050,,,,,
 yearset,,,,,,,2010..2060..10,,,,,,,,,,,,,2010..2050..5,2010..2050..5,2010..2060..10,,,2010_2025_2050,,,,,
-GSw_ZoneSet,,,,,,,z54,z3109,,,,,,,z3109,z3109,z3109,PJMcounty,,,,z54,,z54,z48,,,,,
+GSw_ZoneSet,,,,,,,z54,z3109,,,,,,,z3109,z3109,z3109,PJMcounty,,,,z54,z134,z54,z48,,,,,
```
patrickbrown4: Most minimal way to keep `Simple` working and maintain the original intent (since ReEDS2PRAS doesn't work for single-zone systems)

environment.yml
```diff
@@ -1,50 +1,46 @@
-name: reeds2
+name: reeds
```
pesap, line 1: The environment is now named `reeds`, and `runreeds.py` checks for `reeds`, but the README and a few helper scripts still reference `reeds2`. We should update those references so new users don't create or activate the wrong environment.

environment.yml
```diff
@@ -1,50 +1,46 @@
...
 channels:
-  - defaults
   - conda-forge
```
pesap, line 3: Consider adding `nodefaults` so conda doesn't silently mix packages from user-configured default channels with conda-forge. That should make this env more reproducible and reduce ABI/linking surprises.

```suggestion
  - conda-forge
  - nodefaults
```

environment.yml
```diff
@@ -1,50 +1,46 @@
...
 dependencies:
-  - python=3.11
-  - bokeh=3.2
-  - click=8.0
-  - git-lfs=2.13
+  - python=3.14
```
pesap, line 6: Since this env is on Python 3.14, I'd explicitly pin the normal CPython ABI. conda-forge also has free-threaded `cp314t` builds, but `gamsapi==53.5.0` appears to publish `cp314` wheels, not `cp314t` wheels. This avoids the solver picking an ABI that later makes the pip step fail.

```suggestion
  - python=3.14
  - python_abi=3.14=*_cp314
```

environment.yml
```diff
@@ -1,50 +1,46 @@
...
-  - pytest=7.4
-  - requests=2.31
-  - scikit-learn=1.2
-  - scipy=1.11
-  - tqdm=4.65
+  - h5py=3.16
+  - matplotlib=3.11
+  - numpy=2.5
+  - openpyxl=3.1
+  - pandas=3.0
```
pesap, line 18: `pyyaml` is imported directly in several places, e.g. `reeds/inputs.py`, `reeds/input_processing/check_inputs.py`, and `preprocessing/casemaker.py`. It may arrive transitively today, but we should list it explicitly as a direct dependency.

Suggested addition near the core dependencies:

```yaml
  - pyyaml=6.0
```

pesap (commented):
Posted environment dependency review comments.

environment.yml
```diff
@@ -1,50 +1,46 @@
...
+  - folium=0.20 # for interactive maps
+  - ipykernel=7.3 # for interactive python in VS Code
+  - ipywidgets=8.1 # for jupyter notebooks
+  - mapclassify=2.10 # more mapping tools
   - mscorefonts=0.0 # extra fonts for plotting (only relevant on linux/HPC)
-  - networkx=3.1 # for uncommonly-used network analysis postprocessing
-  - notebook=6.5 # for jupyter notebooks
+  - myst-parser=5.1 # for building documentation
+  - notebook=7.6 # for jupyter notebooks
+  - python-pptx=1.0 # for postprocessing/compare_cases.py
```
pesap, line 38: `postprocessing/uncertainty_plots.py` imports `seaborn` directly, but I don't see it listed in this env. If that script is intended to work from the standard ReEDS environment, we should include it explicitly.

Suggested addition in the optional/postprocessing block:

```yaml
  - seaborn=0.13 # for postprocessing/uncertainty_plots.py
```

environment.yml
```diff
@@ -1,50 +1,46 @@
...
+  - sphinx<9 # for building documentation
+  - sphinx_rtd_theme=3.1 # for documentation
+  - sphinxcontrib-bibtex=2.7 # for documentation
```
pesap, line 42: `docs/source/conf.py` enables the `sphinx_design` extension, but this environment does not install `sphinx-design`. The docs workflow installs it separately, so local docs builds from this conda env may fail unless we add it here too.

Suggested addition:

```yaml
  - sphinx-design=0.7 # for building documentation; provides sphinx_design
```

environment.yml
```diff
@@ -1,50 +1,46 @@
...
-  - mapclassify=2.5 # more mapping tools
+  - fiona=1.10 # for interactive maps
+  - folium=0.20 # for interactive maps
+  - ipykernel=7.3 # for interactive python in VS Code
+  - ipywidgets=8.1 # for jupyter notebooks
+  - mapclassify=2.10 # more mapping tools
   - mscorefonts=0.0 # extra fonts for plotting (only relevant on linux/HPC)
-  - networkx=3.1 # for uncommonly-used network analysis postprocessing
-  - notebook=6.5 # for jupyter notebooks
+  - myst-parser=5.1 # for building documentation
```
pesap, line 36: The docs dependency pins here differ pretty significantly from `.github/workflows/build-docs.yaml` (`myst-parser=5.1` / `sphinx<9` here vs. `myst-parser==2.0.0` / `sphinx==7.2.6` in CI). Can we align these so local docs builds and CI use the same dependency set? Otherwise docs may pass locally but fail in CI, or vice versa.

environment.yml
patrickbrown4: There are a few modules that only show up in rarely used postprocessing scripts or options (e.g. `scikit-learn-extra`, `networkx`, and `seaborn`), so to reduce the number of required dependencies, we leave them out by default. The other ones have notes but I forgot to add one here; I've added it to `uncertainty_plots.py`.

environment.yml
patrickbrown4: Do you have a preference on which set of versions to use, these or the ones in `build-docs.yaml`? For these updates, my process is:

- Remove all the version locks from `environment.yml` and build the environment
- Get the resulting major.minor versions with `conda list` and add them back in to `environment.yml`

I'm not attached to a particular version of sphinx or myst, so happy to do whatever you or @kennedy-mindermann recommend.

pesap: I think my preference is that the environment.yml is the source of truth. But not a blocker IMO as long as we sync them.

reeds/results.py
```diff
@@ -462,7 +462,7 @@ def calc_systemcost(
 
         df.loc[:,op_type_ls] = (
             df.groupby('r')[op_type_ls]
-            .fillna(method='ffill', limit=sys_eval_years-1)
+            .bfill(limit=sys_eval_years-1)
```
kodiobika: We're changing forward-fill to back-fill here; is that intentional?

reeds/input_processing/copy_files.py
```diff
@@ -394,12 +394,7 @@ def read_banned_tech_file(full_path, filepath, inputs_case, r_county):
                 .apply(list)
                 .apply(sorted)
             )
-            r_all_counties_map = (
-                r_county.groupby('r')
-                ['county']
-                .apply(list)
-                .apply(sorted)
-            )
+            r_all_counties_map = r_county.county.tolist()
```
kodiobika: It looks like this change breaks the tech ban logic since `r_all_counties_map` would now just be a list of all counties (previously it was a `pd.Series` mapping zones to sorted lists of their counties)

reeds/input_processing/h5_to_gdx.py
```diff
@@ -49,7 +49,7 @@ def read_inputs(case:str|Path) -> tuple:
             except KeyError:
                 df = pd.DataFrame(columns=columns)
             for col in df:
-                if df[col].dtype == 'O':
+                if pd.api.types.is_string_dtype(df[col].dtype):
```
kodiobika, line 52: Just checking, do we need the `pd.api.types.is_object_dtype` check here too?

reeds/input_processing/hourly_load.py
```diff
@@ -222,15 +222,15 @@ def calibrate_hourly_state_load_to_historical_annuals(
...
+    calibration_diffs = -(
         state_load_hourly_calibration_year
         .groupby(
             state_load_hourly_calibration_year.index
             .get_level_values('datetime')
             .year
         )
         .transform('sum')
-    )
+    ).sub(calibration_year_historical_load)
```
kodiobika, line 233: Just curious, why reverse the operations here and in the `calibration_factors` equation?

reeds/io.py
```diff
@@ -982,9 +1007,8 @@ def read_h5_groups(filepath, parse_timestamps=False):
             if (
                 parse_timestamps
                 and ('datetime' in dfout.index.names)
```
kodiobika, line 1009: Should we add the `pd.DatetimeIndex` check here too?
```suggestion
                and ('datetime' in dfout.index.names)
                and not isinstance(dfout.index, pd.DatetimeIndex)
```

kodiobika (commented):
Thanks for doing this! Just some relatively minor comments

reeds/results.py
patrickbrown4: Nope that's a typo, thanks for the catch!

reeds/input_processing/hourly_load.py
patrickbrown4: Yeah this solution is not pretty but they seem to have changed the rules for `series.div(dataframe)` and `series.sub(dataframe)`, while `dataframe.div(series)` and `dataframe.sub(series)` still work. Here's the error I get with the original version of the code in the new environment; taking the suggestion in the final line of putting the dataframe first avoids the error.

```text
Traceback (most recent call last):
  File "/Users/pbrown/github/ReEDS/runs/v20260721_envM0_github_MA_county_CC/reeds/input_processing/hourly_load.py", line 782, in <module>
    main(reeds_path=reeds_path, inputs_case=inputs_case)
    ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pbrown/github/ReEDS/runs/v20260721_envM0_github_MA_county_CC/reeds/input_processing/hourly_load.py", line 636, in main
    calibrate_hourly_state_load_to_historical_annuals(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        state_load_hourly,
        ^^^^^^^^^^^^^^^^^^
        historical_state_load_annual
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
  File "/Users/pbrown/github/ReEDS/runs/v20260721_envM0_github_MA_county_CC/reeds/input_processing/hourly_load.py", line 199, in calibrate_hourly_state_load_to_historical_annuals
    calibration_factors = model_year_historical_load.div(
        state_load_hourly_model_year
    ...<5 lines>...
        .transform('sum')
    )
  File "/Users/pbrown/opt/anaconda3/envs/reeds/lib/python3.14/site-packages/pandas/core/series.py", line 7333, in truediv
    return self._flex_method(
           ~~~~~~~~~~~~~~~~~^
        other, operator.truediv, level=level, fill_value=fill_value, axis=axis
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
  File "/Users/pbrown/opt/anaconda3/envs/reeds/lib/python3.14/site-packages/pandas/core/series.py", line 6873, in _flex_method
    raise TypeError(
    ...<2 lines>...
    )
TypeError: Series.truediv does not support a DataFrame `other`. Use df.truediv(ser) instead.
```

kodiobika: Got it, thanks!

reeds/input_processing/copy_files.py
patrickbrown4: Thanks for the catch, I didn't read the original approach closely enough. The original gives `ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`, so it looks like checking for equality between lists has changed; I've refactored it to avoid the list comparison.

reeds/input_processing/h5_to_gdx.py
patrickbrown4: `pd.api.types.is_string_dtype('O')` gives True, so the original functionality is maintained. And I think strings are the only type of non-numeric object that could end up in `inputs.h5` (and the next line assumes it's a string), so I think we're ok; if we end up trying to write a weird dtype to .gdx I think it would balk.

reeds/io.py
patrickbrown4: I think that since we now check the type within `parse_h5_timestamps` and bypass the type conversion if it's already a datetime, it's ok to avoid the check here.

We also do the conversion from UTC-# to Etc/GMT+# in `parse_h5_timestamps`, so it seems safest to pass everything through that function to make sure they all come out with the same timezone format.

reeds/input_processing/copy_files.py
kodiobika: Nice, thanks! LGTM

kodiobika (approved):

reeds/reedsplots.py
```diff
@@ -6578,7 +6578,10 @@ def plot_eue_events(
     for iteration in iterations:
         ## Get events
         fpath = Path(case, 'outputs', f'eue_events_{year}i{iteration}.csv')
-        events = pd.read_csv(fpath, index_col=['level','region','number']).loc[level]
+        events = pd.read_csv(fpath, index_col=['level','region','number'])
+        if events.empty:
+            continue
```
pesap, line 6583: This has bite me in the past. Shall we print/log a message?

patrickbrown4: Good point, though since it's just for a plot and none of the other plots in this script give verbose logs, personally I find it more useful in this instance to keep things streamlined and only print more details if the plot failed, making it easy to scan which plots finished (it's the last thing left on the screen at the end of a run).

We don't yet have multiple logging levels but that would be nice to add in a future PR.

```
single_case_plots.py | 2026-07-23 18:08:13 | INFO | map_gencap_transcap-2032-sideplots.png
single_case_plots.py | 2026-07-23 18:08:13 | INFO | plot_interreg_transfer_ratio-transreg.png
single_case_plots.py | 2026-07-23 18:08:14 | INFO | bars_retirements_additions.png
single_case_plots.py | 2026-07-23 18:08:14 | INFO | map_h2_capacity-2032.png
single_case_plots.py | 2026-07-23 18:08:15 | INFO | plot_h2_timeseries-2032.png
single_case_plots.py | 2026-07-23 18:08:15 | INFO | plot_ra_metrics-country,interconnect,transreg,transgrp.png
single_case_plots.py | 2026-07-23 18:08:15 | INFO | plot_eue_events-mean_timesteps-transgrp-2032.png
single_case_plots.py | 2026-07-23 18:08:16 | INFO | plot_eue_events-max_timesteps-transgrp-2032.png
single_case_plots.py | 2026-07-23 18:08:18 | INFO | plot_PRAS_EUE-transgrp-2032i0.png
single_case_plots.py | 2026-07-23 18:08:19 | INFO | map_PRAS_neue-2026i0.png
single_case_plots.py | 2026-07-23 18:08:19 | INFO | map_PRAS_neue-2029i0.png
single_case_plots.py | 2026-07-23 18:08:20 | INFO | map_PRAS_neue-2032i0.png
single_case_plots.py | 2026-07-23 18:08:21 | INFO | plot_capacity_offline-transgrp-2032.png
```

PR #105, opened 2026-05-22
Add faster method for assigning actual to representative periods

### LINKED ISSUE #1: Draw representative periods from all weather years (or reproduce the all-year average)
With the `GSw_HourlyWeatherYears` switch we have the ability to select representative periods from a single year (the default 2012) or the full 2007-2013 + 2016-2023 range of years used for RA. I think we should switch to the full range by default.

Here's a comparison between 2012 and 2007-2013 for reference:

Image

The largest single-site, single-year error in PV CF relative to the 2007–2013 CF is 3.9 %, roughly equal to the difference in state-wide average PV CF between Maryland and Florida; the largest single-site, single-year error in wind CF relative to the 2007–2013 CF is 9.6 %, roughly the difference in state-wide average wind CF between Texas and Mississippi

Some pros of switching to all years:
* It would give more representative wind/solar average CFs and average load
* It would give us electrolyzer/DAC electricity demand profiles for all RA years, which would make the PRAS analysis more robust
* It would let us assign probabilities to stress periods, since every day in all the RA years would be matched to a representative day. This might be the biggest benefit; then we should be able to drop the min-CF constraint

Some cons:
* Selecting representative periods from all RA years is quite a bit slower than selecting from just a single year (a couple hours instead of ~15 minutes). So we'd probably want to pre-select and cache the default representative periods to speed up runtime.
    * Though maybe that's not a con, since folks have noted that scenario-dependent rep days can complicate scenario comparison anyway
* If we only run PLEXOS for a single year, the results might not line up as well with ReEDS
* If someone wants to use the old single-weather-year EFS profiles, the load and VRE wouldn't line up. (But they could always switch back to drawing from just 2012 for their use case.)
* GSw_H2=2 is already slow for a single year, so it'd be much slower for 15 years. We might need to implement an even-lower-temporal-resolution version where H2 capacity is only tracked by week instead of by day.

## TODO
- [x] Speed up representative day selection when drawing from multiple years (heuristic approach implemented on pb/rep15)
    - [ ] Consider additional constraints on rep day assignment (like winter days can only come from winter)?
    - [ ] Try HiGHS instead of CBC
    - [x] #105 
- [ ] Cache representative periods for different combinations of demand/wind/solar
- [ ] Figure out how to speed up interseasonal H2
  - [ ] Try larger H2 balancing regions (pb/h2r)
  - [ ] Try 5-day balancing instead of 1-day balancing
- Try a few other things:
  - [ ] Set `GSw_H2` to 0 and see how different the results are; if not different, consider setting it to 0 by default for now
  - [ ] Only dispatch seasonal storage for a subset of the full representative years
- [ ] **Select rep days from a single weather year but match the all-year regional averages**

### 20260127 update
The plan is to implement the final bullet above: Use a single year for the basis set of representative days, but weight/match them to the multi-year average CF. That will let us decouple the average CF from the need to model multi-year chronology in ReEDS.

We plan to use the latest year for which data are available (currently 2023) as the basis year. Once implemented, that year should probably be the default for downstream PCM runs as well.

- [ ] Update `inputs_cfmapdist_*.png` and `inputs_loadmap*.png` to show the difference from the multi-year average
- [ ] Test a few different years in the PCM

patrickbrown4:
## Summary

When using 7 or 15 weather years in `GSw_HourlyWeatherYears`, the second stage of the representative period selection (mapping actual to representative periods using a MILP) is exceedingly slow. This PR adds an alternative method that runs in seconds and produces actual-to-representative period maps that look pretty much the same.

The defaults are unchanged, but the new method can be used in the meantime until we decide what to do by default.

### Implementation notes 

The new method, implemented in `reeds.timeseries.match_act2rep_bestfirst()`, works as follows:

1. Begin with a table of weights for each representative (rep) period
2. Calculate the distance in (feature × region)-dimensional space between each actual and rep period.
3. While there is still weight remaining in the rep-period weight table:
    a. Match each rep period to the single remaining actual period that is closest to it, recording its match in the actual-to-rep-period table
    b. Decrement the remaining weight for each rep period. If any weights reach zero, remove them from future consideration for matches. (So the lowest-weighted rep periods will fill up and drop out of consideration first.)
    c. Remove the actual periods that were matched from the table of distances
    d. Return to 3.a. and repeat until there are no remaining weights in the rep-period weight table (and no remaining unmatched actual periods)

### Additional changes 

- I moved most of the portable functions from `reeds.inputs_processing.hourly_repperiods` into `reeds.timeseries`
  - Reworked a few to cut down on code duplication
- Tweaked `reedsplots.plot_repdays()` to include the year when `GSw_HourlyWeatherYears` has more than one weather year
- Moved the determination of `nextszn` from `2_temporal_params.gms` to `hourly_repperiods.py`, which cuts off roughly 1 hour of solve time per model year when using 7-15 weather years
- Changed the `Pacific_rep7` test case to `Pacific_rep15`, which uses the new method
- Addressed #109 by reducing the `round_to_integers()` increment

- `GSw_HourlyClusterMapMethod` (default `milp`): How to map actual periods to representative periods: 'milp' minimizes sum of absolute errors but is slow; 'bestfirst' is orders of magnitude faster when using many weather years in `GSw_HourlyWeatherYears`

### Issues resolved 

- Partially resolves #1 (but does not change the default yet)

### Known incompatibilities 

- 7- or 15-year runs are still slow, but can be accelerated by setting:
  - `GSw_H2` = 0
  - `GSw_H2_PTC` = 0
  - `GSw_StartCost` = 0

Compare report is here: [20260522 - rep15.pptx](https://nrel.sharepoint.com/:p:/s/ReEDS/IQB2DpoYzyr1QqMxf97nzkMGATP2_Drh1LCbg_HWqYJxaZQ?e=6nC1Zb)

- Changes to the USA_defaults case are within the noise (≤0.05 GW, ≤0.01 $B) (nonzero change is expected because I rounded some temporal parameters that were not rounded before)
- If we switch to 15 weather years without changing anything else, the model runs into an infeasibility in 2032. There's no conflict info though; all I get is the following message.
    ```
    Infeasibility model not feasible; tightening tolerances.
    Infeasibility model infeasible!
    Problem is feasible; no conflict available.
    ```
- If GSw_StartCost=0, GSw_H2=0, and GSw_H2_PTC=0, the 15-year case solves in 12 hours. A comparison to the default 1-year case with the same settings is included in the compare report linked above.

### Details to double-check

- [x] Documentation updated if necessary (if/when we change the default, I'll add a description of the new method to `model_documentation.md`)
  
No


clairehalloran (commented):
It would be helpful to add a brief description of the bestfirst method in the "Representative periods" section of the docs and/or the "Temporal resolution options" box.

reeds/timeseries.py
```diff
@@ -105,3 +123,327 @@ def timeslice_to_timestamp(case, param):
...
+    assignments = pd.Series(
+        {(a,r):WEIGHT[a,r].varValue for a in actualdays for r in repdays}).astype(int)
+    assignments.index = assignments.index.rename(['act','rep'])
+    a2r = assignments.replace(0,np.nan).dropna().reset_index(level='rep').rep
+
+    return a2r
+
+
+def match_act2rep_bestfirst(profiles_period_mean, iweights, metric='euclidean'):
+    """
```
clairehalloran: Using the Euclidean distance here is consistent with the other clustering metrics, but I'm curious if it would be more appropriate to use another metric in 18 transgrp x 3 feature-dimensional space. It seems like having similar distances between a rep day and many actual days could lead to poor approximations of the optimal mapping. Do you see significant differences in capacity or generation between the MILP and best-first results for a single year? I see that the same rep days are mapped to slightly different actual days but have no idea how much this might matter.

clairehalloran (commented):
Pulling representative days from multiple years leads to a blank plot in [plot_dispatch_yearbymonth](https://github.com/ReEDS-Model/ReEDS/blob/fa4c6dc84a87ff8a2af1dd90875ac6482f34c792/reeds/reedsplots.py#L3568) when `highlight_rep_periods=1`. It may be better to copy the year logic you added for `plot_repdays()` here.

clairehalloran (changes requested):
Overall, a really helpful addition for representing a broader range of weather conditions! Happy to approve with a short addition to the docs and a bugfix for plot_dispatch_yearbymonth(). 

I'm curious how much the mapping algorithm impacts results for a single year, and if a metric other than Euclidean distance could approximate the MILP more closely.

reeds/timeseries.py
patrickbrown4: Ah sorry, I forgot to include that pair of runs in the compare report. They're added now, slides 28-43. The capacity/generation/cost differences are quite small. That makes sense to me, since the only things the 1:1 actual-to-representative period mapping matters for are interperiod H2 storage (of which there's very little in the default-case solution) and linearized startup costs (which only apply to coal, CCS, and gas-cc).

image
image
image

patrickbrown4: On the specifics of the distance calculation, there are a number of other options in the [`scipy.spatial.distance.cdist()`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.cdist.html) function, but I only tried `euclidean`. Are there any others you'd recommend? Maybe `sqeuclidean` (squared Euclidean), to penalize very bad matches more?

(Since it's not used by default and I'm not sure if we'll end up using this method or the reproduce-many-year-average-with-one-year method, I thought it'd be best to leave it as a keyword argument instead of elevating it to a `cases.csv` switch.)

patrickbrown4: I've added the results for a few alternative distance metrics to the deck, starting on slide 44. To me, `seuclidean` (standardized Euclidean) seems closest to the MILP version, so I've made that the new default.

patrickbrown4:
> Overall, a really helpful addition for representing a broader range of weather conditions! Happy to approve with a short addition to the docs and a bugfix for plot_dispatch_yearbymonth().
> 
> I'm curious how much the mapping algorithm impacts results for a single year, and if a metric other than Euclidean distance could approximate the MILP more closely.

Cool, thanks for catching these. I've added a note to the "Temporal resolution options" box in the docs (I figure we can keep it there unless we switch to this method by default, at which point we can replace the description it in the main body), and I've adapted `plot_dispatch_yearbymonth()` to work for multiple weather years, shown below. (It's now a generator, which makes it a bit more complicated than a typical plot, but it saves time since it's slow to format the hourly dispatch within the function.)

I've also added some alternative distance metrics to the [compare report](https://nrel.sharepoint.com/:p:/s/ReEDS/IQB2DpoYzyr1QqMxf97nzkMGATP2_Drh1LCbg_HWqYJxaZQ?e=6nC1Zb) and settled on the standardized Euclidean distance instead of pure Euclidean, which to me looks a little closer to the MILP.

plot_dispatch-yearbymonth-1-2032-w2007.png
plot_dispatch-yearbymonth-1-2032-w2007
plot_dispatch-yearbymonth-1-2032-w2008.png
plot_dispatch-yearbymonth-1-2032-w2008

reeds/timeseries.py
clairehalloran: Thanks for adding the alternative distances to the deck! I agree, `seuclidean` looks the closest to me too. I'm happy with leaving this as a keyword argument instead of a `cases.csv` switch.

clairehalloran (approved):
Thanks for making these changes! Looks good to me now.

PR #107, opened 2026-05-26
Bc cvar updates

bcakire:
## Summary

This PR adds CVAR/NCVAR reporting and target checks to the PRAS resource adequacy workflow.

This implementation builds on the `aa/metrics_extension` / MultiMetricRA workflow. Standard RA metrics are still used for stress-period selection, while CVAR/NCVAR are added as check-only metrics.

Main idea is:

- PRAS can write CVAR/NCVAR-related outputs.
- Python can read sample-level PRAS shortfall outputs.
- CVAR/NCVAR can be calculated by hierarchy level.
- CVAR/NCVAR can be checked against user-defined thresholds.
- For now, CVAR/NCVAR are check-only metrics:
  - they do not add stress periods,
  - they do not update PRM,
  - they do not replace the standard MultiMetricRA stress-period metrics.

Standard stress-period metrics remain:

```text
NEUE / LOLH / LOLE / LOLD / Duration / Depth
```

CVAR/NCVAR are evaluated separately after the standard RA metrics are written.

### `run_pras.jl`

Added:

- `--cvar_alpha`
- `--write_shortfall_samples_totals`

When `PRAS.ShortfallSamples()` is available, `run_pras.jl` can now write:

- `PRAS_{t}i{iteration}-risk_metrics.csv`
- `PRAS_{t}i{iteration}-shortfall_totals_by_sample.h5`

`risk_metrics.csv` includes CVAR/NCVAR-related PRAS risk metrics such as:

- CVAR in MWh
- NCVAR in ppm
- alpha
- estimate
- standard error
- VaR

`shortfall_totals_by_sample.h5` includes total shortfall by PRAS sample for USA and regions. These values are totals over the full PRAS resource adequacy period, not annual values.

### `stress_periods.py`

Added CVAR/NCVAR handling with:

- `get_cvar_alpha()`
- `get_shortfall_totals_by_sample()`
- `_sample_cvar()`
- `get_annual_cvar_stress_metric()`
- `evaluate_cvar_target_check()`

CVAR/NCVAR are calculated by hierarchy level.

NCVAR is normalized by total PRAS load and reported in ppm.

CVAR is written as an annualized MWh value.

The standard `ra_metrics_{t}i{iteration}.csv` output remains focused on the MultiMetricRA stress-period metrics. CVAR/NCVAR are written separately as:

- `cvar_{t}i{iteration}.csv`
- `ncvar_{t}i{iteration}.csv`
- `cvar_target_check_{t}i{iteration}.csv`

CVAR/NCVAR are intentionally kept out of the normal stress-period selection loop. They are only used for reporting and target checks.

### `ra_calcs.py`

Added logic so `write_shortfall_samples_totals` is turned on automatically when CVAR/NCVAR checks are requested.

This output is also kept on when PRAS-informed PRM updates need sample-level shortfall totals.

### `runreeds.py`

Updated compatibility checks so `CVAR` and `NCVAR` are allowed in:

```text
GSw_PRM_StressThresholdMetrics
```

but treated as check-only metrics instead of standard stress-period metrics.

### Switches / cases

Added or updated these switches:

- `GSw_PRM_CVARAlpha`
- `GSw_PRM_StressThresholdCVAR`
- `GSw_PRM_StressThresholdNCVAR`

Updated the description of:

- `GSw_PRM_StressThresholdMetrics`

so it includes the standard MultiMetricRA stress-period metrics and the new check-only CVAR/NCVAR metrics.

The standard MultiMetricRA threshold format follows the `aa/metrics_extension` format:

```text
GSw_PRM_StressThresholdNEUE = transgrp_1
GSw_PRM_StressThresholdLOLH = transgrp_2.4
GSw_PRM_StressThresholdLOLE = transgrp_0.1
GSw_PRM_StressThresholdLOLD = transgrp_0.1
GSw_PRM_StressThresholdDuration = transgrp_12
GSw_PRM_StressThresholdDepth = transgrp_0.1
```

CVAR/NCVAR threshold examples:

```text
GSw_PRM_StressThresholdCVAR = country_50000_CVAR_cvar
GSw_PRM_StressThresholdNCVAR = transgrp_10_NCVAR_cvar
```

Example metric list:

```text
GSw_PRM_StressThresholdMetrics = NEUE/LOLH/LOLE/LOLD/Duration/Depth/CVAR/NCVAR
```

The default remains unchanged unless CVAR/NCVAR are explicitly requested.

## Testing

Added and ran a MultiMetricRA + CVAR/NCVAR test case instead of relying on the Pacific-only test case.

Test case:

```text
NY_CVAR_debug_merged_MultiMetricRA
```

The case produced the expected standard MultiMetricRA outputs:

- `ra_metrics_{t}i{iteration}.csv`
- `eue_events_{t}i{iteration}.csv`

and the new CVAR/NCVAR check-only outputs:

- `cvar_{t}i{iteration}.csv`
- `ncvar_{t}i{iteration}.csv`
- `cvar_target_check_{t}i{iteration}.csv`

Example result interpretation from the test case:

- The country-level CVAR target passed because the annualized absolute tail-risk shortfall was below the MWh threshold.
- The transgrp-level NCVAR target failed because the normalized tail-risk shortfall exceeded the ppm threshold.
- This confirms that CVAR/NCVAR target checks are being evaluated and reported separately from stress-period selection.

Because CVAR/NCVAR are check-only in this PR, a failed CVAR/NCVAR target check does not add stress periods or trigger a PRM update.

### Details to double-check

- [X] Charge code provided to reviewers
  

  - [ ] Units are specified
  - [ ] If spatially resolved inputs are modified, the following visualizations for each file are included in the PR description, time-averaged if the inputs are time-resolved:
    - [ ] Map of differences: after - before or after / before

- [X] Code formatting standardized 
- [X] Reusable functions used where possible instead of copy/pasted code

- [X] Zero impact on results of default case
- [X] No large data file(s) added/modified
- [X] No substantive impact on runtime for full-US reference case when CVAR/NCVAR are not enabled
- [X] No substantive impact on folder size for full-US reference case when CVAR/NCVAR are not enabled
- [ ] No change to process flow (`runreeds.py`, `reeds/core/solve/solve.py`)
- [X] No change to package requirements (`environment.yml` or `Project.toml`)

Used ChatGPT to help draft PR text and troubleshoot merge-conflict resolution. Code changes and test results were reviewed by the author.


reeds/hpc/srun_template.sh
bsergi, line 1: revert the changes to this file before merging

reeds/resource_adequacy/run_pras.jl
```diff
@@ -70,10 +71,20 @@ function parse_commandline()
             default = 0
             required = false
         "--write_shortfall_samples"
-            help = "Write the sample-level shortfall"
+            help = "Write the sample-level shortfall (hourly, large)"
```
bsergi: What are the large and small designations? I might suggest just "Write per-sample hourly shortfall by region" for this one and "Write per-sample total shortfall by region" for the one below.

reeds/resource_adequacy/run_pras.jl
```diff
@@ -70,10 +71,20 @@ function parse_commandline()
...
             arg_type = Int
             default = 0
             required = false
+        "--write_shortfall_samples_totals"
+            help = "Write per-sample total shortfall by region (small)"
```
bsergi: this is total over the entire PRAS time period and not annual, right?

reeds/resource_adequacy/run_pras.jl
```diff
@@ -208,6 +219,34 @@ function run_pras(pras_system_path::String, args::Dict)
     @info "$(PRAS.EUE(results["short"]))"
     @info "$(PRAS.NEUE(results["short"]))"
 
+    # CVAR results
+    if haskey(results, "short_samples")
+        _,_,_,_,energyunit = PRAS.get_params(sys)
+        alpha = Float64(args["cvar_alpha"])
+        cvar_obj = PRAS.CVAR(energyunit, results["short_samples"], alpha)
+        @info(cvar_obj)
```
bsergi: I think PRAS is setup so that the above info statements provide context (e.g., `@info "$(PRAS.EUE(results["short"]))"` results in `EUE = 851000±5000 MWh/131400h MWh)` being printed). Does this do something similar?

reeds/resource_adequacy/run_pras.jl
```diff
@@ -302,31 +342,39 @@ function run_pras(pras_system_path::String, args::Dict)
...
-            dictshort[s] = DF.DataFrame(
-                transpose(getindex.(results["short_samples"][:, :], s)),
-                sys.regions.names
-            )
-            # subset to regions (filter out DC regions) 
-            dictshort[s] = dictshort[s][:,findall(regions .∈ Ref(sys.regions.names))]
-        end
-        ## Write it
+        sf = results["short_samples"]
+        region_names = sf.regions.names
```
bsergi, line 352: the previous code filtered using `sys.regions.names`. I think this was because there are some pseudo regions for DC converter stations that don't have load and thus don't need to be included here, so we may want to continue to drop them (@patrickbrown4 might be able to weigh in on this one).

reeds/resource_adequacy/run_pras.jl
```diff
@@ -302,31 +342,39 @@ function run_pras(pras_system_path::String, args::Dict)
         @info("Wrote PRAS storage energy to $(energyfile)")
     end
 
-    ### Sample-level shortfall
+    ### Sample-level shortfall (hourly, large)
```
bsergi: See previous note about the "large" flag. I think this is a reference to the size of the data file, but I'd opt for leaving it out and just describing the dimensions.

reeds/resource_adequacy/run_pras.jl
```diff
@@ -208,6 +219,34 @@ function run_pras(pras_system_path::String, args::Dict)
...
     @info "$(PRAS.NEUE(results["short"]))"
 
+    # CVAR results
+    if haskey(results, "short_samples")
+        _,_,_,_,energyunit = PRAS.get_params(sys)
+        alpha = Float64(args["cvar_alpha"])
+        cvar_obj = PRAS.CVAR(energyunit, results["short_samples"], alpha)
+        @info(cvar_obj)
+
+        total_load = sum(sys.regions.load)
```
bsergi, line 237: Can you add a comment describing the calculation of these normalized metrics?

cases_test.csv
bsergi, line 1: let's revert the changes back on this file

load_atm.sh
bsergi, line 1: remove this file

runreeds.py
bsergi, line 1: revert the changes on this file before merging

reeds/resource_adequacy/stress_periods.py
```diff
@@ -204,6 +206,115 @@ def get_and_write_neue(sw, write=True):
         eue.to_csv(os.path.join(sw['casedir'],'outputs','eue.csv'))
     return neue
 
+def get_cvar_alpha(sw):
+    val = sw.get('GSw_PRM_CVARAlpha', 0.95)
+    if pd.isna(val) or str(val).strip().lower() in ['', 'none', 'nan']:
```
bsergi: I think compatibility should be handled with the `float` type check from the cases file so may not be needed here. In any case, I worry a little bit about falling silently back to a default, so if you do want to keep this it might be worth a print statement flagging that this is happening.

reeds/resource_adequacy/stress_periods.py
```diff
@@ -204,6 +206,115 @@ def get_and_write_neue(sw, write=True):
...
+        return 0.95
+    return float(val)
+
+
+def get_shortfall_totals_by_sample(case, t, iteration=0):
+    filepath = os.path.join(case, 'handoff', 'PRAS', f'PRAS_{t}i{iteration}-shortfall_totals_by_sample.h5')
+    if not os.path.isfile(filepath):
+        raise FileNotFoundError(f"{filepath} not found. Re-run PRAS with --write_shortfall_samples_totals 1.")
```
bsergi, line 157: Do we need some logic in the ReEDS call to PRAS that turns this on when needed? I'm thinking of something along the lines of what we have here: https://github.com/ReEDS-Model/ReEDS/blob/1a8ab631bb7bc20b0e97cb98a8d6106fa8f29344/reeds/resource_adequacy/ra_calcs.py#L161

reeds/resource_adequacy/stress_periods.py
```diff
@@ -204,6 +206,115 @@ def get_and_write_neue(sw, write=True):
...
+    x = pd.Series(samples).dropna().astype(float)
+    if x.empty:
+        return np.nan
+    if not (0 <= alpha < 1):
+        raise ValueError(f"CVaR alpha must be in [0, 1). Got alpha={alpha}")
+    n_tail = max(1, int(np.ceil((1 - alpha) * len(x))))
+    return x.sort_values(ascending=False).iloc[:n_tail].mean()
+
+
+def get_annual_cvar_stress_metric(case, t, stress_metric='NCVAR', iteration=0, alpha=0.95):
```
bsergi: this function and the next one get pretty long, so it would be good to add comments throughout to provide some documentation on what they are doing.

reeds/resource_adequacy/stress_periods.py
```diff
@@ -204,6 +206,115 @@ def get_and_write_neue(sw, write=True):
...
+    elif df.index.name != 'sample':
+        df.index = pd.RangeIndex(1, len(df) + 1, name='sample')
+    return df.apply(pd.to_numeric, errors='coerce').clip(lower=0)
+
+
+def _sample_cvar(samples, alpha=0.95):
+    x = pd.Series(samples).dropna().astype(float)
+    if x.empty:
+        return np.nan
+    if not (0 <= alpha < 1):
```
bsergi: Feels like this check might fit better in `get_cvar_alpha`. An even better approach might be to add a check before the ReEDS run which would avoid having runs get this far and then running into an error (best place for that would probably be here: https://github.com/ReEDS-Model/ReEDS/blob/1a8ab631bb7bc20b0e97cb98a8d6106fa8f29344/runreeds.py#L222)

reeds/resource_adequacy/stress_periods.py
```diff
@@ -204,6 +206,115 @@ def get_and_write_neue(sw, write=True):
...
+        raise ValueError(f"CVaR alpha must be in [0, 1). Got alpha={alpha}")
+    n_tail = max(1, int(np.ceil((1 - alpha) * len(x))))
+    return x.sort_values(ascending=False).iloc[:n_tail].mean()
+
+
+def get_annual_cvar_stress_metric(case, t, stress_metric='NCVAR', iteration=0, alpha=0.95):
+    stress_metric = stress_metric.upper()
+    if stress_metric not in CVAR_METRICS:
```
bsergi: The multi-metrics PR (#99) now drops the metric tag in the switch value; for example, `GSw_PRM_StressThresholdNEUE` now just has a value of `transgrp_1_sum` instead of `transgrp_1_EUE_sum` (see [here](https://github.com/ReEDS-Model/ReEDS/pull/99/changes#diff-f299c7ed032ed6017f03b570d645235760239ba8455e86ac78da955e29431910R275)). I think we could do the same in this PR and then you could drop the `CVAR_METRICS` and the check and just iterate over whichever switches are activated.

reeds/resource_adequacy/stress_periods.py
```diff
@@ -414,27 +529,18 @@ def get_stress_metrics_sorted_periods(sw, t, iteration):
...
-                                                                   criterion,
-                                                                   sw,
-                                                                   t,
-                                                                   iteration,
-                                                                   dfenergy_r,
-                                                                   stressperiods_this_iteration,
-                                                                   )
+        for criterion in sw[f'GSw_PRM_StressThreshold{metric}'].split('/'):
+            print(f"Evaluating GSw_PRM_StressThreshold {metric} with criterion: {criterion}")
+            stress_criteria = _evaluate_stress_threshold_criterion(stress_criteria, criterion, sw, t, iteration, dfenergy_r, stressperiods_this_iteration,
```
bsergi: This line looks like just a formatting change--I'd suggest keeping as the original had it.

reeds/resource_adequacy/stress_periods.py
```diff
@@ -204,6 +206,115 @@ def get_and_write_neue(sw, write=True):
...
+            load_regions = [c for c in dfload.columns if c in rmap.index]
+            load_agg = dfload[load_regions].rename(columns=rmap).groupby(axis=1, level=0).sum().sum()
+            cvar = cvar / load_agg.reindex(cvar.index) * 1e6
+
+        _metric[hierarchy_level,'cvar'] = cvar
+
+    return pd.concat(_metric, names=['level','metric','region']).rename(stress_metric)
+
+
+def evaluate_cvar_target_check(sw, t, iteration=0, stress_metrics=None):
```
bsergi: Much of this function, including the core check of `failed = this_test.loc[this_test > threshold]`, seems to replicate behavior in `_evaluate_stress_threshold_criterion`. Do you think we could merge the CVAR treatment into that existing function and just have some control statements to turn off adding new stress periods for now when using CVAR metrics? It's possible having a second function makes the most sense here, but I worry a little bit about maintaining two similar functions.

bsergi (commented):
Here are some initial thoughts from a quick look. Happy to touch base when you get back to discuss any of these. Let me know when you've taken another pass here and I can review again.

One high-level comment: can you organize the PR summary text a bit more? The technical details section has a lot of information but is a bit difficult to follow. Also, I didn't quite understand your test output `GSw_PRM_StressThreshold = country_-1_NCVAR_cvar failed`. What does it mean to have a -1 threshold? Ultimately for the PR we'll also want a comparison of default case from `cases.csv` from both this branch and main to show there weren't any changes.

PR #120, opened 2026-06-11
Bugfix: Fix handling of longitudes for existing plants in hourlize

patrickbrown4:
## Summary

Jarrad noted that the regional distribution of our existing wind capacity is wrong.
- Here's existing EIA-NEMS 2025 capacity:
  - image
- Here's Texas wind capacity from USGS:
  - image
- But here's 2023 wind capacity from ReEDS:
  - image
- All the exogenous and prescribed wind and PV capacity from ReEDS was smushed to the east side of each zone:
  - image
  - image
- The problem was a `df_exist['LONG'] = df_exist['LONG'] * -1` line in `hourlize/resource.py`. At some point (I'm not sure when) we switched from positive (wrong) longitudes to negative (correct) longitudes in the EIA-NEMS database, but did not update that line in hourlize.
  - This PR removes that line (instead raising an error if any longitudes are positive) and updates the exogenous and prescribed capacities.

### Additional changes

- Added `reedsplots.validate_regional_capacity()`, which now runs at the start of `single_case_plots.py`
  - Here it is before:
    - 20260609-validate_cap_regional-FIPS
  - And after:
    - validate_cap_regional-FIPS
- Added a compatibility check to `runreeds.py` to raise an error earlier if using 7-year load data (EER2023) with 15-year VRE data

Exogenous and prescribed capacities now look more right:
image
image

The 2023 Texas wind capacity out of ReEDS also looks more right:
image

Here's a compare report for the USA_defaults case: [20260612 - longitude.pptx](https://nrel.sharepoint.com/:p:/s/ReEDS/IQCalFvz1TW_RJ_hXkpT5prjAVhSi68qSerpUJxcP8_m8Gw?e=qgSXdq)
- Wind capacity, transmission capacity, curtailment, and cost through the 2030s are lower
  - Which makes sense if existing wind and transmission are in relatively economically optimal places

### Details to double-check

  - [x] If spatially resolved inputs are modified, the following visualizations for each file are included in the PR description (time-averaged if the inputs are time-resolved):
    - [x] Map of absolute values before
    - [x] Map of absolute values after

No


bsergi (approved):
Looks good--thanks for the fix. The validation plot is a nice addition.

I did a little sleuthing and it seems like this may have been in as early as [this commit](https://github.nrel.gov/ReEDS/ReEDS-2.0/commit/b69e1104043778b8cbdaf659c2cf46666d2e2416) (that one actually has both positive/negative values) from 2023, so it seems to have been in there for a while.

PR #124, opened 2026-06-18
Stress metrics extension

abdelrahman-ayad:
## Summary

- This PR extends the Resource Adequacy multi-metrics framework introduced in #99 by adding four new stress metrics: `Duration`, `Depth`, `LOLD`, and `LOLE`.
- Users can include any combination of these metrics in `GSw_PRM_StressThresholdMetrics` alongside the existing metrics (`NEUE`, `LOLH`).
- Stress metric definitions are adopted from: [Gord et al., 2022. Clarifying the Interpretation and Use
of the LOLE Resource Adequacy Metric](https://spp.org/documents/69303/lole%20metric.pdf)

### Implementation notes

- All new metrics are computed in `calc_ra_metrics()` after reading PRAS outputs. `Duration` and `Depth` derive from the hourly EUE timeseries via two helper functions:
  - `calc_max_duration(dfeue_agg)` — length of the longest consecutive run of hours with EUE above threshold
  - `calc_peak_eue(dfeue_agg, dfload_agg, norm='peak')` — peak hourly EUE normalized by peak load
- `LOLD` and `LOLE` use the PRAS `_LOLE` hourly columns (same source as `LOLH`):
  - `LOLD` (`calc_lold`) — a day is counted as an event-day if at least one hour has LOLE > 0
  - `LOLE` (`calc_lole`) — number of loss-of-load events, where an event is a contiguous run of hours with LOLE > 0
- Note: `GSw_PRM_StressThresholdOutageMagnitude` (a raw-MW magnitude threshold) was dropped in favor of keeping only the normalized `Depth` metric, since an absolute-MW threshold isn't comparable across transgrps of very different sizes (same reasoning as the existing EUE-vs-NEUE tradeoff). `GSw_PRM_StressThresholdEUE` was also removed as a standalone switch; its role is now split between `Duration` and `Depth`.
- Also fixed: `update_prm()` (used when `GSw_PRM_UpdateMethod` is 1, 2, or 3) still assumed the pre-multi-metric shape of the failed-criteria data and would crash (`KeyError`) as soon as any threshold failed. It's now updated to correctly select NEUE-based failures from the new multi-metric structure — since the PRM update math is inherently NEUE/ppm-based — and to no-op (leave the PRM unchanged) when NEUE didn't fail even though another metric did. Added a companion check in `runreeds.check_compatibility()` that raises a clear error if `GSw_PRM_UpdateMethod` is 1-3 but `NEUE` isn't included in `GSw_PRM_StressThresholdMetrics`.

### New stress metric switches

| Metric | Switch | Definition | RA dimension | PRAS input | Example |
|--------|--------|------------|--------------|------------|---------|
| `Duration` | `GSw_PRM_StressThresholdDuration` | Expected (average) length of the longest consecutive sequence of shortfall hours per year [h] | Duration | EUE | `transgrp_12`: threshold at 12 hours, per `transgrp` |
| `Depth` | `GSw_PRM_StressThresholdDepth` | Expected (average) peak single-hour shortfall normalized by peak load [MW EUE / MW peak demand] | Magnitude | EUE + load | `transgrp_0.1`: threshold at 0.1 (10% of peak load), per `transgrp` |
| `LOLD` | `GSw_PRM_StressThresholdLOLD` | Expected (average) count of event-days per year; a day is counted if at least one hour experiences a shortfall [event-days/year] | Frequency | LOLE | `transgrp_0.1`: threshold at 0.1 event-days/year, per `transgrp` |
| `LOLE` | `GSw_PRM_StressThresholdLOLE` | Expected (average) count of loss-of-load events per year, where an event is a contiguous run of shortfall hours [events/year] | Frequency | LOLE | `transgrp_0.1`: threshold at 0.1 events/year, per `transgrp` |

Zero change for [USA_defaults](https://github.com/user-attachments/files/30233474/results-v20260719_903843eb_USA_defaults.v20260719_stressK0_USA_defaults.pptx), [Mid_Case](https://github.com/user-attachments/files/30233521/results-v20260719_903843eb_Mid_Case.v20260719_stressK0_Mid_Case.pptx), or [USA_decarb](https://github.com/user-attachments/files/30233590/results-v20260719_903843eb_USA_decarb.v20260719_stressK0_USA_decarb.pptx).

There's a small change if LOLD=0.1 (1 event-day in 10 years) is included, so that would be good to do in a followup PR: [results-v20260719_903843eb_USA_defaults,v20260719_stressK0_USA_NEUE_LOLD.pptx](https://github.com/user-attachments/files/30233801/results-v20260719_903843eb_USA_defaults.v20260719_stressK0_USA_NEUE_LOLD.pptx)

- [x] USA (combined: NEUE/LOLH/LOLD/LOLE/Duration/Depth)
Comparison report: 
[Uploading results-defaults,main_defaults,multimetric.pptx…]()

### Details to double-check

- I used Claude Code to implement the metric calculations, review the branch/PR, and fix a bug in `update_prm()` uncovered during that review (see Implementation notes).


cases.csv
```diff
@@ -263,14 +263,20 @@ GSw_PRM_NetImportLimit,Turn on (1) or off (0) the eq_firm_transfer_limit constra
...
+GSw_PRM_StressThresholdMetrics,"/-delimited list of metrics for identifying stress periods (supported options are: NEUE, LOLH)",N/A,NEUE,
+GSw_PRM_StressThresholdLOLH,LOLH threshold [hours/year] above which to re-solve the latest model year with new stress periods; formulated as HierarchyLevel_LOLH_PeriodAggMethod where HierarchyLevel is a column in hierarchy.csv; LOLH is loss of load event-hours per year [event-h/year]; PeriodAggMethod is 'sum' or 'max' over the hours in each period (only used in period selection) (see README.md for detailed notes),N/A,transgrp_2.4_sum,
+GSw_PRM_StressThresholdLOLE,LOLE threshold [event-days/year] above which to re-solve the latest model year with new stress periods; formulated as HierarchyLevel_LOLEdays_PeriodAggMethod where HierarchyLevel is a column in hierarchy.csv; LOLEdays is loss of load event-days per year (a day is counted if at least one hour has a shortfall) [event-day/year]; PeriodAggMethod is 'sum' or 'max' over the hours in each period (only used in period selection) (see README.md for detailed notes),N/A,transgrp_0.1_max,
```
patrickbrown4: I've seen different definitions used in different sources, but at least in some places, "LOLE" is taken to mean "events" and LOLD as "event-days". So if the units used here are "event-days", then it might be clearer to label it as LOLD.

Here's the figure I'm thinking of, from <https://doi.org/10.1109/PMAPS53380.2022.9810615>:

image

cases.csv
```diff
@@ -263,14 +263,20 @@ GSw_PRM_NetImportLimit,Turn on (1) or off (0) the eq_firm_transfer_limit constra
...
+GSw_PRM_StressThresholdNEUE,Annual NEUE level [ppm] threshold above which to re-solve the latest model year with new stress periods; formulated as HierarchyLevel_NEUEppm_PeriodAggMethod where HierarchyLevel is a column in hierarchy.csv; NEUEppm is normalized expected unserved energy in parts per million [ppm]; PeriodAggMethod is 'sum' or 'max' over the hours in each period (only used in period selection) (see README.md for detailed notes),N/A,transgrp_1_sum,  
+GSw_PRM_StressThresholdOutageDuration,Outage duration; formulated as HierarchyLevel_OutageDurationHours_PeriodAggMethod where HierarchyLevel is a column in hierarchy.csv; OutageDurationHours is the max outage duration in hours; PeriodAggMethod is 'max' over the hours in each period (only used in period selection) (see README.md for detailed notes),N/A,transgrp_10000_max,
```
patrickbrown4: 1. 10000 hours is an extremely long event. I think it'd be most convenient to set the defaults to values in the range of what's come up in the literature review. If I'm interpreting correctly, the values include 8 hours for NWPCC, 12 hours for ERCOT, and 18 hours for ISONE. So how about 12 hours?
2. Just to keep the switch name shorter, you could change it to `GSw_PRM_StressThresholdDuration`.

```suggestion
GSw_PRM_StressThresholdDuration,Outage duration; formulated as HierarchyLevel_OutageDurationHours_PeriodAggMethod where HierarchyLevel is a column in hierarchy.csv; OutageDurationHours is the max outage duration in hours; PeriodAggMethod is 'max' over the hours in each period (only used in period selection) (see README.md for detailed notes),N/A,transgrp_12_max,  
```

cases.csv
```diff
@@ -263,14 +263,20 @@ GSw_PRM_NetImportLimit,Turn on (1) or off (0) the eq_firm_transfer_limit constra
...
+GSw_PRM_StressThresholdOutageDuration,Outage duration; formulated as HierarchyLevel_OutageDurationHours_PeriodAggMethod where HierarchyLevel is a column in hierarchy.csv; OutageDurationHours is the max outage duration in hours; PeriodAggMethod is 'max' over the hours in each period (only used in period selection) (see README.md for detailed notes),N/A,transgrp_10000_max,  
+GSw_PRM_StressThresholdOutageMagnitude,Outage magnitude; formulated as HierarchyLevel_OutageMagnitudeMW_PeriodAggMethod where HierarchyLevel is a column in hierarchy.csv; OutageMagnitudeMW is the max outage magnitude in MW; PeriodAggMethod 'max' over the hours in each period (only used in period selection) (see README.md for detailed notes),N/A,transgrp_0.1_max,  
+GSw_PRM_StressThresholdNormalizedOutageMagnitude,Normalized outage magnitude; formulated as HierarchyLevel_NormalizedOutageMagnitude_PeriodAggMethod where HierarchyLevel is a column in hierarchy.csv; NormalizedOutageMagnitude is the max outage normalized magnitude (ratio of the maxmimum load) ; PeriodAggMethod is 'max' over the hours in each period (only used in period selection) (see README.md for detailed notes),N/A,transgrp_10_max,
```
patrickbrown4: 1. I can't remember how much we discussed this, but in the same vein as the earlier discussion on the EUE metric, it'd be hard to apply a single magnitude threshold in units of MW across different transgrps (with very different peak loads) and over time. In the same way that we supply and enforce the PRM as a percent (even if some reasons think of it as MW), it seems clearest to stick with the normalized approach here. So I think we could drop `GSw_PRM_StressThresholdOutageMagnitude` and only keeping the normalized version.
  1. Just to keep the switch name shorter, you could drop 'Normalized' from the switch name. So just keep a single switch, `GSw_PRM_StressThresholdMagnitude`, with the functionality of the current `GSw_PRM_StressThresholdNormalizedOutageMagnitude` switch.
1. It's unclear what units are being used for `GSw_PRM_StressThresholdNormalizedOutageMagnitude` if the value is 10 (the description says it's a ratio). From your review, it looks like thresholds range from 3% in ISONE and NWPCC to 20% in Tri-state and 25% in ERCOT. It will be good to do a sweep of the values, but for now, I would use something between 3% and 25%.
  1. But I would use fractional units in the switch (in keeping with the related switches), so `transgrp_0.03_max` if you go with the ISONE/NWPCC number.

reeds/resource_adequacy/stress_periods.py
```diff
@@ -643,17 +755,91 @@ def update_prm(sw, t, iteration, failed, combined_periods_write):
...
+            print(f"Calculating and writing annual {stress_metric} for iteration {iteration}")
+            dfmetric = get_annual_stress_metric(sw.casedir, t, stress_metric, iteration=iteration)
+
+            # Metric thresholds are defined on a per-year basis, but PRAS reports the total over all resource adequacy years,
+            # For all metrics except NEUE, divide by the number of resource adequacy years to get the average per year
+            if stress_metric != 'NEUE':
+                dfmetric /=  len(sw['resource_adequacy_years'])
+
+            dfmetric.round(2).to_csv(
+                os.path.join(sw.casedir, 'outputs', f"{stress_metric.lower()}_{t}i{iteration}.csv")
```
patrickbrown4: It's kind of a lot of csv's if we're saving each metric, year, and iteration to its own file; on the order of 100 files for a full-US run.

Could we instead write a single file for each year/iteration with all the metrics?
- You could change the existing `metric` column to `aggmethod` since it only records `sum` or `max`
- Then you could either add a `metric` column for NEUE/LOLH/etc., or put the metrics in wide format (so the columns would be `level`, `aggmethod`, `region`, `NEUE`, `LOLH`, etc.). Long is clearer organizationally but wide is ok if the files get too large.

reeds/resource_adequacy/stress_periods.py
```diff
@@ -643,17 +755,91 @@ def update_prm(sw, t, iteration, failed, combined_periods_write):
...
+        dfmetric (pd.DataFrame): Hourly EUE values (MW) indexed by timestamp, one
+            column per region. Typically obtained from :func:`get_pras_stress_metric`.
+        eue_threshold (float, optional): Minimum EUE (MW) to qualify as an outage
+            hour. Hours at or below this value are excluded. Defaults to 0.
+
+    Returns:
+        pd.Series: Peak hourly EUE (MW) over the full timeseries, indexed by region.
+            Regions with no outage hours return 0.
+    """
+    return dfmetric.max()
```
patrickbrown4: I'm in favor of docs but this seems like a lot for a single `.max()` call. I think it'd be clearer and more readable to just use `.max()` when you need it. (Maybe same for the above.)

reeds/resource_adequacy/stress_periods.py
```diff
@@ -643,17 +755,91 @@ def update_prm(sw, t, iteration, failed, combined_periods_write):
...
-        )
+        ## TODO: check if need to refactor or remove
+
+        for stress_metric in ['EUE', 'NEUE', 'LOLH', 'LOLE', 'OutageDuration', 'OutageMagnitude', 'NormalizedOutageMagnitude']:
+            print(f"Calculating and writing annual {stress_metric} for iteration {iteration}")
+            dfmetric = get_annual_stress_metric(sw.casedir, t, stress_metric, iteration=iteration)
+
+            # Metric thresholds are defined on a per-year basis, but PRAS reports the total over all resource adequacy years,
+            # For all metrics except NEUE, divide by the number of resource adequacy years to get the average per year
+            if stress_metric != 'NEUE':
```
patrickbrown4: Outage magnitude is an hourly metric, and duration is a per-event metric, so neither of them should be divided by the number of RA years.

I think only EUE (which was dropped) and the LOLx metrics would be divided by the number of weather years.

reeds/resource_adequacy/stress_periods.py
```diff
@@ -201,42 +184,88 @@ def get_and_write_neue(sw, write=True):
...
+        if stress_metric == 'LOLE':
+            # Count a day as an event-day if at least one hour has LOLE > 0
+            dfmetric_agg = dfmetric.rename(columns=rmap).groupby(axis=1, level=0).sum()
+            daily_max = dfmetric_agg.groupby(
+                [dfmetric_agg.index.year, dfmetric_agg.index.month, dfmetric_agg.index.day]
+            ).max()
+            event_days = (daily_max > 0).sum()
+            _metric[hierarchy_level, 'sum'] = event_days
+            _metric[hierarchy_level, 'max'] = event_days
+            continue
```
patrickbrown4: The `continue`s here are kind of hard to follow. It looks like they're being used to control which metrics get summed, but those metrics are implicit rather than explicit. I think it'd be clearer to drop the `continue`s and name the metrics that need to be summed.

That being said, since you will always run this function for each of the metrics (since we always save all of the values), I think it might be clearer to reorganize the function:

1. Drop the `stress_metric` kwarg. Instead of the `if stress_metric == ` blocks, just explicitly calculate each metric.
    1. So above, you'd run `get_pras_stress_metric()` twice, once for EUE and once for LOLE (since those are the only two things you can get from it), and use the appropriate one for the calculations below.
    1. Add the metric name as a third key in the `_metric` dataframe
    1. Calculate each metric explicitly
5. Then at the end you'd have a single dataframe you could write, which addresses the suggestion below to put them all in the same file.

patrickbrown4: Here's my take on the structure: 475a94af0c565dc4c0c44ab59b7cee43ea061893. It's not complete because it doesn't yet adapt the rest of the processing for the single `ra_metrics_{year}i{iteration}.csv` file, but it makes the individual aggregate RA metric calculations more explicit (at least in my mind).

I might try a few more tweaks on the [pb/multimetric](https://github.com/ReEDS-Model/ReEDS/compare/pb/multimetric) branch to get some more test cases running, so let's check in about integration when you get back.

reeds/resource_adequacy/stress_periods.py
```diff
@@ -692,6 +879,7 @@ def main(sw, t, iteration=0, logging=True):
     if (
         (int(sw.GSw_PRM_UpdateMethod) == 0)
         or (len(new_stressperiods_write) and (int(sw.GSw_PRM_UpdateMethod) == 3))
+        or not any(failed[c].name == 'NEUE' for c in failed)
```
patrickbrown4: We still want to allow `GSw_PRM_UpdateMethod = 1` (fixed PRM increment controlled by `GSw_PRM_UpdateFraction`) to work regardless of the RA metric chosen, so I think it'd be best to:
1. Drop the conditional here (done on pb/multimetric)
2. Add a check to `runreeds.check_compatibility()` that raises an error if `int(sw.GSw_PRM_UpdateMethod) in [2, 3]` and `'neue' not in sw.GSw_PRM_StressThresholdMetrics.lower().split('/')`

patrickbrown4 (approved):
Looks great, thanks a lot!

Good to merge as long as the USA_defaults case looks the same.

PR #128, opened 2026-06-23
FINITO Link

mmowers:
## Summary
This PR provides a linkage to the FINITO model, which is currently a private repository at https://github.com/NatLabRockies/FINITO). The linkage allows the full electricity sector detail of the ReEDS model to be used in combination with the industrial (and other) sector representation of the FINITO model.

The linkage is turned off by default, so this PR itself should not have impact on ReEDS results (although very small differences were observed in out years, presumably because the ReEDS model itself was touched, even if the matrix is identical).

For a high-level overview of FINITO and the linkage points see [this slide deck](https://nrel.sharepoint.com/:p:/s/DOE-OP-NG-PowerModeling/IQBIzWwGpRmXTpTCXEnU9-EDAckGoZz0iI4plOIJpTit8Fw?e=h8d5Fg). 

See companion FINITO PR: https://github.com/NatLabRockies/FINITO/pull/438

## Technical details 
Turning on the linkage adds the constraints of the FINITO model to those of the ReEDS model, and adds the two objective functions together.

To turn on the linkage, clone the FINITO model and checkout the `bs/reeds_link2` branch, unless this branch has already been merged into `main`, at which point  `main`.

Configure these switches:
- `GSw_FINITO_Link`: This switch turns on the linkage
- `finito_dir`: This switch needs to point to the FINITO directory as a complete, absolute path
- `finito_cases_file`: This switch should be the suffix of the cases file in the FINITO directory, which contains the corresponding FINITO scenario configuration
- `finito_case`: This switch is the specific case in `finito_cases_file` to be run
- `first_year_finito`: This switch is the first year that the linkage will be implemented

The `reeds` conda environment should be used for the linked model.

Linked runs result in the the same set of ReEDS outputs, but new outputs are also added from FINITO in the `finito` subdirectory of the run directory.

FINITO documentation is at https://github.com/NatLabRockies/FINITO

A comparison of default (Ref) was made berfore and after this change, showing no changes, as expected: 
[reeds_report_ref_2026-08-05.zip](https://github.com/user-attachments/files/30906898/reeds_report_ref_2026-08-05.zip)

Running the USA_fast case with the linkage turned on results in moderate generation differences to the ReEDS standalone "USA_fast", but similar amonts of fossil and renewables by 2050:
image

Electricity load differences show that FINITO endogenous industrial electricity demand is lower than the reference demand that was removed from the EER load profiles, even in the near term:
image
 
Gas price differences align with gas generation differences by 2050, although alignment isn't perfect for the years before:
image

See more results of the comparison here:
[reeds_report_finito_vs_usa-fast_plus4_2026-08-08.zip](https://github.com/user-attachments/files/30907403/reeds_report_finito_vs_usa-fast_plus4_2026-08-08.zip)

Also, a more thorough comparison of the reference industrial electricity load that was removed from ReEDS load profiles, as compared to the endogenous finito load that was added back in, is here: 
[finito_load_check_2026-08-06.zip](https://github.com/user-attachments/files/30952817/finito_load_check_2026-08-06.zip)

### Details to double-check

  - [ ] Units are specified


reeds/core/terminus/report.gms
```diff
@@ -1768,6 +1825,11 @@ error_check('z') = (
         )
 * account for penalty paid to deploy capacity beyond interconnection queue limits        
         + sum{(tg,r), cap_penalty(tg) * CAP_ABOVE_LIM.l(tg,r,t) }  
+* account for costs from FINITO: deflate from $2018 to $2004,
+* remove any FINITO scaling, and then apply ReEDS scaling
+$ifthene.linked_objective Sw_FINITO_Link==1
+        + cost_scale * ( Z_finito.l(t)$t_finito(t) * deflator('2018') / obj_scale )
+$endif.linked_objective
```
bsergi, line 1832: The absolute value of the error check gets larger even after accounting for the FINITO objective function. It's still small but I'm not sure why this is happening yet, so would be interested in any suggestions folks have on this.

Image

reeds/core/terminus/report_params.csv
```diff
@@ -107,6 +107,7 @@ planned_outage(i),fraction,average scheduled outage rate (h-weighted) by technol
 "h2_usage(r,allh,t)",metric tons/hour,total H2 usage by hour (H2-CT/CC and non-electric H2 consumption),,,
 "load_cat(loadtype,r,t)",MWh,Annual exogenous load by category,,,
 "load_rt(r,t)",MWh,Annual exogenous load,,,
+"loadvar_rt(r,t)",MWh,Annual LOAD during rep periods,,,
```
wesleyjcole, line 110: This says "during rep periods," but you sum over all h, so it includes stress periods as well.

reeds/core/terminus/report_params.csv
```diff
@@ -145,6 +146,7 @@ raw_op_cost(t),2004$,sum of operational costs from systemcost,,,
 RE_gen_price_nat(t),2004$/MWh,marginal cost of the national RE generation constraint,,,
 "repbioprice(r,t)",2004$/MMBtu,highest marginal bioprice of utilized bins for each region,,,
 "repgasprice(cendiv,t)",2004$/MMBtu,highest marginal gas price of utilized gas bins for each census division,,,
+"repgasprice_finito(cendiv,allh,t)",2004$/MMBtu,highest marginal gas price of utilized gas bins by hour for each census division when using linked model,,,
```
wesleyjcole, line 149: It's not clear from the comment whether this is the price from ReEDS sent to FINITO or from FINITO sent to ReEDS.

reeds/core/terminus/report.gms
```diff
@@ -550,20 +563,52 @@ repgasquant_nat(t)$tmodel_new(t) = sum{cendiv, repgasquant(cendiv,t) } ;
...
+;
+
+* approach with detailed fuels representation (GSw_DetailedFuels=1)
+repgasprice_finito(cendiv,h,t)$[tmodel_new(t)$(not tfuel(t))$Sw_DetailedFuels] =
+    deflator('2018') * 1/(obj_scale) * 1/(pvf_onm(t)) 
+*   citygate price of natural gas
+    * [ smax{(cfp,st)$st_cendiv(st,cendiv), eq_supplydemand_cf.M(cfp,'NG',st,h,t) } / hours(h) 
+*   electric-sector markup for natural gas
+* TODO: activate after FINITO pricing PR is merged
+*       + smax{cfp$[gasp(cfp)$map_cf_fe(cfp,'NG')$valcft(cfp,t)], cf_markup(cfp,'NG','Electric_Power',cendiv,t) }
```
wesleyjcole, line 593: Can this be left out of the code and added as an issue instead?

reeds/core/terminus/report.gms
wesleyjcole: I think it might have to do with your fuel cost reporting.  You have `$tfuel(t)` on biopower and gas price reporting, but not for nuclear and coal.

helpers/restart_runs.py
```diff
@@ -95,6 +96,14 @@
...
     for f in more_copyfiles:
+        if f.lower().startswith('finito'):
+            # if file starts with 'finito' append the finito directory
+            sw = reeds.io.get_switches(case)
+            if int(sw.GSw_FINITO_Link):
+                f_copy = re.sub("^finito/", "" , f, flags=re.IGNORECASE)
+                shutil.copy(os.path.join(sw.finito_dir,f_copy), os.path.join(case,f))
+        else:
+            shutil.copy(os.path.join(reeds_path,f), os.path.join(case,f))
         shutil.copy(f, Path(case, f))
```
wesleyjcole, line 107: Should this be here twice in a row like this?  Is line 107 needed?

postprocessing/bokehpivot/reeds2.py
```diff
@@ -3079,7 +3079,7 @@ def pre_spur(dfs, **kw):
         }
     ),
     ('Load (TWh)',
-        {'file':'load_rt',
+        {'file':'loadvar_rt',
```
wesleyjcole, line 3082: Won't this change the behavior of default runs?  `load_rt` is exogenous load only, while `loadvar_rt` is based on `LOAD.l` which includes things like Canadian exports and H2 production.

reeds/core/terminus/report.gms
```diff
@@ -514,21 +516,32 @@ ptc_out(i,v,t)$[tmodel_new(t)$ptc_value_scaled(i,v,t)] = ptc_value_scaled(i,v,t)
...
-repgasquant_gb(cendiv,gb,t)$[(Sw_GasCurve = 0 or Sw_GasCurve = 3)$tmodel_new(t)] =
+repgasquant_gb(cendiv,gb,t)$[(Sw_GasCurve = 0 or Sw_GasCurve = 3)$tmodel_new(t)$tfuel(t)] =
     sum{h, GASUSED.l(cendiv,gb,h,t) * hours(h) } * gas_scale/ 1e9 ;
 
-repgasquant(cendiv,t)$[(Sw_GasCurve = 0 or Sw_GasCurve = 3)$tmodel_new(t)] =
+repgasquant(cendiv,t)$[(Sw_GasCurve = 0 or Sw_GasCurve = 3)$tmodel_new(t)$tfuel(t)] =
     sum{gb, repgasquant_gb(cendiv,gb,t) };
 
-repgasquant(cendiv,t)$[(Sw_GasCurve = 1 or Sw_GasCurve = 2)$tmodel_new(t)] =
+repgasquant(cendiv,t)$[(Sw_GasCurve = 1 or Sw_GasCurve = 2 or Sw_FINITO_Link = 1)$tmodel_new(t)] =
```
wesleyjcole, line 544: Should this have a `$tfuel(t)` or a `$(not tfuel(t))` conditional?

reeds/core/setup/c_model.gms
```diff
@@ -3572,6 +3580,12 @@ eq_h2_demand(p,t)$[(sameas(p,"H2"))$tmodel(t)$(yeart(t)>=h2_demand_start)$(Sw_H2
     + sum{(i,v,r,h)$[valgen(i,v,r,t)$h2_combustion(i)$h_rep(h)],
             GEN(i,v,r,h,t) * hours(h) * h2_combustion_intensity * heat_rate(i,v,r,t)
     }
+
+* hydrogen demand from industry when linked with FINITO: demand [MMBtu/yr] * conversion [metric tons-H2/MMBtu-H2]
+* TODO: should we disable exogenous H2 demand when linked?
```
wesleyjcole, line 3558: Would you remove the TODO statements throughout?

reeds/core/solve/3_solve_oneyear.gms
```diff
@@ -23,6 +23,13 @@ tload(t) = no ;
 tmodel(t) = no ;
 tmodel("%cur_year%") = yes ;
 
+* (ReEDS-FINITO) also reset t_finito
+$ifthene.linked_finito_time %GSw_FINITO_Link% == 1
+t_finito(t) = no ; 
+t_finito(t)$[tmodel(t)$(t.val>=first_year_finito)$(t.val<=endyear)] = yes ;
```
wesleyjcole: In `b_inputs.gms` you used "%first_year_finito%".  Are these interchangable, or do they refer to different things?

reeds/core/setup/c_model.gms
```diff
@@ -3603,6 +3617,11 @@ eq_h2_demand_regional(r,h,t)
     + sum{(i,v)$[valgen(i,v,r,t)$h2_combustion(i)],
             GEN(i,v,r,h,t) * h2_combustion_intensity * heat_rate(i,v,r,t)
        }
+
+* when linked includeregional hydrogen demand for industry from FINITO
```
wesleyjcole, line 3594:
```suggestion
* when linked include regional hydrogen demand for industry from FINITO
```

reeds/core/setup/c_model.gms
```diff
@@ -3845,6 +3864,11 @@ eq_co2_capture(r,h,t)
 
 * capture from DAC
     + sum{(i,v)$[dac(i)$valcap(i,v,r,t)$i_p(i,"DAC")], PRODUCE("DAC",i,v,r,h,t) }$Sw_DAC
+* capture from industry when linked with FINITO [metric_tons-CO2/hr]: 
+* calculation: hours_per_year [yrs/hr] * capture [scaled_metric_tons-CO2/yr] / co2_scale [scaled_metric_tons-CO2/metric_tons-CO2]
+$ifthene.linked_co2_capture Sw_FINITO_Link==1
+    + [CAPTURE_CO2EM(r,h,t) / co2_scale]$[t_finito(t)$h_rep(h)]
```
wesleyjcole, line 3843: `$h_rep(h)` is already on the equation conditional, so shouldn't be needed here.

reeds/core/setup/c_model.gms
```diff
@@ -3874,6 +3898,11 @@ eq_co2_spurline_caplimit(r,cs,h,t)$[Sw_CO2_Detail$r_cs(r,cs)$tmodel(t)$(yeart(t)
     =g=
 
     CO2_STORED(r,cs,h,t)
+* (ReEDS-FINITO) extraction of CO2 [metric_tons-CO2/hr] 
+* calculation: hours_per_year [yrs/hr] * (1 / co2_scale [scaled_metric_tons-CO2/metric_tons-CO2]) * use [scaled_metric_tons-CO2/yr] 
+$ifthene.linked_co2_spurline_caplimit Sw_FINITO_Link==1
+    + [ (1 / co2_scale) * EXTRACT_CO2_CS(cs,r,h,t) ]$[h_rep(h)$t_finito(t)]
```
wesleyjcole, line 3877: This might be a broader issue. It's not clear to my why `$h_rep(h)` would be included here--but it's also not clear why `$h_rep(h)` is included for `eq_co2_capture` but not in the other equations.  It might be better to just move the `$h_rep(h)` to the equation definitions rather than filter them out just for the FINITO linkage.

reeds/core/setup/c_model.gms
```diff
@@ -3906,6 +3947,14 @@ eq_co2_injection_limit(cs,h,t)$[Sw_CO2_Detail$tmodel(t)$(yeart(t)>=co2_detail_st
 
 * must exceed metric tons per hour entering storage
     sum{r$r_cs(r,cs), CO2_STORED(r,cs,h,t) }
+
+
+
```
wesleyjcole, line 3925: Some of this whitespace can go.

reeds/core/setup/c_model.gms
```diff
@@ -3922,6 +3971,14 @@ eq_co2_cumul_limit(cs,t)$[tmodel(t)$Sw_CO2_Detail$(yeart(t)>=co2_detail_startyr)
         $[(yeart(tt)<=yeart(t))$(tmodel(tt) or tfix(tt))$(yeart(tt)>=co2_detail_startyr)
         $r_cs(r,cs)$h_rep(h)],
         yearweight(tt) * hours(h) * CO2_STORED(r,cs,h,tt) }
+
+* (ReEDS-FINITO) cumulative amount extracted over time
+$ifthene.linked_co2_storage_cumul_limit Sw_FINITO_Link==1
+    - sum{(r,h,tt)
+        $[(yeart(tt)<=yeart(t))$(tmodel(tt) or tfix(tt))$(yeart(tt)>=co2_detail_startyr)
+        $r_cs(r,cs)$(tfuel(tt))],
+        yearweight(tt) * hours(h) * EXTRACT_CO2_CS(cs,r,h,tt) }$[t_finito(t)]
```
wesleyjcole, line 3953: Just flagging that this one does not have a `$h_rep(h)`.

reeds/input_processing/copy_files.py
```diff
@@ -1297,6 +1297,11 @@ def main(reeds_path, inputs_case):
     
     source_deflator_map = get_source_deflator_map(reeds_path)
 
+    ### Rewrite the  finito scalar as GAMS-readable definitions
+    if int(sw['GSw_FINITO_Link']) == 1:
+        pd.read_csv(os.path.join(os.path.dirname(inputs_case),'finito','inputs','scalars.csv')).iloc[:, :-1].to_csv(os.path.join(os.path.dirname(inputs_case),'finito','inputs','scalars.csv'),index=False)
```
wesleyjcole: Will this cause problems if copy_files gets stuck and then you restart it?  Won't it remove a row each time it is run?

Also, it would be great to split this line so it isn't so long.

reeds/input_processing/hourly_load.py
```diff
@@ -596,6 +596,85 @@ def reaggregate_to_model_regions(
 
     return regional_load_hourly
 
+#TODO: add docstrings
```
wesleyjcole: Another TODO.

reeds/resource_adequacy/prep_data.py
```diff
@@ -147,8 +147,38 @@ def main(t, casedir, iteration=0):
...
+
+    # when running linked model with FINITO (GSw_FINITO_Link=1), 
+    # we also need to add in the load from FINITO
+    if int(sw.GSw_FINITO_Link):
+        load_finito_rt = gdxreeds['load_finito_rt'].rename(columns={'allh':'h', 'Value':'load_MW'})
+
+        # down select to relevant model year
+        load_finito = load_finito_rt.loc[load_finito_rt.t==t].drop('t', axis=1).copy()
+        #TODO: why so much shifting? validate the dynamics here with FINITO team
+        #TODO: also check against reference quantity
```
wesleyjcole, line 163: More TODOs to remove.

cases.csv
```diff
@@ -32,6 +32,11 @@ coalscen,"Coal price scenario (inputs\fuelprices\{coalscen}.csv). The options fo
 demandscen,Demand scenario (see inputs\load),N/A,AEO_2026_baseline,
 dr_shedscen,DR shed scenario,demo_data_January_2025,demo_data_January_2025,
 evmcscen,EVMC scenario for load and supply curve and flexibility,Baseline,Baseline,
+GSw_FINITO_Link,Switch to enable linkage with FINITO,0; 1,0,
```
wesleyjcole: I think we should stick with convention and keep all the GSw_* switches together, so I think this is worth moving down with the rest of them.

cases_test.csv
```diff
@@ -1,66 +1,69 @@
...
-GSw_OffshoreZones,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,1,
-GSw_OffshoreBackbone,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,1,
-GSw_OffshoreBackflow,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,1,
-pras_agg_ogs_lfillgas,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,1
-pras_existing_unit_size,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,0
-pras_scheduled_outage,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,0
-pras_unitsize_source,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,r2x
-pras_vre_combine,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,1
-pras_samples,,,,,,10,10,10,,,,,,,,,,,,,,,,10,10,,,,,,
+,Default Value,Pacific,USA_defaults,Mid_Case,USA_decarb,github_Pacific,github_Everything,github_MA_county_CC,Pacific_CC,Pacific_weks,Pacific_full_year,Interday_storage,Pacific_2020,Pacific_rep15,WY_county,WECC_county,PJM_county_CC,NYVT_mixed,OR_water,MonteCarlo_Random,MonteCarlo_LHS,Everything,Simple,USA_fast,USA_faster,MultiMetricRA,Pacific_DR,Pacific_MGA,Pacific_LoadSite95,MARICTNYNJPAOH_Offshore,R2P,FINITO,FINITO_Pacific
```
wesleyjcole: Why add two FINITO cases here instead of just one?

reeds/input_processing/hourly_load.py
```diff
@@ -596,6 +596,85 @@ def reaggregate_to_model_regions(
...
+        raise ValueError(
+            'Cannot aggregate load_finito.csv from z134 to model regions because '
+            'these p regions span multiple model regions: '
+            f'{zones_per_p.loc[zones_per_p > 1].index.tolist()}'
+        )
+    p2zone = p2zone.set_index('p')['zone']
+    load_finito = load_finito.rename(columns=p2zone).T.groupby(level=0).sum().T
+
+    # allocate annual load to hours, assuming flat demand
+    # TODO: should this use h_weight_finito?
```
wesleyjcole: Another TODO.

reeds/input_processing/hourly_repperiods.py
wesleyjcole: If I'm reading things right, `get_hourly_finito_load` returns end-use load, but this comment says `load.h5` is bubar load, so I think you are adding end-use load to bus-bar load.

It also would be good to add docstrings to the functions to clarify which this kind of thing.

wesleyjcole (commented):
This was interesting to see. Nice work on helping to get this where it is.

You do use "linked model" a lot throughout.  It might be nice to be more specific with the language there because we have had other linked models, so those comments can get confusing.

reeds/inputs.py
```diff
@@ -907,3 +909,56 @@ def validate_zoneset(GSw_ZoneSet):
...
+    shutil.copy2(os.path.join(caseSwitches['finito_dir'], 'cases.csv'), os.path.join(casedir, 'finito'))
+    shutil.copy2(os.path.join(caseSwitches['finito_dir'], f"cases_{caseSwitches['finito_cases_file']}.csv"), os.path.join(casedir, 'finito'))
+
+    #%% (GSw_Trade_PriceResponse > 0) If doing a price-responsive trade run, retrieve the reference exports/imports prices
+    if int(caseSwitches['GSw_Trade_PriceResponse']) > 0:
+        initialize_price_response_path = os.path.join(casedir_finito, 'input_processing', 'processing', 'initialize_price_response.py')
+        # Collect all arguments for initialize_price_response.py
+        initialize_price_response_args = f" -c {casedir} -b {BatchName} -cr {caseSwitches['GSw_Trade_PriceResponse_RefScen']} -l {caseSwitches['GSw_FINITO_Link']}"
+        # Call initialize_price_response.py file before starting the runs
+        os.system('python ' + initialize_price_response_path + initialize_price_response_args)
```
patrickbrown4: Can you use `subprocess.run()` instead? It's [recommended](https://docs.python.org/3/library/os.html#:~:text=using%20that%20module%20is%20recommended%20to%20using%20this%20function) over `os.system()` for new code, and is what we use for running PRAS etc.

reeds/inputs.py
```diff
@@ -907,3 +909,56 @@ def validate_zoneset(GSw_ZoneSet):
...
+def setup_finito(casedir, caseSwitches, BatchName):
+    #%% Copy FINITO code folders and inputs into [casedir]/finito
+    # ... finito directory within the case directory
+    casedir_finito = os.path.join(casedir,'finito')
+    # ... define the inputs case directory for FINITO
+    inputs_case_finito = os.path.join(casedir,'finito','inputs_case')
+
+    # copy directories
+    os.makedirs(casedir_finito, exist_ok=True)
+    shutil.copytree(os.path.join(caseSwitches['finito_dir'], 'inputs'),os.path.join(casedir,'finito', 'inputs'))
```
patrickbrown4: (nitpick) Sorry to quibble but in my opinion, `Path` makes these a bit more readable since it's shorter
```suggestion
    shutil.copytree(Path(caseSwitches['finito_dir'], 'inputs'), Path(casedir, 'finito', 'inputs'))
```

reeds/input_processing/hourly_repperiods.py
```diff
@@ -50,11 +52,18 @@
...
 #%%### Functions
-def get_load(inputs_case, keep_modelyear=None, keep_weatheryears=[2012]):
+def get_load(inputs_case, sw, keep_modelyear=None, keep_weatheryears=[2012]):
     """
     """
     ### Subset to modeled regions
     load = reeds.io.read_file(os.path.join(inputs_case,'load.h5'))
+
+    ### When running the linked model (GSw_FINITO_Link=1) we add reference load estimates
+    ### for FINITO load back in to use when selecting representative periods
```
patrickbrown4: (nitpick) `sw` is so fast to load that I leave it out of required arguments if `case`/`inputs_case` is already included; that simplifies testing and ensures the inputs are in sync. (Would need to adjust the call in `reedsplots.plot_seed_stressperiods()` too.)

```suggestion
def get_load(inputs_case, keep_modelyear=None, keep_weatheryears=[2012]):
    """
    """
    ### Subset to modeled regions
    load = reeds.io.read_file(os.path.join(inputs_case,'load.h5'))

    ### When running the linked model (GSw_FINITO_Link=1) we add reference load estimates
    ### for FINITO load back in to use when selecting representative periods
    sw = reeds.io.get_switches(inputs_case)
```

reeds/input_processing/copy_files.py
patrickbrown4: Agreed, it seems risky to me to rely on position. I would explicitly specify the entries to be dropped (ignoring errors if they aren't there).

reeds/core/terminus/report_params.csv
patrickbrown4: We restrict `h` to rep periods at the top of `report.gms` though:

https://github.com/ReEDS-Model/ReEDS/blob/bc804315b2296f8c95c7c8eeb4554c53c380fdea/reeds/core/terminus/report.gms#L121-L122

But if you want to be extra careful (in case we decide to change that down the road and miss it here), you could add a `$h_rep(h)` to the sum.

runreeds.py
```diff
@@ -761,6 +768,74 @@ def setup_window(
...
+    if duplicate_columns:
+        raise ValueError(f"The FINITO cases_{df_cases[case]['finito_cases_file']}.csv has the following duplicate column names: {duplicate_columns}")
+    ### identify the FINITO case 
+    if df_cases[case]['finito_case'] == 'same':
+        finito_case=case
+    else:
+        finito_case=df_cases[case]['finito_case']
+    # ensures **exact** match of names between the ReEDS cases_{}.csv and the FINITO cases_linked.csv
+    if finito_case not in (df_cases_suf_finito.columns):
+        raise ValueError(f"The 'finito_case' input '{finito_case}' in the ReEDS cases file does not exist in FINITO's cases_{df_cases[case]['finito_cases_file']}.csv.")
```
patrickbrown4: (nitpick) Would you mind wrapping new long lines to ~100 characters, here and elsewhere?
```suggestion
        raise ValueError(
            f"The 'finito_case' input '{finito_case}' in the ReEDS cases file does not "
            "exist in FINITO's cases_{df_cases[case]['finito_cases_file']}.csv."
        )
```

cases.csv
```diff
@@ -32,6 +32,11 @@ coalscen,"Coal price scenario (inputs\fuelprices\{coalscen}.csv). The options fo
...
+finito_dir,Directory with cloned FINITO repository,N/A,/path/to/FINITO/directory,
```
patrickbrown4: Have you considered other approaches for this switch? It's the only switch that will never work with its default value, which is kind of unsatisfying.

Some options:
1. On the TSC repo we have a [candidate_reeds_paths](https://github.nrel.gov/ReEDS/TSC/blob/76cb93671e7ad48c293049eac02d51c64468455a/tsc/defaults.yaml#L5) switch that contains a list of possible paths to a ReEDS folder, which then gets resolved to the first match [here](https://github.nrel.gov/ReEDS/TSC/blob/76cb93671e7ad48c293049eac02d51c64468455a/tsc/__init__.py#L67). It's of course not comprehensive, but works for a small team.
2. You could make the default `{reeds_path}/../FINITO` (i.e. putting FINITO in the same folder as ReEDS, which seems better than nothing) and then resolve it to an absolute path in `inputs.add_intermediate_switches()` via `Path(finito_dir.format(reeds_path=reeds.io.reeds_path)).resolve()`
3. You could add a new standalone `user_settings.yaml` file where we can start putting this kind of input. (I've been considering the same thing for `inputs/remote`, to provide the option for multiple ReEDS repos to share the same Zenodo folder and cut down on storage size)

cases.csv
```diff
@@ -32,6 +32,11 @@ coalscen,"Coal price scenario (inputs\fuelprices\{coalscen}.csv). The options fo
...
+finito_cases_file,"extension of the FINITO cases file including all scenarios to be run, e.g., 'linked' for cases_linked.csv",N/A,linked,
+finito_case,Case name for FINITO in {finito_dir}/cases_linked.csv; use 'same' if the scenario name in cases_linked.csv matches the scenario name in the cases file,N/A,BAUfast,
```
patrickbrown4: It seems like "same" would be a better default value

cases.csv
```diff
@@ -32,6 +32,11 @@ coalscen,"Coal price scenario (inputs\fuelprices\{coalscen}.csv). The options fo
...
+first_year_finito,First year when the FINITO module is enabled,N/A,2018,
```
patrickbrown4: Just to check, would it simplify things to use `GSw_StartMarkets` here? In other parts of the model we use that to indicate the start of endogenous optimization behavior (deviating from historical), so it might simplify to use it here too instead of adding another switch. But if it helps to keep them independent, that's ok.

cases_test.csv
```diff
@@ -1,66 +1,69 @@
...
+GSw_OffshoreBackbone,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,1,,,
+GSw_OffshoreBackflow,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,1,,,
+pras_agg_ogs_lfillgas,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,1,,
+pras_existing_unit_size,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,0,,
+pras_scheduled_outage,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,0,,
+pras_unitsize_source,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,r2x,,
+pras_vre_combine,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,1,,
+pras_samples,,,,,,10,10,10,,,,,,,,,,,,,,,,10,10,,,,,,,10,10
+GSw_FINITO_Link,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,1,1
+finito_dir,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,/path/to/FINITO/directory,/path/to/FINITO/directory
```
patrickbrown4: See note in `cases.csv` - I worry that having switch values that need to be modified by every user in a fairly high-traffic (and difficult-to-resolve-conflicts-for csv file) will exacerbate merge conflicts

cases_test.csv
patrickbrown4: Agreed, I would keep it to a single small test case.

reeds/inputs.py
```diff
@@ -907,3 +909,56 @@ def validate_zoneset(GSw_ZoneSet):
...
+        'python ' + os.path.join(caseSwitches['finito_dir'], 'input_processing', 'processing', 'copy_files.py') + 
+        f" -c {casedir_finito} -d {inputs_case_finito} --link"
+        )
+
+    # Print an error message if copy_files.py encounters any issue 
+    if copy_files_run != 0:
+        print("\nERROR: FINITO copy_files.py encountered an issue and did not complete successfully.")
+        print("Please check the console output above for details.")
+        print("The issue could be due to regionality, focus sector filtering, or file reading errors.") 
+        os._exit(1)
```
patrickbrown4: It looks like `os._exit()` isn't standard/recommended. How about putting the message in `err` and then `raise RuntimeError(err)`?

```suggestion
        err = (
            "FINITO copy_files.py encountered an issue and did not complete successfully.\n"
            "Please check the console output above for details.\n"
            "The issue could be due to regionality, focus sector filtering, or file reading errors."
        )
        raise RuntimeError(err)
```

runreeds.py
patrickbrown4, line 1: The PR mentions that FINITO doesn't work for `GSw_ZoneSet = z90` (and presumably the other more recently added resolutions). Can you add that and any other mutually incompatible switch options to `check_compatibility()`?

Some to check:
- [ ] `GSw_ZoneSet`
- [ ] `GSw_PRM_CapCredit`
- [ ] `GSw_GasPriceAdjMethod`
- [ ] `GSw_Region` (might be fine)

reeds/resource_adequacy/prep_data.py
```diff
@@ -147,8 +147,38 @@ def main(t, casedir, iteration=0):
...
+
+        # map from rep day to actual hour
+        # since we don't have multi-year profiles for FINITO load 
+        # we assume they repeat across all weather years
```
patrickbrown4, line 167: Can you add this to a user-facing list of caveats somewhere (`user_guide.md` or a new page on the FINITO linkage)? If the FINITO load is fairly flexible then this assumption might be pretty far off, and it will get the weekday/weekend lineup wrong.

reeds/resource_adequacy/prep_data.py
```diff
@@ -147,8 +147,38 @@ def main(t, casedir, iteration=0):
...
+        load_finito = pd.merge(load_finito, h_dt_szn.reset_index(), on='h', how='outer')[['timestamp', 'r', 'load_MW']]
+        load_finito = load_finito.rename(columns={'timestamp':'datetime'})
+        # add model year back in 
+        #load_finito = load_finito.assign(year=t)
```
patrickbrown4, line 171:
```suggestion
```

reeds/resource_adequacy/prep_data.py
```diff
@@ -147,8 +147,38 @@ def main(t, casedir, iteration=0):
...
+        load_finito = load_finito.pivot(index=['datetime'], columns='r', values='load_MW')
```
patrickbrown4, line 172: Can you do this before the `h_dt_szn` merge? Reshaping big hourly dataframes can be slow so I'd try to avoid it if possible.

reeds/core/setup/b_inputs.gms
```diff
@@ -943,7 +944,7 @@ tfix(t) = no ;
 stfeas(st) = no ;
 tprev(t,tt) = no ;
 tsolved(t) = no ;
-
+tfuel(t)=no;
```
patrickbrown4: (nitpick, sorry) Can you put spaces around operators for readability?

reeds/core/setup/b_inputs.gms
```diff
@@ -930,7 +930,8 @@ set tmodel(t) "years to include in the model",
     tfix(t) "years to fix variables over when summing over previous years",
     tprev(t,tt) "previous modeled tt from year t",
     stfeas(st) "states to include in the model",
-    tsolved(t) "years that have solved" ;
+    tsolved(t) "years that have solved",
+    tfuel(t) "years that use ReEDS fuel supply curve module (otherwise uses supply curves in FINITO)" ;
```
patrickbrown4, line 950: Still going through the code so I might be off base, but the fuel supply curve module isn't always used (it's turned off if `GSw_GasCurve = 2`), so suggest clarifying.

```suggestion
    tfuel(t) "years that use ReEDS fuel costs (years not included here use FINITO costs)" ;
```

Since FINITO isn't the default, it might be clearer to do the inverse (i.e. add a `tfinito(t)`). But if that's messier in downstream code then it's fine to keep this approach.

runreeds.py
```diff
@@ -761,6 +768,74 @@ def setup_window(
             OPATH.writelines( "python valuestreams.py" + '\n')
 
 
+# (ReEDS-FINITO) update 'df_cases' to include FINITO switches
+def linked_cases(df_cases,case):
```
patrickbrown4: What would you think about putting all the FINITO-related functions in a new `reeds/finito.py` module and then calling them throughout? It feels like that might make future updates easier (as opposed to searching for all the FINITO references throughout the code.)

reeds/core/setup/c_model.gms
```diff
@@ -396,6 +396,14 @@ eq_loadcon(r,h,t)$tmodel(t)..
...
     + OP_LOADSITE(r,h,t)$[Sw_LoadSiteCF$(Sw_LoadSiteCF<1)$val_loadsite(r)]
     + CAP_LOADSITE(r,t)$[(Sw_LoadSiteCF=1)$val_loadsite(r)]
+
+* [plus] load for industrial and converted fuel facilities (FINITO),
+* including the PRM for stress periods
+* USE_ELE_FINITO is end-use, so divide by (1-distloss) to convert it to busbar
+* [MWh/hr = MW]
+$ifthene.linked_load Sw_FINITO_Link==1
+    + (USE_ELE_FINITO(r,h,t) / (1.0 - distloss))$[t_finito(t)] * (1 + prm(r,t)$h_stress(h))
+$endif.linked_load
```
patrickbrown4, line 404: (nitpick) elsewhere we do this through dollar conditionals
```suggestion
    + (
        (USE_ELE_FINITO(r,h,t) / (1.0 - distloss)) * (1 + prm(r,t)$h_stress(h))
    )$[t_finito(t)$(Sw_FINITO_Link=1)]
```

reeds/core/setup/c_model.gms
```diff
@@ -3572,6 +3580,12 @@ eq_h2_demand(p,t)$[(sameas(p,"H2"))$tmodel(t)$(yeart(t)>=h2_demand_start)$(Sw_H2
     + sum{(i,v,r,h)$[valgen(i,v,r,t)$h2_combustion(i)$h_rep(h)],
             GEN(i,v,r,h,t) * hours(h) * h2_combustion_intensity * heat_rate(i,v,r,t)
     }
+
+* hydrogen demand from industry when linked with FINITO: demand [MMBtu/yr] * conversion [metric tons-H2/MMBtu-H2]
+* TODO: should we disable exogenous H2 demand when linked?
+$ifthene.linked_h2_nat Sw_FINITO_Link==1
```
patrickbrown4, line 3559: Could you use the dollar-conditional approach suggested above throughout (here and in `d_objective.gms`), since we don't have `$ifthene` elsewhere in this script?

reeds/core/setup/c_model.gms
```diff
@@ -3845,6 +3864,11 @@ eq_co2_capture(r,h,t)
 
 * capture from DAC
     + sum{(i,v)$[dac(i)$valcap(i,v,r,t)$i_p(i,"DAC")], PRODUCE("DAC",i,v,r,h,t) }$Sw_DAC
+* capture from industry when linked with FINITO [metric_tons-CO2/hr]: 
+* calculation: hours_per_year [yrs/hr] * capture [scaled_metric_tons-CO2/yr] / co2_scale [scaled_metric_tons-CO2/metric_tons-CO2]
```
patrickbrown4, line 3841: 1. The first line of this comment looks right but not the second
2. It looks like the units in the comment on the `eq_co2_capture` declaration are off; they should be "metric tons per hour"

reeds/core/setup/c_model.gms
patrickbrown4: Yeah it looks like `h_rep(h)` wasn't applied consistently to the CO2 variables/equations. I think the intent was to leave out the h-dependent CO2 variables/constraints for stress periods just to reduce the model size (it doesn't feel like we need all these variables at `r,cs,h` resolution; at least that's the thinking behind #2)

reeds/core/setup/c_model.gms
```diff
@@ -3893,6 +3922,18 @@ eq_co2_sink(r,h,t)$[tmodel(t)$Sw_CO2_Detail$(yeart(t)>=co2_detail_startyr)]..
 
 * net trade
     + sum{rr$co2_routes(r,rr), CO2_FLOW(rr,r,h,t) - CO2_FLOW(r,rr,h,t) }
+
+* (ReEDS-FINITO) extraction - use of CO2 [metric_tons-CO2/hr] 
+*calculation: hours_per_year [yrs/hr] * (1 / co2_scale [scaled_metric_tons-CO2/metric_tons-CO2]) * extraction/use [scaled_metric_tons-CO2/yr]
```
patrickbrown4, line 3900: Can you double check all the units in the comments? It seems like some might be out of date.

reeds/core/setup/d_objective.gms
```diff
@@ -19,7 +19,16 @@ Variable    Z        "--$-- total cost of operations and investment, scale varie
...
 * objective function is the sum over modeled years of the investment
 * and operations components
-eq_ObjFn.. Z =e= cost_scale * sum{t$tmodel(t), Z_inv(t) + Z_op(t) } ;
+eq_ObjFn.. Z =e= cost_scale * (
+* electricity and H2 costs
+                    sum{t$tmodel(t), Z_inv(t) + Z_op(t) } 
+* economy-wide costs from FINITO: deflate from $2018 to $2004 
+* and remove any FINITO scaling
+$ifthene.linked_objective Sw_FINITO_Link==1
+                    + deflator('2018') / obj_scale * sum{t$[t_finito(t)], Z_finito(t) }
```
patrickbrown4: Can you please read 2018 from an input file somewhere instead of hardcoding?
It shows up in `report.gms` too.

reeds/core/solve/2_temporal_params.gms
```diff
@@ -899,6 +899,14 @@ gasprice_adj_r(r,h)$(Sw_GasPriceAdjMethod = 0) = 1 ;
 gasprice_adj_cendiv(cendiv,h)$(Sw_GasPriceAdjMethod = 0) = 1 ;
 
 
+*=============================================
+* ----- ReEDS-FINITO temporal parameters -----
+*=============================================
+$ifthene.linked_finito_temporal_params %GSw_FINITO_Link% ==1
+$include finito/model/linked_finito_temporal_params.gms
```
patrickbrown4: At least I assume this change is required for Windows, but if not we should get rid of `%ds%` everywhere (in a separate PR)
```suggestion
$include finito%ds%model%ds%linked_finito_temporal_params.gms
```

reeds/core/solve/3_solve_oneyear.gms
```diff
@@ -283,6 +290,12 @@ $include reeds%ds%core%ds%solve%ds%4_post_solve_adjustments.gms
 tfix("%cur_year%") = yes ;
 $include reeds%ds%core%ds%solve%ds%5_varfix.gms
 
+*** Fix FINITO decision variables to their optimized levels for this solve year
+$ifthene.linked_finito_varfix %GSw_FINITO_Link% == 1
+* Industrial variable fix
+$include finito/model/finito_varfix.gms
```
patrickbrown4:
```suggestion
$include finito%ds%model%ds%finito_varfix.gms
```

reeds/core/setup/c_model.gms
```diff
@@ -2888,7 +2896,7 @@ eq_national_gen(t)$[tmodel(t)$national_gen_frac(t)$Sw_GenMandate]..
 * ---------------------------------------------------------------------------
 
 *gas used from each bin is the sum of all gas used
-eq_gasused(cendiv,h,t)$[tmodel(t)$((Sw_GasCurve=0) or (Sw_GasCurve=3))]..
+eq_gasused(cendiv,h,t)$[tmodel(t)$tfuel(t)$((Sw_GasCurve=0) or (Sw_GasCurve=3))]..
```
patrickbrown4, line 2872: Note that I didn't try to check for cases where `tfuel(t)` should be applied but is not

reeds/input_processing/hourly_load.py
```diff
@@ -596,6 +596,85 @@ def reaggregate_to_model_regions(
...
+
+    # any missing region or model year in load_finito.csv shows up as NaN
+    if result.isnull().any().any():
+        raise ValueError(
+            'FINITO reference load is missing regions or years present in the '
+            'ReEDS load data; check load_finito.csv'
+        )
+
+    # Validation check: FINITO reference load should not exceed baseline load.
+    # If it does, clip to zero and report
```
patrickbrown4: Should this raise an exception instead?

runreeds.py
```diff
@@ -761,6 +768,74 @@ def setup_window(
...
+    df_cases_finito = df_cases_finito[['Choices', 'Default Value']]
+
+    ### load the scenario-specific switches from ~\FINITO\cases_linked.csv
+    cases_linked_path = os.path.join(df_cases[case]['finito_dir'],f"cases_{df_cases[case]['finito_cases_file']}.csv")
+    df_cases_suf_finito = pd.read_csv(cases_linked_path, dtype=object, index_col=0)
+    ## check that case names are unique in cases_linked.csv
+    # grab the scenario names **exactly** as they appear in the csv file
+    with open(cases_linked_path, 'r', newline='') as csvfile:
+        reader = csv.reader(csvfile)
+        header = next(reader)
```
patrickbrown4: Both are fine but a one-line would be `pd.read_csv(cases_linked_path, header=None).iloc[0].values`

patrickbrown4 (commented):
Thanks, it's cool to see this coming together.

I guess I have a set of general concerns about maintainability - questions like:
- What's the proposed approach for keeping the linked models working together? It seems hard when one is not open source.
  - How do we know when a change to ReEDS (variable/parameter names, indices, concepts) breaks the FINITO linkage? (We've had such issues with reeds2plexos/r2x; running the r2x test as part of the github runner seems to help a lot.)
- There's a lot of duplication of ReEDS data/code within the FINITO repo; what happens when the ReEDS side changes?

I don't think those concerns should keep us from moving forward, but it might be helpful to discuss as a team what the expectations are. For example, the PR template specifies to run the full `cases_test.csv` for "changes to model structure, spatial/temporal resolution, or other large changes"; since a FINITO case is now included in `cases_test.csv`, does that mean everyone who makes big changes to ReEDS should get set up to run FINITO? Or do we leave that to periodic checks by a FINITO developer (in which case maybe we should move it from `cases_test.csv` into a separate file like `cases_examples.csv`)?

I do think we should consider moving the new python code to a dedicated `reeds.finito` module to help with testing/maintenance/organization, though I can't think of a good way to do it for the GAMS code.

cases_test.csv
bsergi: The original motivation for two was to have a quick test and a test with the recommended defaults for a national-scale linked run. I'll move the latter to the `cases_examples.csv`.

My thinking for including the quick test here was that it would be nice as part of Kennedy's regular checks to run a linked case to see if it's working or not. There wouldn't necessarily be an obligation to fix before merging, but it could be helpful for tracing when something broke (it already seems like we have some test cases that are [failing](https://github.com/ReEDS-Model/ReEDS/releases/tag/2026.08.03) in the tagged versions).

cases.csv
bsergi: agreed

cases.csv
bsergi: I'm open to the idea, but how would this work given that FINITO starts in 2018 and StartMarkets is set to 2026? Are you proposing to try not running FINITO until 2026?

reeds/core/solve/2_temporal_params.gms
bsergi: Thought I rooted all these out previously but looks like I missed a few.

reeds/core/solve/3_solve_oneyear.gms
bsergi: The refer to the same thing--standardized this one to match.

reeds/input_processing/copy_files.py
bsergi: Agreed on all counts. It actually looks like the FINITO copy files does the conversion earlier so we don't even need this block.

reeds/core/terminus/report.gms
bsergi: This chunk will be needed once a PR on the FINITO side is finished. If it's not done but the time this one is ready I'll remove an make this an issue.

reeds/core/terminus/report.gms
bsergi: I don't think so. This is just a recalculation of the quantity of gas consumed by ReEDS so it should be the same for `tfuel` and `not tfuel` years. It seems like the only reason we have conditionals here is because with Sw_GasCurve = 0 or 3 you can take a shortcut to this quantity by summing `GASUSED` (although the sum for the quantity in `eq_gasused` omits SMR gas consumption that's found here, so I'm not sure if that approach has been maintained).

reeds/core/setup/c_model.gms
bsergi: `USE_H2_FINITO` is defined in FINITO, so the dollar-conditional approach wouldn't work for standalone ReEDS. Can you talk through your concerns about using the $ifthene approach? We do make use of these already in other parts of the model (`b_inputs.gms`, `d_mga.gms`). 

If we really wanted to use the dollar-conditional we could define the linked parameters and set to zero for a standalone ReEDS run, but that feels a bit more duplicative and fragile to me.

Issue #43, opened 2026-04-20
Docs: Describe input files consistently


patrickbrown4:
## Summary

We currently have a mix of formats for providing data on input files:

- [sources.csv](https://github.com/ReEDS-Model/ReEDS/blob/cd0ee68f684ab61e107e07da6dbe4c8bf27c1389/sources.csv#L1) and [sources_documentation.md](https://github.com/ReEDS-Model/ReEDS/blob/cd0ee68f684ab61e107e07da6dbe4c8bf27c1389/sources_documentation.md#L1)
- [docs/source/user_guide.md](https://github.com/ReEDS-Model/ReEDS/blob/cd0ee68f684ab61e107e07da6dbe4c8bf27c1389/docs/source/user_guide.md#L1)
- [inputs/transmission/README.md](https://github.com/ReEDS-Model/ReEDS/blob/cd0ee68f684ab61e107e07da6dbe4c8bf27c1389/inputs/transmission/README.md#L1), [inputs/reserves/README.md](https://github.com/ReEDS-Model/ReEDS/blob/cd0ee68f684ab61e107e07da6dbe4c8bf27c1389/inputs/reserves/README.md#L1), etc.
- Comments at the top of `.csv` files (e.g. [outage_forced_temperature_murphy2019.csv](https://github.com/ReEDS-Model/ReEDS/blob/cd0ee68f684ab61e107e07da6dbe4c8bf27c1389/inputs/plant_characteristics/outage_forced_temperature_murphy2019.csv#L1))

I've bounced around on `README.md` vs `user_guide.md`, but my current thinking is below. If you have another proposal, please go ahead and add it, either to this text or in a comment; it'd be best to discuss and try to come to consensus before working on implementation.

## Proposal 1
- Remove `sources.csv` and `sources_documentation.md`
  - I'm curious how other folks feel about these. I never use them; I think a 4000-line programmatically generated markdown file is not very readable, they're often out of date, a single .csv cell is often too small to fully describe a file, and for me they slow down development and PR reviews.
  - I'd be interested to hear if anyone is using them though, and if so, how.
- Move documentation of files in `inputs/{subfolder}` from `docs/source/user_guide.md` to `inputs/{subfolder}/README.md`
  - Sorry, I know I originally suggested moving the `README.md` files to `user_guide.md`. But to me it feels like `user_guide.md` has gotten more scattered and hard to navigate as we've dropped more stuff into it.
[...]

PR #138, opened 2026-07-08
Add regional interconnection costs

Max-Vanatta:
## Summary

Replaces the flat intra-zone point-of-interconnection (POI) / network-reinforcement adder (`GSw_TransIntraCost`) with an optional **increasing-cost supply curve**. Successive amounts of new capacity built in a zone pay a higher marginal reinforcement cost, applied to **all** technologies through the existing `INV_POI` / `eq_POI_cap` mechanism.

While this is referred to as the POI in the code currently, this is purely the scaffolding that was used to implement the inclusion. POI in reV is separate from the reinforcement costs (POI is not the network costs). In this implementation, the POI from reV is still included but the reinforcement assumptions previously in reV are stripped out and replaced by the binned costs.

The bin structure reuses the VRE resource-supply-curve pattern (`rtscbin` as the indexer rather than `rscbin`): each bin has an incremental capacity width and a per-bin cost. The flat cost is the degenerate one-bin case, so **`numpoibins = 1` (the default) reproduces current behavior exactly** — the binned method is fully opt-in.

Method documentation: [`inputs/transmission/poi_supply_curve.md`](inputs/transmission/poi_supply_curve.md).

### Implementation notes

- **Input processing** (`reeds/input_processing/`): `make_poi_supply_curve.py` (new) builds the zonal cost bins from `raw_interconnection_TSC_data.csv` via optimal capacity-weighted least-squares segmentation. `transmission.py::write_poi_supply_curve` writes `poi_supply_curve.csv` (flat `GSw_TransIntraCost` when `numpoibins=1`, otherwise the binned curve with an unlimited `GSw_POIUpperCost` backstop bin). `raw_interconnection_TSC_data.csv` is the single source for every spatial resolution (like the hashed transmission cost files) — no per-zone-set fallback; a run whose regions it does not cover fails loudly (matching the transmission cost/distance validation).
- **GAMS**: `INV_POI(r,t)` → `INV_POI(r,rtscbin,t)`; new set `rtscbin`, params `cost_poi_bin`/`cap_poi_bin`/`poi_bin_feas`, and `eq_POI_binlim` (incremental bin-width cap) in `b_inputs.gms`/`c_model.gms`. Objective (`d_objective.gms`), var-fixing (`5_varfix.gms`), and reporting (`report.gms`, `report_params.csv`) updated to carry the `rtscbin` index. The single zonal curve applies to every technology (the free existing capacity `poi_cap_init` and the non-generator POI terms — spur, converter, LCC — attach here too).
- **Double-counting avoided**: when the curve is active (`numpoibins != 1` and `GSw_TransIntraCost > 0`), `reeds/io.py` drops the reinforcement cost already embedded in the VRE resource supply curves (all regions, since coverage is enforced), so it is not counted twice. Retained unchanged when `numpoibins = 1`.

### Additional changes

- `reeds/reedsplots.py` + `postprocessing/single_case_plots.py`: plotting helper for the POI supply curve.
- `inputs/transmission/README.md`: provenance for the raw input file; `dollaryear.csv` entry added (`raw_interconnection_TSC_data.csv`, USD2024). The per-zone-set `poi_supply_curve_{GSw_ZoneSet}.csv` placeholder files and their `runfiles.csv` copy entry were removed — the curve is built at run time from the single raw source.

| Switch | Change | Description |
|---|---|---|
| `numpoibins` | **added** | Number of POI cost bins. `1` (default) = legacy flat `GSw_TransIntraCost`; `>1` = binned curve; `0` = defualt (one bin per raw segment). |
| `GSw_POIUpperCost` | **added** | [USD2004/kW] cost of the unlimited backstop bin above the finite binned capacities; only active when `numpoibins != 1`. |
| `GSw_TransIntraCost` | unchanged | Now also serves as the `numpoibins = 1` flat-cost value. |

### Known incompatibilities

- `numpoibins != 1` (binned or defualt curve) requires `GSw_TransIntraCost > 0`; the curve is gated on that switch, so the combination `numpoibins != 1` with `GSw_TransIntraCost = 0` is rejected with a `ValueError` in `runreeds.py` rather than silently charging no reinforcement.
- `numpoibins = 1` (the default) reproduces the legacy flat-cost behavior exactly.

- Interconnection cost data from the [TSC](https://github.nrel.gov/ReEDS/TSC) model with still to be PR-ed interconnection optimization. **TODO:** Add in the exact links and docs (noted in `inputs/transmission/README.md`).

- [ ] **Pending.** This is a model + data change, so full-US reference and full-US decarb comparison reports are required and will be added before merge.
- Test-case configs are included for exercising the feature at ERCOT resolution: `cases_transmission_test.csv` (`ERCOT_default` = flat legacy, `ERCOT_0` = near-zero flat, `ERCOT_regional` = binned curve).

### Details to double-check
- [ ] Included comparison reports for appropriate test cases *(pending)*
- [x] Documentation updated (`inputs/transmission/README.md`, `inputs/transmission/poi_supply_curve.md`, `docs/source/model_documentation.md`)
  - [x] Dollar year recorded (`dollaryear.csv`, USD2024) and converted to 2004$ for GAMS
  - [x] Units are specified
  - [ ] Preprocessing steps documented/committed to ReEDS_Input_Processing *(TSC-side; to confirm)*
  - [x] New large data files handled with .h5 instead of .csv *(only `raw_interconnection_TSC_data.csv` is added, which is small; the per-zone-set curve files were removed in favor of the single raw source)*

- [ ] Zero impact on results of default case (`numpoibins=1` default)
- [x] No large data file(s) added/modified *(only the small `raw_interconnection_TSC_data.csv`)*
- [x] No change to package requirements
- [ ] No change to process flow *(`runreeds.py` gains one compatibility check: `numpoibins != 1` requires `GSw_TransIntraCost > 0`)*

Yes. Claude Code (Anthropic) was used across several sessions as a pair-programming and drafting assistant, with the author directing the design and reviewing/testing all changes performed by the LLM. Specifically, this was involved in:

- **Feature implementation**: planning and scaffolding the conversion of the flat `INV_POI` adder into a binned supply curve — GAMS sets/params/equations (`b_inputs.gms`, `c_model.gms`, `d_objective.gms`), `make_poi_supply_curve.py`, the `transmission.py` writer, and the objective/report/var-fix wiring.
- **Hardening & hygiene**: the USD2024→2004$ dollar-year conversion via `get_inflatable()`/`dollaryear.csv`; replacing most silent-default `sw.get()` switch lookups with direct indexing (keeping `.get()` only where switches are re-read from pre-feature runs); relocating `make_poi_supply_curve.py` into `reeds/input_processing/`; renaming the mislabeled `capacity_MW`→`capacity_GW` column; consolidating and restyling the documentation; and diff-hygiene cleanup.
- **Review response**: a multi-agent code review of the branch drove several fixes — gating the double-count reinforcement drop on `numpoibins != 1` (so defualt mode doesn't double-count), a `runreeds.py` guard rejecting `numpoibins != 1` with `GSw_TransIntraCost = 0`, consolidating the curves to the single `raw_interconnection_TSC_data.csv` source with fail-loud region-coverage validation (removing the per-zone-set placeholder files), and the `sw.get()` backward-compat for postprocessing on older runs.

All LLM-assisted changes were reviewed, directed, and tested by the author.

_Note: parts of this PR — including the summary of LLM usage above — were themselves drafted by Claude Code (from the branch diff and prior session history) and reviewed/revised by the author._


cases_transmission_test.csv
patrickbrown4, line 1: Before merging, remove this one and add a test for the new method to `cases_test.csv`

reeds/core/setup/b_inputs.gms
```diff
@@ -331,6 +331,25 @@ hyd_add_pump('hydED_pumped-hydro-flex') = yes ;
...
 
+* Sets involved with the POI / network-reinforcement cost supply curve.
+* POI reinforcement is charged on a single increasing-cost zonal curve that applies to every
+* technology. bin_upper is an unlimited backstop bin priced at GSw_POIUpperCost. When numpoibins=0
+* (full-reV resolution) the bins are loaded from data (rtscbin.csv written by transmission.py).
+$ifThenE.poifr %numpoibins%==0
+* rtscbin.csv is a single column (one bin label per line, no header), so $ondelim is neither
+* needed nor allowed here (the set's dimension is the universe, which $ondelim cannot infer):
+* a plain $include pastes the labels straight into the set body.
+set rtscbin "POI / network-reinforcement supply curve bins (full-reV; loaded from data)"
```
patrickbrown4: For readability, can you use a name that's more visually distinct from `tscbin` and `rscbin`? Maybe `icbin` (for "interconnection cost bin") or `poibin` (for "point of interconnection (capacity) bin")?

Since the comments refer to them as POI bins (and the variable is already `INV_POI`), `poibin` seems best. (Understanding it conflicts with the reV usage of the term, but the reV component is really just a substation upgrade cost, so should probably be renamed)

inputs/transmission/raw_interconnection_TSC_data.csv
patrickbrown4, line 1: - The PR text says this file uses zonehashes like transmission cost, but it does not. I can't remember if we discussed; zonehashes are preferred but either format is ok at this point. If zone names are used instead of hashes, the file should be moved to the associated `inputs/zones` folder. (See note elsewhere on how a given zone name can be used differently across `GSw_ZoneSet`s.)
  - If they are indexed by zonehash, add a reader function that pulls the data for the specified `GSw_ZoneSet`
- (nitpick) "raw" and "data" don't say much, and it can be helpful to keep filenames short when possible; could just call it "interconnection_cost_{scenario}.csv"
  - The upgrade curve also depends on tech assumptions (like "is reconductoring with advanced conductors allowed" and "what's the maximum voltage for new lines allowed in a region"), so I expect we're going to have multiple scenarios (at least with/without reconductoring). To support that, you could use a new switch like `GSw_TransUpgradeMethod` (see the [pb/reconductor](https://github.nrel.gov/ReEDS/ReEDS-2.0/tree/pb/reconductor) for an example of how it's used for different reconductoring scenarios)
- Round to fewer significant figures (should be ok with 2 decimals for GW and 0 for the rest)

reeds/core/setup/b_inputs.gms
```diff
@@ -331,6 +331,25 @@ hyd_add_pump('hydED_pumped-hydro-flex') = yes ;
...
     rscfeas(i,r,rscbin) "feasibility set for technologies that have resource supply curves" ;
 
+* Sets involved with the POI / network-reinforcement cost supply curve.
+* POI reinforcement is charged on a single increasing-cost zonal curve that applies to every
+* technology. bin_upper is an unlimited backstop bin priced at GSw_POIUpperCost. When numpoibins=0
+* (full-reV resolution) the bins are loaded from data (rtscbin.csv written by transmission.py).
+$ifThenE.poifr %numpoibins%==0
+* rtscbin.csv is a single column (one bin label per line, no header), so $ondelim is neither
+* needed nor allowed here (the set's dimension is the universe, which $ondelim cannot infer):
+* a plain $include pastes the labels straight into the set body.
```
patrickbrown4: Personally I don't think this kind of thing needs to be described (otherwise b_inputs.gms would be full of comments like this)
```suggestion
```

reeds/core/setup/b_inputs.gms
```diff
@@ -331,6 +331,25 @@ hyd_add_pump('hydED_pumped-hydro-flex') = yes ;
...
+set rtscbin "POI / network-reinforcement supply curve bins (full-reV; loaded from data)"
+/
+$offlisting
+$include inputs_case%ds%rtscbin.csv
+$onlisting
+/ ;
+$else.poifr
+set rtscbin "POI / network-reinforcement supply curve bins" /bin1*bin%numpoibins%, bin_upper/ ;
+$endIf.poifr
```
patrickbrown4: We're trying to move the definition of sets and parameters upstream (out of GAMS), so can you move this if/then logic to `transmission.py` and just write the final set of bins from there?

reeds/core/setup/b_inputs.gms
```diff
@@ -1280,6 +1299,46 @@ $offdelim
...
+* as for the flat Sw_TransIntraCost) so it multiplies INV_POI [MW] directly in the objective.
+parameter cost_poi_bin(r,rtscbin) "--$/MW-- POI / network-reinforcement cost in each bin" ;
+parameter cap_poi_bin(r,rtscbin)  "--MW-- incremental capacity width available in each POI bin (0 = unlimited)" ;
+
+* Binned POI supply curve loaded from inputs_case/poi_supply_curve.csv (long format:
+* r, rtscbin, sc_cat in {cost ($/kW), cap (MW)}, value), written by transmission.py. The region
+* dimension is read against the GAMS universe (*) rather than the model region set r so that
+* regions present in the file but filtered out of the model by GSw_Region do not raise a domain
+* error.
+parameter poi_sc_dat(*,rtscbin,sc_cat) "--$/kW or MW-- POI / network-reinforcement supply curve data by region, bin, and category"
```
patrickbrown4: For everything else we downselect to modeled regions before getting to `b_inputs.gms`; please do the same here too, so you can change the `*` to `r`.

patrickbrown4: "$/kW or MW" could be unclear. For new parameters I think it's safest to keep one set of units per parameter, so can you split this up into two parameters, one for cost and one for capacity (similar to the `tsc_binwidth`, `tsc_forward`, `tsc_reverse` parameters)?

reeds/core/setup/b_inputs.gms
```diff
@@ -1280,6 +1299,46 @@ $offdelim
...
+* file does not cover.
+cost_poi_bin(r,"bin1")$Sw_TransIntraCost = Sw_TransIntraCost * 1000 ;
+
+* Overlay the supply-curve data for any (region,bin) it covers (only model regions r match).
+* poi_sc_dat costs are already in the model dollar year (2004$): transmission.py deflates them from
+* their input dollar year (USD2024, registered in inputs/transmission/dollaryear.csv) when writing
+* poi_supply_curve.csv, so here they are only unit-converted from $/kW to $/MW with *1000
+* (capacities in MW are used as-is), exactly as for the flat Sw_TransIntraCost.
+cost_poi_bin(r,rtscbin)$poi_sc_dat(r,rtscbin,"cost") = poi_sc_dat(r,rtscbin,"cost") * 1000 ;
+cap_poi_bin(r,rtscbin)$poi_sc_dat(r,rtscbin,"cap")   = poi_sc_dat(r,rtscbin,"cap") ;
```
patrickbrown4: Given the direction of #143, would you mind doing all of this in the `input_processing` scripts and then just loading `cost_poi_bin` and `cap_poi_bin` directly?

cases.csv
```diff
@@ -70,6 +70,7 @@ numbins_upv,Number of interconnection supply curve bins for upv; options are 20
 numbins_csp,Number of interconnection supply curve bins for csp,int,5,
 numbins_geohydro_allkm,Number of interconnection supply curve bins for hydrothermal geothermal,int,10,
 numbins_egs_allkm,Number of interconnection supply curve bins for deep enhanced geothermal,int,10,
+numpoibins,Number of POI / network-reinforcement interconnection cost bins (increasing cost per MW of new capacity above existing); 1 reproduces the legacy flat Sw_TransIntraCost,int,1,
```
patrickbrown4: In general, the comments and docs should only refer to what the model currently does, so try to avoid the use of terms like "legacy" (or "currently", etc.). (The exception is the 134 zones, which are still supported (and probably will need to be forever for downstream compatibility) but are no longer the default.)

```suggestion
numpoibins,Number of POI / network-reinforcement interconnection cost bins; uses GSw_TransIntraCost for all additions if set to 1,int,1,
```

cases.csv
```diff
@@ -321,6 +321,7 @@ GSw_TransCapMaxAllTypes,[GW] maximum transmission capacity allowed across each i
 GSw_TransCapMaxTypes,Transmission types to limit according to GSw_TransCapMaxEachType and GSw_TransCapMaxAllTypes [0 = none; 1 = all; 2 = VSC; 3 = all DC],0; 1; 2; 3,1,
 GSw_TransConductor,Which conductor to assume for new AC transmission ('acss' is a mix of ACSS and ACSR as specified by the MISO Transmission Cost Estimation Guide; 'acsr' is just ACSR),^(acss|acsr)$,acsr,
 GSw_TransIntraCost,[USD2004/kW] intra-zone network reinforcement cost for all technologies (100 USD2022/kW using gas cost from slide 14 of https://emp.lbl.gov/publications/generator-interconnection-costs),N/A,65,
+GSw_POIUpperCost,[USD2004/kW] cost applied to all POI / network reinforcement above the binned supply-curve capacities (unlimited backstop bin_upper); only active when numpoibins>1,N/A,2000,
```
patrickbrown4:
```suggestion
GSw_POIUpperCost,[USD2004/kW] cost applied to all POI / network reinforcement above the binned supply-curve capacities; only active when numpoibins>1,float,2000,
```

runreeds.py
```diff
@@ -287,8 +291,36 @@ def check_compatibility(sw):
             f"GSw_Region={sw['GSw_Region']}, GSw_GasCurve={sw['GSw_GasCurve']}"
         )
 
+    if (int(sw['numpoibins']) != 1) and (float(sw['GSw_TransIntraCost']) == 0):
+        raise ValueError(
+            'The binned/native POI reinforcement curve (numpoibins != 1) is gated on '
+            'GSw_TransIntraCost > 0; with GSw_TransIntraCost=0 the entire curve is silently '
+            'disabled (no reinforcement cost is charged).\n'
+            'Set GSw_TransIntraCost > 0 to use the curve, or numpoibins = 1 for the flat legacy '
+            f"cost.\nnumpoibins={sw['numpoibins']}, GSw_TransIntraCost={sw['GSw_TransIntraCost']}"
```
```suggestion
            'Set GSw_TransIntraCost > 0 to use the curve, or numpoibins = 1 to use GSw_TransIntraCost for all capacity additions '
            f"\nnumpoibins={sw['numpoibins']}, GSw_TransIntraCost={sw['GSw_TransIntraCost']}"
```

reeds/input_processing/writesupplycurves.py
```diff
@@ -150,6 +156,9 @@ def main(
         "geohydro": int(sw.numbins_geohydro_allkm),
         "egs": int(sw.numbins_egs_allkm),
     }
+    # Use agglevel_variables function to obtain spatial resolution variables
```
patrickbrown4: Note that `agglevel_variables` has been removed on main

inputs/transmission/README.md
```diff
@@ -59,6 +59,14 @@ Calculated using the [TSC](https://github.nrel.gov/pbrown/TSC) model as describe
 - `newlinks_offshore_backbone.csv`: Candidate connections between offshore zones
   - Similarly formatted files for candidate connections between offshore and coastal land-based zones are found at `inputs/zones/{GSw_ZoneSet}/newlinks_offshore_radial.csv`
 
+- `raw_interconnection_TSC_data.csv`: Raw cumulative interconnection / network-reinforcement cost curve per ReEDS zone from the [TSC](https://github.nrel.gov/ReEDS/TSC) model (nodal network data as described by [Brown et al.](https://arxiv.org/abs/2308.03612)). This is the **single source** for the point-of-interconnection (POI) / network-reinforcement cost supply curve at every spatial resolution (region names span all zone sets' `hierarchy.csv` vocabularies, analogous to the hashed transmission cost files); there is no per-zone-set fallback. One cumulative supply-curve point per row:
```
patrickbrown4: "region names span all zone sets' `hierarchy.csv` vocabularies, analogous to the hashed transmission cost files" - not sure I follow but it doesn't seem right; the file uses zone names but those are not unique (different zone sets have different definitions for zones like CA_N), and the hierarchy files don't use the zonehashes.

Since these data are only for zones (not interfaces) and don't need to be added for counties, I think it makes sense to go ahead and use the zonehash approach.

Either way could you revise this file (and `model_documentation.md`, as discussed)?

inputs/transmission/README.md
```diff
@@ -59,6 +59,14 @@ Calculated using the [TSC](https://github.nrel.gov/pbrown/TSC) model as describe
...
+  - `region`: ReEDS zone (must match the zone set's `hierarchy.csv`)
+  - `capacity_GW`: cumulative POI capacity [GW]; the row with zero cumulative cost marks the zone's existing (free) capacity
+  - `cum_cost_$`: cumulative reinforcement cost [USD2024]
+  - `slope_$/kW`: marginal cost from the previous point [USD2024/kW] (informational)
+  - When `numpoibins > 1` (or `= 0`, native), `reeds/input_processing/transmission.py` builds the binned curve `inputs_case/poi_supply_curve.csv` (`*r, rtscbin, sc_cat in {cost, cap}, value`) from this file via `reeds/input_processing/make_poi_supply_curve.py`, deflating the costs from USD2024 (registered in `dollaryear.csv`) to the model dollar year; with `numpoibins = 1` the model instead uses the flat `GSw_TransIntraCost` adder. A run whose model regions this file does not cover fails loudly (matching the transmission cost/distance validation).
```
patrickbrown4: Not sure what "native" means here but the default is 1, not 0

inputs/transmission/README.md
```diff
@@ -59,6 +59,14 @@ Calculated using the [TSC](https://github.nrel.gov/pbrown/TSC) model as describe
...
+  - **TODO:** confirm the exact TSC dataset version / study citation before publishing.
```
patrickbrown4: Reminder to remove TODOs before merging

inputs/transmission/poi_supply_curve.md
patrickbrown4, line 1: `README.md` is the only file that renders when you navigate to the `inputs/transmission` folder on github, so I would just move this to a new section in that file; otherwise it doesn't get rendered anywhere. (Our docs are already scattered, as discussed in #43, so personally I'd lean away from adding another level of organization. We don't have a great system for script-specific documentation, so there's still more organizational work to be done, but the `inputs/*/README.md` files seem the closest.)

inputs/transmission/poi_supply_curve.md
```diff
@@ -0,0 +1,49 @@
+# POI / Network-Reinforcement Cost Supply Curve
+
+## Overview
+
+ReEDS charges new capacity in each zone a point-of-interconnection (POI) / network-reinforcement cost through the `INV_POI` variable and `eq_POI_cap`, which requires POI capacity to cover all generation capacity built in the zone.
+By default this cost is a single flat adder (`GSw_TransIntraCost`, applied to every technology).
```
patrickbrown4: Not really; I would say the default representation of network reinforcement costs is the geospatial treatment described at https://reeds-model.github.io/ReEDS/model_documentation.html#network-reinforcement

inputs/transmission/poi_supply_curve.md
```diff
@@ -0,0 +1,49 @@
...
+This feature replaces the flat adder with an **increasing-cost supply curve**: successive amounts of new capacity in a zone cost more per MW, so the marginal reinforcement cost steps up with cumulative build.
+
+The bin structure reuses the VRE resource-supply-curve pattern (`rscbin`): each bin has an incremental capacity width and a per-bin cost, and the LP fills the cheapest bins first under cost minimization — no ordering / SOS constraint is needed as long as bin costs strictly increase.
```
patrickbrown4: Revealing my ignorance but what is an "SOS constraint"?

inputs/transmission/poi_supply_curve.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Inputs
+
+The zonal curve is built at run time by `reeds/input_processing/transmission.py`:
+
+- `write_poi_supply_curve` writes `inputs_case/poi_supply_curve.csv` (`*r, rtscbin, sc_cat in {cost, cap}, value`).
+  With `numpoibins = 1` it is the flat `GSw_TransIntraCost`.
+  Otherwise the zonal bins are built from `raw_interconnection_TSC_data.csv` via `make_poi_supply_curve.make_regional_poi_bins` (optimal capacity-weighted least-squares segmentation to `numpoibins` bins), with an unlimited `bin_upper` backstop appended at `GSw_POIUpperCost`.
+
+`raw_interconnection_TSC_data.csv` is the single source of these curves for every spatial resolution (like the hashed transmission cost files) — there is no per-zone-set fallback. When `numpoibins != 1`, a run whose model regions the file does not cover fails loudly (matching the transmission cost/distance validation).
```
patrickbrown4: To me at least, not clear what "there is no per-zone-set fallback" (used multiple times) means. If the point is that only one `GSw_ZoneSet` value is supported, it'd be best to add that to `runreeds.check_compatibility()`.

inputs/transmission/poi_supply_curve.md
```diff
@@ -0,0 +1,49 @@
...
+
+The wind / solar / CSP / geo resource supply curves already embed a network-reinforcement cost (`cost_reinforcement_usd_per_mw`).
+When the binned POI method is active (`numpoibins > 1` **and** `GSw_TransIntraCost > 0`), `reeds/io.py` drops that embedded reinforcement for all regions (recomputing `cost_total_trans` as spur + POI connection only) so it is not counted twice.
+With `numpoibins = 1` it is retained unchanged, preserving legacy results.
+
+## Notes and assumptions
+
+- **Validate before switching real scenarios to multi-bin.** Confirm the single-bin (`numpoibins = 1`) case reproduces current flat-cost results, and that the binned costs and dollar year are correct, before relying on multi-bin runs.
+- The binned POI is zonal and technology-agnostic: it replaces the site / bin / region-specific VRE reinforcement granularity with a zonal representation, and extends reinforcement pricing to **all** technologies (not just VRE).
+- A parallel-variable alternative (`INV_REINF`, leaving `INV_POI` untouched) was considered but not implemented; extending `INV_POI` in place avoids duplicate POI accounting.
```
patrickbrown4: `INV_REINF` doesn't show up in the code

reeds/input_processing/make_poi_supply_curve.py
patrickbrown4, line 1: 1. What do you think about sticking to the convex hull method for now? Reasoning:
    - Consistency with ITL upgrade supply curve
    - Simpler, with fewer things to go wrong (I'm not sure I follow the `optimal_bins` method, particularly with the single-letter variable names)
    - If the segments are linear, it doesn't seem like we gain much by breaking them up
2. If we stick to the convex hull, I think you'd drop the `numpoibins` switch too, and replace the combined `numpoibins`/`GSw_TransIntraCost` logic with a single switch - is that right?
    - One way to do it would be like the `GSw_MethaneLeakageScen` and `GSw_GWP` switches, which accept either a named scenario (which in your case would be the interconnection curve scenario) or a float (in your case the old flat `GSw_TransIntraCost`). But if that's too much of a headache, you could instead just make something like a simple `GSw_TransIntraCurve` switch that is 0 for the geospatial method and 1 for the curve
    - Oh I think I understand what you mean by "0 (native)" for `numpoibins`. So I think what I'm suggesting is just to keep the 0/1 functionality and update the name
3. If it's simplified to convex hull then I think you could drop this script, but if not, how about picking one approach or the other: either run it outside of a ReEDs run (in which case it's probably most natural to keep it in the TSC repo) or run in `transmission.py` (in which case it could just be a function in that module)

reeds/input_processing/transmission.py
```diff
@@ -641,6 +640,90 @@ def check_nonac_costs(trancap_fut, trancap_init_energy, transmission_cost_nonac)
...
+        reg = pd.DataFrame({
+            'r': valid_regions['r'], 'rtscbin': 'bin1', 'sc_cat': 'cost',
+            'value': float(sw.GSw_TransIntraCost)})
+    else:
+        ## Build the zonal curve from the raw cumulative interconnection data, re-binned to
+        ## numpoibins (numpoibins=0 -> native: one bin per raw segment), cost capped at
+        ## GSw_POIUpperCost. See reeds/input_processing/make_poi_supply_curve.py.
+        ## raw_interconnection_TSC_data.csv is the single source (curves for every spatial
+        ## resolution). No per-zone-set fallback: if it is missing, or does not cover every model
+        ## region, fail loudly -- the same way get_distances/ITL validation does for transmission.
```
patrickbrown4: I guess I'm on the fence about comments like this. Describing what code does and why is good, but referencing how other functions work seems like it could balloon the amount of commentary and lead to inconsistencies over time. Some of this wording is also repeated near-verbatim in multiple places.

reeds/input_processing/transmission.py
```diff
@@ -709,6 +792,11 @@ def main(case):
     ## Write transmission_cost_ac for R2X
     outputs['transmission_cost_ac'] = transmission_cost_ac
 
+    ### POI / network-reinforcement cost supply curve
+    ## Writes poi_supply_curve.csv (and rtscbin.csv) directly; already region-scoped, so it
+    ## bypasses the outputs/hierarchy-downselect mechanism below.
+    write_poi_supply_curve(case)
+
```
patrickbrown4, line 845: The `outputs` dict is used and expanded on in #139, so would you mind using it here instead of bypassing?

reeds/core/setup/c_model.gms
```diff
@@ -1999,34 +2000,63 @@ eq_prescribed_transmission(r,rr,trtype,t)
...
-* New point-of-interconnection (POI) intra-zone transmission capacity must be
-* added for new generation capacity
+* Point-of-interconnection (POI) / intra-zone network-reinforcement COST accounting.
+* This is a PRICE-ONLY mechanism: new generation is charged a per-MW reinforcement cost off the
+* supply curve and NO deliverable intra-zone transmission capacity is created. INV_POI never
+* enters the power balance, the transmission flow limits, or CAPTRAN; it relieves no flow. The
+* accounting quantity is sized to the zone's interconnected generation capacity (CAP/ilr, AC)
+* solely so the curve price can be levied on it -- it is a cost basis, not transmission capability.
+* A single zonal reinforcement curve covers all technologies (the free existing capacity
+* poi_cap_init and the non-generator POI terms -- spur, converter, LCC -- attach here too).
```
patrickbrown4: I'm not sure about this new emphasis on a price-only mechanism. We require `INV_POI` to be expanded to expand the generation capacity, and we include interconnection miles in the output transmission reporting. It's not interzonal but it's still transmission, and [the docs](https://reeds-model.github.io/ReEDS/model_documentation.html#transmission) frame these as two different types of transmission (intrazonal + interzonal), not that one is "price-only".

This conversation is helpful though because it reminds me that we need to include the data and processing to get reinforcement TW-miles into the outputs. (Sorry, I think I said I would do some of the below in the TSC repo and forgot about it.) (These changes aren't required when we drop the reinforcement costs/distances for county resolution because the distances still get counted as interzonal; here, we need to do it because otherwise there's nothing to replace the reinforcement distances that are dropped.) That involves:
- [ ] TSC
  - [ ] Add line miles to each line of `upgradelog.csv`
  - [ ] When determining the convex hull, sum and record these parameters for each bin:
    - [ ] miles and MW-miles of reconductoring (by voltage would be great but ok to save for next round)
    - [ ] miles and MW-miles of new circuits (")
    - Nice to have (for unrelated supply-curve projects) but ok to skip for first round since not currently used in ReEDS:
      - [ ] MW and number of substation upgrades (")
      - [ ] MW and number of new transformers (by high-side voltage would be great, but again, ok to save for next round)
  - [ ] Decide how to extend for the unlimited top bin
  - [ ] Write that data from TSC for use in ReEDS (see [here](https://github.nrel.gov/ReEDS/TSC/blob/e8457614dc4cf113a6b7b4b723a2b081bd517dc4/analysis/combine_for_reeds.py#L49) for the analogous part of the ITL upgrade supply curve)
- [ ] ReEDS
  - [ ] For parity with current features we would need two things in the core optimization:
    - [x] cost per bin (already done)
    - [ ] Line miles by bin, as a replacement for `distance_reinforcement(i,r,rscbin)` in [`eq_transmission_investment_max`](https://github.com/ReEDS-Model/ReEDS/blob/bb735febca4b9d29235b5120242d9d063aca4246/reeds/core/setup/c_model.gms#L2341), assuming we want to keep supporting limited-transmission scenarios via the `GSw_TransInvMaxLongTerm` switch
  - [ ] In output-processing, include the TW-miles of reinforcements in the outputs
    - That's messy because we currently have two almost-identical functions for it:
      - [pre_spur](https://github.com/ReEDS-Model/ReEDS/blob/bb735febca4b9d29235b5120242d9d063aca4246/postprocessing/bokehpivot/reeds2.py#L1196) in bokeh
      - [calc_reinforcement_spur_capacity_miles](https://github.com/ReEDS-Model/ReEDS/blob/bb735febca4b9d29235b5120242d9d063aca4246/reeds/results.py#L509) in `reeds/results.py`
    - Brian is working on removing the bokeh versions so they only live in one place. So we can focus on the earlier steps for now in case that's done by the time we get to it.

inputs/transmission/poi_supply_curve.md
```diff
@@ -0,0 +1,49 @@
...
+See `README.md` in this folder for the input-file provenance.
+
+## GAMS implementation
+
+The bin dimension folds straight into the existing `INV_POI` mechanism (`INV_POI(r,t)` → `INV_POI(r,rtscbin,t)`):
+
+- `b_inputs.gms` — set `rtscbin`; parameters `cost_poi_bin(r,rtscbin)` `[$/MW]`, `cap_poi_bin(r,rtscbin)` `[MW]` (incremental bin width, `0 = unlimited`), and `poi_bin_feas(r,rtscbin)`. Any region the file omits defaults to the flat `Sw_TransIntraCost`.
+- `c_model.gms` — variable `INV_POI(r,rtscbin,t)`; `eq_POI_cap` (POI covers all zone capacity, now summed over `rtscbin`); `eq_POI_binlim(r,rtscbin,t)` caps cumulative `INV_POI` at each bin width, which is what makes the marginal cost step up.
+- `d_objective.gms` — the capital and FOM terms use `cost_poi_bin` summed over `rtscbin` in place of the flat `Sw_TransIntraCost` scalar.
+- `5_varfix.gms`, `report.gms`, `report_params.csv` — `INV_POI` fixing and reporting take the `rtscbin` index. Reported `poi_capacity` is a cost-accounting quantity, **not** deliverable transmission capacity, and relieves no flow.
```
patrickbrown4: It might be helpful to discuss as a group what kinds of things we think are missing from the docs that would be helpful to include, and where they should live.

The docs problem I've been dealing with most often lately is consistency: code doing one thing and docs saying something else (in some cases, saying different things in different places). The code is ground truth, and every time we describe something in another place, it's another place where things can fall out of sync. That makes me reluctant to add more description of explicit code features in more different places; to me, documentation of nitty gritty code details (variable names etc.) seems easiest to keep in sync if it's right next to the code.

We don't have sections like this highlighted section for other parts of the code, which makes me inclined to move it to the PR text (which is where we've put these kinds of descriptions in the past), but I don't want to push back on things just because they're new. It would just be a lot of content if we did this for every part of the code, and would be one more thing that needs to be updated every time the code changes.

docs/source/model_documentation.md
```diff
@@ -1859,6 +1855,12 @@ Because new generation capacity is assumed to reuse the network infrastructure f
...
 
+By default the intrazone adder is flat (`GSw_TransIntraCost`, applied to every MW of net new capacity in the zone).
+It can optionally be represented as a regional *increasing-cost supply curve* by setting `numpoibins` greater than 1.
+In that case successive amounts of new point-of-interconnection (POI) capacity in each zone are priced in increasing cost bins (built at run time from `inputs/transmission/raw_interconnection_TSC_data.csv`), with an unlimited backstop bin priced at `GSw_POIUpperCost` so that reinforcement above the binned capacities remains feasible at a high cost.
+When this binned POI supply curve is active (`numpoibins` > 1 and `GSw_TransIntraCost` > 0), the geographically resolved network reinforcement cost embedded in the wind, solar, CSP, and geothermal supply curves is dropped entirely (and their total interconnection cost is recomputed as spur line plus substation/POI connection only) so that network reinforcement is not double counted between the two representations.
+With `numpoibins` = 1 (the default flat adder) the wind, solar, CSP, and geothermal supply-curve reinforcement cost is retained.
+
```
patrickbrown4: In the rest of `model_documentation.md`, we don't refer to switch names or file paths directly in the text; when they need to be referred to, they go in admonition boxes (and we typically only do that for switch names, for the most part leaving out discussion of filepaths).

patrickbrown4 (commented):
Thanks a lot and sorry for the late followup (and the nitpicky comments). I haven't gotten through the whole thing yet, but adding some notes now. Let's discuss how to coordinate the work on getting distances out of the TSC results.

reeds/core/setup/b_inputs.gms
Max-Vanatta: I prefer icbin (and added that to the next commits) due to the reV discussion, but also because it feels like these costs expand past poi only. 

That being said, the rest of the POI constraints would then need to be renamed if we really try to separate the concepts.

reeds/input_processing/make_poi_supply_curve.py
Max-Vanatta: Agreed and implemented

inputs/transmission/poi_supply_curve.md
Max-Vanatta: It is a special order set. Just learned about it also, but I don't think I should include  all methods I -didn't- include in this.

Overall, removing this file and combining it into other areas of documentation as other comments suggest.

reeds/core/setup/c_model.gms
Max-Vanatta: Completely agreed, this emphasis was more of a caution than focus. With the other branch we will make that maps the upgradelog (or other method) then it should be completely defined.

inputs/transmission/poi_supply_curve.md
Max-Vanatta: Agreed completely. I am dropping this file and moving conceptual to docs and then any technical that is not clear to the PR.

PR #139, opened 2026-07-09
Move more inputs to `inputs.h5` and add compatibility with `forecast.py`

patrickbrown4:
## Summary

This PR builds on #61 by moving more inputs to `inputs.h5`, removing another ~10% (~600 lines) of `b_inputs.gms`.

### Implementation notes 

Most of the changes are just normalizing headers in input .csv files and changing `.to_csv()` to `reeds.io.write_input_to_h5()`. Outputs of the following input-processing scripts are moved to `inputs.h5`:
- `calc_financial_inputs.py`: complete
- `transmission.py`: complete
- `copy_files.py`: a smattering (not complete)

More substantive changes are:
- `forecast.py` is adapted to work for parameters in `inputs.h5`
  - `forecast.py` is also moved later in the input-processing pipeline; in its old location it was missing some parameters from `transmission.py`. The new parameters are added to the bottom of `futurefiles.csv`.
- The CO2 storage parameter processing is moved from `b_inputs.gms` to `get_co2_site_char()` in `transmission.py`
  - The `t` index is removed from `cost_co2_stor_bec` because it doesn't vary with time
- The application of `GSw_TransCostMult` is moved upstream to `transmission.py`
- `inputs/waterclimate` files are reshaped from wide to long
- `overwrite` in `reeds.io.write_input_to_h5()` is turned on by default, making the input-processing scripts easier to debug

### Additional changes 

- `reedsplots.plot_diff_maps()`: A `level` parameter is added, and when used in `compare_cases.py`, state resolution is used when runs with different spatial resolution are compared (thanks @kodiobika for the idea and initial implementation)
- `plots.rainbowmapper()`: colormap moved to keyword argument

## Validation, testing, and comparison report(s)

Only rounding differences for the Pacific test case (expected when moving from .csv to .h5): 
[results-v20260708_mainM0_Pacific,v20260708_inputsM2_Pacific.pptx](https://github.com/user-attachments/files/29854892/results-v20260708_mainM0_Pacific.v20260708_inputsM2_Pacific.pptx)

Same for USA_defaults: [results-v20260708_mainK0_USA_defaults,v20260708_inputsK1_USA_defaults.pptx](https://github.com/user-attachments/files/29987843/results-v20260708_mainK0_USA_defaults.v20260708_inputsK1_USA_defaults.pptx)

### 20260805 update

[...]

PR #142, opened 2026-07-12
Add structure for moving `report.gms` to Python

patrickbrown4:
## Summary
Following in the vein of #61 and #139, this PR adds a structure for moving the calculations in `report.gms` to Python and uses it for a couple blocks of outputs as examples.

### Implementation notes 

The proposed structure is:
- Sets, parameters, and output variables necessary for output-processing calculations are added in `report_data.csv` (in `reeds/core/terminus`)
  - These get written to a new intermediate file, `{case}/outputs/results.gdx`
  - Output variables always include all attributes (level, marginal, upper/lower bound, scale)
- The existing `report_dump.py` calls a new `report_calcs.main()`, which returns a dictionary of output parameters calculated in Python. These parameters then get written to `{case}/outputs/{param}.csv` and/or `{case}/outputs/outputs.h5`, as usual.
- `report_calcs.main()` has some helper functions for common operations at the top, followed by blocks of calculations for the different types of outputs.
  - Only the transmission, CO2 storage, and interconnection queue outputs are currently included; additional sections can be moved over in batches in subsequent PRs.

Some notes given the current mix of standards:
- Parameters that are moved from `report.gms` to `report_calcs.py` are listed in a block comment at the top of `report.gms`, so searching for them will turn up both scripts.
- Parameters not calculated in `report.gms` will not show up in `{case}/outputs/rep_{case}.gdx`
  - But `{case}/outputs/outputs.h5` is unchanged
- It's possible that doing these calculations in Python may be slower than doing them in GAMS, particularly for big hourly outputs, but some notes on potential speedups are added as comments in `report_calcs.py`

### Additional changes 

- A new `reeds.results.diff_outputs()` function has been added. This function works like `gdxdiff` but for `outputs.h5` files, and can be used to do a global comparison of outputs between pairs of cases.
- Moved the execution of `calculate_historical_capex()` to `report_dump.py`
- Changed `report_dump.py` errors from 'warn' to 'raise' and added a post-script error check to the call file in `runreeds.py`
- Fixed `results.calc_systemcost()` for `inputs.h5`
- Switched from `allt` to `t` index for `hours_t(allh,t)` and `h_stress_t(allh,t)` (`allt` is unnecessary since these are populated during the run)

Zero change for the USA_defaults case: [results-v20260708_mainK0_USA_defaults,v20260709_reportK0_USA_defaults.pptx](https://github.com/user-attachments/files/29947430/results-v20260708_mainK0_USA_defaults.v20260709_reportK0_USA_defaults.pptx)

### 20260805 update
Still zero change for USA_defaults.

### Details to double-check

No


reeds/core/terminus/report_calcs.py
```diff
@@ -0,0 +1,165 @@
...
+    ## (r,rr,h,trtype,t)
+    dfs['tran_util_h_rep'] = dfs['tran_flow_all_rep'] / dfs['tran_cap_energy']
+    ## (r,rr,allh,trtype,t)
+    dfs['tran_util_h_stress'] = dfs['tran_flow_all_stress'] / dfs['tran_cap_prm']
+    ## (r,rr,trtype,t)
+    dfs['tran_util_ann_rep'] = (
+        (dfs['tran_flow_all_rep'] * g['hours'] / dfs['tran_cap_energy']).groupby(['r','rr','trtype','t']).sum()
+        / g['hours'].sum()
+    )
+    ## NOTE: We here assume that all solve years use the same total weighting for stress
```
patrickbrown4: The method on main is off for years before the last year; see #146. This implementation fixes it and is self-consistent when stress periods are always weighted at 6 hours per solve year, but if #146 is fixed as suggested, we should instead use `hours_h(allh,t)` here to make it future-proof in case we switch to nonuniform stress period weights.

reeds/core/terminus/report_calcs.py
```diff
@@ -0,0 +1,165 @@
+#%% Imports
+import gdxpds
+import pandas as pd
+from pathlib import Path
+from typing import Literal
+
+
+#%% Helper functions
+def get_gams_results(case):
+    print('Loading resultsgdx')
```
bsergi:
```suggestion
    print('Loading results.gdx')
```

reeds/results.py
```diff
@@ -710,4 +713,94 @@ def get_level_map():
...
+    threshold_rel=0,
+    verbose=0,
+):
+    """
+    Diff two {case}/outputs/outputs.h5 files and save the result to outpath.
+
+    Args:
+        casebase: Absolute path to base ReEDS case
+        casecomp: Absolute path to comparison ReEDS case
+        outpath: Absolute path to resulting difference .h5 file
```
bsergi, line 727: perhaps also add here that setting to False just checks the diff without saving to file

reeds/core/terminus/report.gms
```diff
@@ -122,34 +122,53 @@ $include autocode%ds%report_params.gms
...
-*=====================
-
-CO2_CAPTURED_out(r,h,t)$tmodel_new(t) = CO2_CAPTURED.l(r,h,t) ;
-CO2_CAPTURED_out_ann(r,t)$tmodel_new(t) = sum(h,hours(h) * CO2_CAPTURED.l(r,h,t) );
-CO2_STORED_out(r,cs,h,t)$[tmodel_new(t)$csfeas(cs)] = CO2_STORED.l(r,cs,h,t) ;
-CO2_STORED_out_ann(r,cs,t)$[tmodel_new(t)$csfeas(cs)] = sum(h,hours(h) * CO2_STORED.l(r,cs,h,t) );
-CO2_TRANSPORT_INV_out(r,rr,t)$tmodel_new(t) = CO2_TRANSPORT_INV.l(r,rr,t) ;
-CO2_SPURLINE_INV_out(r,cs,t)$[tmodel_new(t)$csfeas(cs)] = CO2_SPURLINE_INV.l(r,cs,t) ;
-
-CO2_FLOW_out(r,rr,h,t)$[(ord(r) < ord(rr))$tmodel_new(t)] = CO2_FLOW.l(r,rr,h,t) + CO2_FLOW.l(rr,r,h,t) ;
```
bsergi, line 142: Shouldn't this be `CO2_FLOW.l(r,rr,h,t) - CO2_FLOW.l(rr,r,h,t) ;`? The sum is used in `eq_co2_transport_caplimit` which makes sense, but if the goal is to just get flows for r < rr region pairs but preserve direction then I would think the second term should be negative as is done with the `tran_flow_rep` calculation. Not sure who is the lead on CO2 these days but might be worth checking with them.

If this is a bug it might make sense to just flag this in an issue and address it in a separate PR since the goal of this one is just to do the reporting and have it match what we have. If we did want to address here I think we would just to `agg='net'` over in `report_calcs.py`.

bsergi: I guess there is the net flow calculation below which is what I was expecting, so maybe this one is fine.

bsergi (approved):
Looks good to me. I added one small fix to get the logging working again.

Do you have any concerns about memory usage as we add more ouputs here? We already read in the whole gdx so it might be fine but wasn't sure. 

You noted that the outputs.h5 file is unchanged--is that saying that you ran the new diff_outputs() function and didn't come up with any differences between this branch and main?

reeds/core/terminus/report.gms
patrickbrown4: Yeah I'm not sure why it's here; the number of permutations doesn't really seem necessary, but because the net flow is also included, it seems intentional to have both. It might be used to identify hours where the pipeline capacity is binding. I figured I'd leave it for parity with the existing outputs, and it could maybe be cleaned up as part of #2.

patrickbrown4:
> Do you have any concerns about memory usage as we add more ouputs here? We already read in the whole gdx so it might be fine but wasn't sure.

Yeah it will be interesting to see how it scales for big `h`-dependent outputs. I added a note to the procedure section: "## NOTE: If calculations slow down for large runs, consider dropping zeros upfront in get_gams_results() to speed up processing".

For the next batch, we should start with whichever calculation is slowest in `report.gms` (printed at the bottom of the report list file) and see how different it is in python. Here are the slowest for a recent USA_defaults run:

```text
---- Profile Summary (435 records processed)
    42.522  20.422GB  47503195 Assignment emit_r (9619)
    30.642  20.422GB  47503228 Assignment emit_irt (27068)
    26.459  19.684GB  47502699 Assignment gen_h (2929940)
    25.211  20.377GB  47502819 Assignment curt_h (20113)
    24.510  19.462GB  47502445 Assignment reqt_quant (1727033)
    23.631  19.762GB  47502720 Assignment gen_h_stress (1940223)
    23.496  20.421GB  47503048 Assignment revenue (33083)
    15.673  20.459GB  47504150 Assignment h2_usage (2414)
    13.213  20.422GB  47503080 Assignment revenue_en_nat (1824)
    12.069  20.422GB  47503067 Assignment revenue_en (36648)
```

> You noted that the outputs.h5 file is unchanged--is that saying that you ran the new diff_outputs() function and didn't come up with any differences between this branch and main?

Yeah that's right. The differences aren't exactly zero, but I also get nonzero differences even for two identical runs on main, so maybe degeneracy-breaking is not as deterministic as I thought. But the scale of the differences for the parameters I moved to python looks good.

patrickbrown4:
I went back and ran a single Pacific case with the new calculations but without removing the old calculations from `report.gms`, to be able to calculate both for a single run. The only difference shown by `results.diff_outputs()` (within `threshold_abs=0.0001` and `threshold_rel=0.0001`) was for `tran_util_ann_rep`, which did actually have a small bug (dividing by all hours, 8766, instead of rep hours, 8760)!

Noting here so that, for future work on this file, we remember to do this step of comparing before/after for the exact same run.

PR #148, opened 2026-07-15
Add HVV convention in docs

rrolph575:
…ntion.

## Summary
Added a subsection under "Fuel Prices" section of documentation saying that the HHV convention is used for heating value in ReEDS.  This is not explicitly stated in the docs and can be a source of confusion for users who are used to input values from IEA, EU, or manufacturers fuel prices.

### Details to double-check
- [WCRD.13600.01.01.01 ] Charge code provided to reviewers

- [ x] Zero impact on results of default case
- [x ] No large data file(s) added/modified
- [x ] No substantive impact on runtime for full-US reference case
- [x ] No substantive impact on folder size for full-US reference case
- [x ] No change to process flow (runreeds.py, reeds/core/solve/solve.py)
- [x ] No change to code organization

I did not use verbatim text but I used an LLM to help me in my research (which I double checked on other sources for correctness).


wesleyjcole (commented):
A few comments:

- What do you think about moving this section to the fossil and nuclear technologies section where these things (fuel costs, heat rate) are first introduced?
- I don't think we need to explain the difference between HHV and LHV, or note locations where one is more common than another.  I think the goal is just so people know we use HHV, which can be pretty short (a single sentence would probably be fine).
- In order to make it easier to identify changes to .md files, we prefer one sentence per line for any multi-sentence sections.
- Would you update the pull request title so that it fits on the single line?

rrolph575:
I removed most of the text so that it is now just 2 sentences with now one sentence per line.  I also moved those sentences to the section you recommended and agree it fits better there.  I shortened the PR title.

docs/source/model_documentation.md
```diff
@@ -680,6 +680,8 @@ Fossil and nuclear technologies are characterized by the following parameters:
 - Scheduled and forced outage rates (%).
 
 Cost and performance assumptions for all new fossil and nuclear technologies are taken from the ATB {cite}`nrel2024AnnualTechnology2024` with options to use the Conservative, Moderate, or Advanced trajectories.
+The EIA and U.S. natural gas is priced using $/MMBtu derived from HHV convention, and heat rates and fuel prices that come from NLR’s ATB are consistent with the HHV convention.
+We use the HHV convention in ReEDS.
```
wesleyjcole: How about something like this:

```suggestion
Thermal plant heat rates and fuel costs are specified in higher heating value (HHV) terms, consistent with the EIA and the ATB. 
```

HHV will also needed added to the acronyms list.

wesleyjcole (commented):
Would you add "HHV" to the acronymns list?

rrolph575:
> Would you add "HHV" to the acronymns list?

Yes, added!

wesleyjcole (approved):

PR #162, opened 2026-07-27
Add random vector sampling method to MGA

bsergi:
## Summary

This PR adds a new capability to the existing MGA method to support random vector sampling across subcategories of the variables being explored.

## Technical details 

The current MGA method applies the minimization (or maximization) to the sum of a variable across all modeled regions (e.g., total U.S. solar capacity or fossil generation). The random vector method assigns randomly-sampled weights to the regional quantities applied before the summation. This technique allows for exploring the regional distribution of capacity and generation results in the near-optimal space. 

As an illustration, the plot below shows the range of natural gas capacity by region (using the 54-zones) from 100 runs with the MGA random vector method with a 1% cost threshold. The MGA max/min squares indicate the results from the previous, nationwide-method; the original min/max approach is better suited for getting a wider range of total capacity, whereas the new method is better geared toward understanding uncertainty at a regional level. 

image

### Implementation notes 

- The weights are generated by the`main_mga_rv` function in `mcs_sampler.py`, with the number of samples (runs) controlled by `GSw_MGA_RV_runs`.
- Regional weights are drawn from a support of {-1,1} with equal probability; including both negative and positive values allows for exploring minimization in one region and maximization in another.
  - The method in `mcs_sampler.py` can support linear sampling from [-1,1] using a uniform distribution. Initial tests suggested this didn't yield as much of a signal so we haven't implemented a switch to control this. 
  - The regionality at which unique weights are sampled is controlled by `GSw_MGA_RV_region` and can be any level in the hierarchy file.
- Currently this method is only supported for `GSw_MGA_Objective = (capacity or generation)`, with weights generated by region and technology group (`i_subtech`).
  - If `GSw_MGA_SubObjective` is an aggregated tech category (e.g., `gentech`), unique weights will be sampled for the different `i_subtech` groups within the larger category.
- Weights are read in by `b_inputs.gms` and applied to the MGA objective function in `d_mga.gms`. 

### Additional changes 

Added a run using the random vector method to `cases_test.csv`.

- `GSw_MGA_RV_runs`: number of random weight vectors to draw; corresponds to the number of ReEDS runs.
- GSw_MGA_RV_region`: level of regionality at which to sample the random weights. 

The random vector method has been used in previous MGA studies (examples [here](https://www.sciencedirect.com/science/article/pii/S0360544217304097?via%3Dihub) and [here](https://iopscience.iop.org/article/10.1088/2753-3751/ad7d10)

No difference on default cases: [results-main,update.pptx](https://github.com/user-attachments/files/30905597/results-main.update.pptx).

### Details to double-check

  - [ ] Units are specified
  - [ ] If new parameters are added to `d_objective.gms`, they are included in `objective_function_params.yaml` for completeness checking


reeds/inputs.py
```diff
@@ -263,7 +263,7 @@ def parse_cases(
                 ):
                     print(f"\n[Warning] Case Column: '{c}'")
                     print(
-                        "You are attempting to run a Monte Carlo simulation with "
+                        "You are attempting to run a Monte Carlo or MGA Random Vector simulation with "
```
bsergi: This probably doesn't matter since the MGA runs should all have the same inputs, but I left it in for now in case there's a non-deterministic component to the representative day sampling.

reeds/input_processing/runfiles.csv
```diff
@@ -132,7 +132,7 @@ i_p.csv,inputs/sets/i_p.csv,1,ignore,ignore,,,,,0,,,set,i_p,mapping from technol
 i_water_nocooling.csv,inputs/sets/i_water_nocooling.csv,1,ignore,ignore,,,,,0,,,set,i_water_nocooling,technologies that use water but are not differentiated by cooling tech and water source,
 inflation.csv,inputs/financials/inflation_{inflation_suffix}.csv,1,ignore,ignore,,,,0,0,,,,,,
 interconnection_queues.csv,inputs/capacity_exogenous/interconnection_queues.csv,1,ignore,ignore,r,"tg,r",,1,0,,,,,,
-jtype.csv,inputs/sets/jtype.csv,1,ignore,ignore,,,,,,,,set,jtype,job types used in model (construction and om),
+jtype.csv,inputs/sets/jtype.csv,1,ignore,ignore,,,,,0,,,set,jtype,job types used in model (construction and om),
```
patrickbrown4: `jtype.csv` doesn't have a header so I think the changes to this file should be reverted

reeds/inputs.py
patrickbrown4: If the inputs are unchanged then the rep days will be unchanged, so I'm pretty sure we can remove the check for RV and only keep it for MCS, which makes it a little easier to use.

reeds/inputs.py
```diff
@@ -283,23 +283,29 @@ def parse_cases(
...
-                # c (column) is a case with monte carlo runs.
-                # replicate this column N (NumMonteCarloRuns) times
-                NumMonteCarloRuns = int(dfcases.loc['MCS_runs',c])
+                if int(dfcases[c].MCS_runs) > 0:
+                    mcs_sampler.general_mcs_dist_validation(reeds.io.reeds_path, mcs_dist_path, sw)
+                    numruns = int(dfcases[c].MCS_runs)
+                    run_type = 'MC'
+                else:                    
+                    numruns = int(dfcases[c].GSw_MGA_RV_runs)
+                    run_type = 'R'
```
patrickbrown4, line 296: MCS and RV are mutually incompatible, right (at least in the current implementation, which is totally ok)? If so, should probably add a check to `runreeds.check_compatibility()` that raises an exception if `MCS_runs` and `GSw_MGA_RV_runs` are both nonzero.

reeds/input_processing/mcs_sampler.py
```diff
@@ -14,6 +14,7 @@
 import pandas as pd
 import scipy.stats
 import sys
+import re
```
patrickbrown4: (minor cleanup) can remove since unused
```suggestion
```

reeds/input_processing/mcs_sampler.py
```diff
@@ -1902,21 +1985,30 @@ def main(
...
+
     # Final log/timing update.
     reeds.log.toc(
         tic=tic, 
         year=0, 
         process='input_processing/mcs_sampler.py',
         path=os.path.join(os.path.dirname(inputs_case))
     )
+
+# %%
```
patrickbrown4:
```suggestion
```

reeds/input_processing/mcs_sampler.py
```diff
@@ -1872,6 +1873,88 @@ def main(
...
+    lhs_sampling: int = 1,
+    seed: int = 0,
+    discrete: bool = True,
+):
+
+    # get dimensions based on number of regions and subojectives
+
+    ## regions
+    # get list of valid regions (val_r generated in copy_files.py)
+    val_r = list(reeds.io.read_input(inputs_case, 'r')['*'])
```
patrickbrown4: (nitpick) very slightly safer (the '*' label is only for GAMS and is kind of arbitrary, so seems a little better to avoid referencing it directly)
```suggestion
    val_r = list(reeds.io.read_input(inputs_case, 'r').squeeze(1))
```

reeds/input_processing/mcs_sampler.py
```diff
@@ -1872,6 +1873,88 @@ def main(
...
+    ## objective (assumes sw.GSw_MGA_Objective in ['capacity', 'generation'] based on 
+    ## check in runreeds.check_compatibility()
+    ## if the subojective is an aggregated category, break it up into smaller groups
+    ## otherwise just use the subobjective as the group
+    mapped_categories ={
+        'gentech': ['coal', 'gas', 'nuclear', 'h2_combustion', 'geo', 'hydro', 'ofswind', 'onswind', 'storage', 'upv'],
+        'fossil' : ['coal', 'gas'],
+        're': ['geo', 'hydro', 'ofswind', 'onswind', 'pv', 'vre', 'wind'],
+        'vre': ['ofswind', 'onswind', 'upv'],
+    }
```
patrickbrown4: Feels a little fragile but I'm not sure it's worth adding a bunch of processing to determine group/subgroup relations automatically from `tech-subset-table.csv`.

Maybe you could at least move this block up to `MCSConstants` so all the hard-coded stuff is defined in one place at the top?

reeds/input_processing/mcs_sampler.py
```diff
@@ -1872,6 +1873,88 @@ def main(
...
+    subsets = mapped_categories.get(sw.GSw_MGA_SubObjective, [sw.GSw_MGA_SubObjective])
+    
+    ## sample weights for each subojective group and sampling region
+    dimensions = len(subsets) * len(sampling_regions)
+
+    # setup output
+    runs_folder_name = os.path.basename(os.path.dirname(inputs_case.rstrip(os.path.sep)))
```
patrickbrown4: (nitpick) a little more readable in my opinion
```suggestion
    runs_folder_name = Path(reeds.io.standardize_case(inputs_case)).name
```

reeds/input_processing/mcs_sampler.py
```diff
@@ -1872,6 +1873,88 @@ def main(
...
+        # set random seed using the global seed + MGA run number to allow reproducibility
+        np.random.seed(seed + mga_run_number)
+        # get the weights for this specific run (-1 to 1 to allow for simultaneous min/max)
+        if discrete:
+            mga_weights_raw = np.random.choice([-1,1], dimensions)
+        else:
+            mga_weights_raw = np.random.uniform(-1, 1, dimensions)
+
+    # save vector of weights for this run, mapped back to r regions and rounded to 6 decimal places
+    mga_weights = pd.DataFrame({sw['GSw_MGA_RV_region']: region_labels, 'i_subtech': subset_labels, 'weight': mga_weights_raw.round(6)})
```
patrickbrown4: (whitespace nitpick)
```suggestion
    mga_weights = pd.DataFrame({
        sw['GSw_MGA_RV_region']: region_labels,
        'i_subtech': subset_labels,
        'weight': mga_weights_raw.round(6)
    })
```

docs/source/user_guide.md
```diff
@@ -631,8 +631,17 @@ Options are the column names in the `inputs/tech-subset-table.csv` file.
 
 Users familiar with GAMS can add alternative objective functions to the `d_mga.gms` file and associated options to the `GSw_MGA_Objective` switch in `cases.csv`.
 
+By default the MGA min/max is applied to the sum of the variable across all regions being modeled.
+The MGA approach also supports an option to randomly sample of a vector of weights to apply to the regional values of the variable being optimized.
+This method can be useful to characterizing the uncertainty in the regional distribution of the results.
```
patrickbrown4:
```suggestion
This method can be used to characterize the uncertainty in the regional distribution of the results.
```

docs/source/user_guide.md
```diff
@@ -631,8 +631,17 @@ Options are the column names in the `inputs/tech-subset-table.csv` file.
...
+Weights are sampled as discrete values from a support of {-1,1} to allow for simultaneous minimization and maximization.
```
patrickbrown4: Just to check, is "support" the right word here? (I'm unfamiliar with this usage but also don't know much stats)

docs/source/user_guide.md
```diff
@@ -631,8 +631,17 @@ Options are the column names in the `inputs/tech-subset-table.csv` file.
...
 
+The MGA random vector capability is controlled by the following switches:
+- `GSw_MGA_RV_runs` (default `0`): Number of random samples of weight vectors to draw; corresponds to the number of runs.
+- `GSw_MGA_RV_region` (default `r`): Regionality level (specified by hierarhcy file) over which to sample the random weights.
```
patrickbrown4:
```suggestion
- `GSw_MGA_RV_region` (default `r`): Regionality level (specified by hierarchy file) over which to sample the random weights. 
```

reeds/core/setup/b_inputs.gms
```diff
@@ -6025,6 +6027,35 @@ employment_factor_plant(i,"construction")
...
+mga_weights(r,i_subtech) = 1 ;
+$endif.mga_rv
+
+$endif.mgaobj
+
+
+*TODO: add different handling for other subobjectives with different dimensions
+* should we also include error checking?
+
+
```
patrickbrown4: Best to move to issue if you still want to address it (or just remove if it's old)
```suggestion
```

reeds/core/setup/d_mga.gms
```diff
@@ -21,7 +21,25 @@ eq_MGA_Objective$Sw_MGA..
...
+Variable MGA_OBJ "--MWh-- Generation of technology to be minimized/maximied" ;
+eq_MGA_Objective$Sw_MGA..
+    MGA_OBJ
+    =e=
+    sum{(i,v,r,h,t)
+        $[tmodel(t)
+        $valgen(i,v,r,t)
+        $%GSw_MGA_SubObjective%(i)],
+        GEN(i,v,r,h,t) * hours(h) 
+        * sum{i_subtech$i_subsets(i,i_subtech), mga_weights(r,i_subtech)}
```
patrickbrown4: Duplicates the next block below; can instead just add `* sum{i_subtech$i_subsets(i,i_subtech), mga_weights(r,i_subtech)}` to the existing `$elseif.mgaobj %GSw_MGA_Objective% == 'generation'` block

patrickbrown4 (approved):
Looks great! The only critical comment is in `d_mga.gms`; all the rest are minor cleanups. Approving now but should still confirm zero change before merging.

reeds/core/setup/b_inputs.gms
atpham88: Would you move these employment factors to copy_files.py as we discussed in my PR? I'm happy to help do this here as well if needed.

cases.csv
```diff
@@ -225,6 +225,8 @@ GSw_MGA_CostDelta,MGA: Fraction by which to allow objective function to increase
 GSw_MGA_Direction,MGA: Directionality of second optimization,(min|max),min,
 GSw_MGA_Objective,MGA: Objective for MGA (uses GSw_MGA_SubObjective to specify technology subset if set to capacity),(capacity|generation|transmission|rasharing|co2|employment),capacity,
 GSw_MGA_SubObjective,MGA: Technology subset to minimize or maximize the capacity or generation of (only used for GSw_MGA_Objective=(capacity or generation)),(battery|ccs|coal|dac|fossil|gas|gentech|geo|h2_combustion|hydro|nuclear|ofswind|onswind|pv|re|storage|upv|vre|wind),gentech,
+GSw_MGA_RV_runs,"MGA: Number of samples drawn for the random vector weights that are applied to variables specified in GSw_MGA_Objective and GSw_MGA_SubObjective. Only active when GSw_MGA_CostDelta != 0; set to an integer N>0 to run N ReEDS simulations with sampling (0 disables sampling).",int,0,
```
atpham88, line 227: We use `GSw_MGA_RV_runs` for number of MGA RV runs but `MCS_runs` for number of MC runs. Should it be just `MGA_RV_runs` for consistency?

cases.csv
atpham88: Can you add in the description that these sampling settings also apply to sample MGA RV coefficients?

docs/source/user_guide.md
atpham88: Can we just say "weights are sampled as discrete values between {-1,1}"? Also we still have the option to sample linearly also right?

reeds/input_processing/runfiles.csv
bsergi: ah ok, I didn't see the `0 if file has column labels` comment in the second row and had thought 0 meant no header.

reeds/inputs.py
bsergi: Yea that's right. We talked about eventually enabling both, but that might warrant a different approach to the run labeling that I didn't tackle. I'll add a compability check.

docs/source/user_guide.md
bsergi: Support just means where the distribution has non-zero density. It's probably not that useful to include here so I'll put something simpler to avoid the jargon (An, I think your suggestion works).

On the linear weights, the python function supports it but there's no switch to control this so it isn't exposed to the user. It's not that much additional code so I figured we could leave it in and if we change our minds about using it in the future we could either add a switch or just toggle the function argument default for a specific set of runs.

reeds/core/setup/d_mga.gms
bsergi: oh weird, must have been duplicated in a merge with main. thanks for flagging.

reeds/core/setup/b_inputs.gms
bsergi: ah I forgot I said I was going to do that in this one--thanks for the reminder!

cases.csv
bsergi: I debated this one a bit since this does eventually get passed to GAMS and a lot of the MGA switches have this prefix, but I think changing it to `MGA_RV_runs` might be cleaner.

patrickbrown4: FWIW my vote would be to keep all related switches together (which for me at least makes it easier to remember how they interact), so either all `MGA` or all `GSw_MGA`

bsergi: The rest of the MGA switches have the `GSw` so I'll leave as is.

reeds/input_processing/mcs_sampler.py
bsergi: I took a pass at a function to try this (was a welcome distraction). Might be a bit overkill so we can discuss whether it's worth keeping. Initial tests of just the mcs_sampler.py return the subcategories I was hoping for but going to test in a full run now.

PR #170, opened 2026-08-05
Update choropleth maps

clairehalloran:
## Summary

Adds [adjustText ](https://github.com/Phlya/adjustText/tree/master) package to environment as an optional package to avoid label overlap in maps and adds regional labels to some maps.

Before:
image

After:
image
## Technical details 
Added labels to net import, technology capacity, and H2 capacity maps in `single_case_plots.py`:

image

image

Added labels to difference maps in `compare_cases.py` for both 2-case and multi-case comparison:
image

image

In `hourly_plots.py`, NaN values (such as for land-locked regions' offshore wind cf) are now colored light gray and unlabeled:
image

### Implementation notes 
I incorporated the `adjust_text` function into the existing `reedsplots.label_region_value` function for easy reuse.

### Additional changes 
- Changes absolute color map from cmocean rain to tempo
image

- Added outlines to sub-state regions in `plot_diff_maps` used for 2-case comparison in `compare_cases.py`
- Squashed a bug that didn't simplify techs to their maptechs groups in `plot_diff_maps` function in `reedsplots.py`
- Squashed small bugs in `diagnostic_plots.py`
  - Removed slashes from savenames that raised errors ("Oil/Gas Steam" tech)
  - Re-assigned dropped CRS

I compared USA_defaults run on this branch with a run from main. There were no changes in results, and a negligible increase in runtime (0.15 hours). Note that the difference in directory size is because I was saving vre_gen, max_cap, and load files from PRAS to run `diagnostic_plots.py`.

Full results here: [results-main,update-maps.pptx](https://github.com/user-attachments/files/31431403/results-main.update-maps.pptx)

### Details to double-check

~~- [ ] Documentation updated if necessary~~
  
- [ ] No change to package requirements (environment.yml or Project.toml): adjustText package added

Yes, I used Copilot to simplify techs to their maptechs groups in `plot_diff_maps` function in `reedsplots.py`. I attempted this change manually and couldn't find a more elegant solution.


environment.yml
```diff
@@ -25,6 +25,7 @@ dependencies:
   - scipy=1.18
   - shapely=2.1
   - tqdm=4.68
+  - adjustText
```
patrickbrown4, line 28: With the prevalence of LLM code I'm more wary of introducing dependencies on small / less-well-vetted packages; there are examples of them getting taken over and hosting malicious code. (That's not unique to LLM code but it is easier to do now, though I might be overreacting.) So I'm on the fence about introducing a new dependency for something that's not required for the model.

* How hard do you think it would be to implement this functionality ourselves by tweaking `reeds.plots.optimize_label_positions()`?
* Could this be an optional package (like `seaborn` or `scikit-learn-extra`; you can see examples of how we handle those [here](https://github.com/ReEDS-Model/ReEDS/blob/4fb93ed5cf3560be0ead839ab75d272d24f89c1e/postprocessing/uncertainty_plots.py#L26-L29) and [here](https://github.com/ReEDS-Model/ReEDS/blob/4fb93ed5cf3560be0ead839ab75d272d24f89c1e/runreeds.py#L538))? i.e., only used and imported within a single function, and bypassed if it isn't present?
* If we keep it as is, we should at least pin the version number installed.

reeds/plots.py
patrickbrown4, line 1: What's the motivation behind the change from `cmocean.cm.rain` to `cmocean.cm.tempo`? I've been using `cmocean.cm.rain` since it has more of a color range while still being perceptually uniform, but it might just be what I'm used to.

In general, if we keep the change, I would suggest applying it via the existing global arguments at the top of `compare_cases.py` and `single_case_plots.py` (and making sure they're supplied to the called functions) rather than changing the argument defaults throughout the function definitions. Then if different users prefer different colormaps, they only need to change it in one place, and we avoid some churn in the code itself.

reeds/reedsplots.py
```diff
@@ -981,59 +982,73 @@ def plot_diff_maps(
     dfbase[valcol] *= unitscaler
     dfcomp[valcol] *= unitscaler
 
+    ### Aggregate selected technologies to one value per region for the target year
```
patrickbrown4, line 985: Could you say more about what these changes do? I've also been working on a rewrite of this function on the [pb/mapdiff](https://github.com/ReEDS-Model/ReEDS/tree/pb/mapdiff) branch but I should have opened an issue to clarify.

reeds/resource_adequacy/diagnostic_plots.py
```diff
@@ -987,7 +1000,7 @@ def map_pras_failure_rate(sw, dfs, aggfunc='mean', repair=False):
     plottechs = failsum.loc[failsum != 0].index.get_level_values('i').unique()
 
     for tech in plottechs:
-        savename = f"hourly_failure_rate-year,month-{aggfunc}-{tech.replace('-','')}-{sw['t']}"
+        savename = f"hourly_failure_rate-year,month-{aggfunc}-{tech.replace('-','').replace('/','')}-{sw['t']}"
```
patrickbrown4, line 1003: Nice, thanks

environment.yml
```diff
@@ -1,4 +1,4 @@
-name: reeds
+name: reeds-at
```
patrickbrown4, line 1:
```suggestion
name: reeds
```

patrickbrown4 (commented):
Thanks Claire and sorry for the slow followup. Some high-level questions first and then I'll take another closer look.

reeds/reedsplots.py
clairehalloran: The current behavior plots all technologies in the `titles` list without combining into a single `i_plot` value (so 'DPV' and 'UPV 'capacity are plotted on top of each other on the same map, rather than being combined into a single 'Solar' capacity for each region).

These changes combine all technologies in the value list `titles` into a single total capacity for the `i_plot` category. This sum occurs prior to calculating differences so a single `i_plot` difference between cases can be calculated. This change fixes a bug that made the absolute and difference plots misleading.

The difference calculation now calculates both absolute and percent differences so that there's only one conditional like this:
```
    elif plot in ['absdiff', 'abs_diff', 'diffabs', 'diff_abs']:
```
This change is less important.

reeds/plots.py
clairehalloran: I find it confusing that `cmocean.cm.rain` goes to white for 0% to brown for 20%, but I guess that's a matter of preference. 

Either way, happy to point to global arguments to keep the code simpler.

environment.yml
clairehalloran: I lean towards either the first or third bullet. I think labeling maps without overlap is a helpful feature for many existing figures, so it would be harder to use as an optional package.

I'll look further into how `adjustText` and `reeds.plots.optimize_label_positions` are implemented to see if it would be easy to tweak the latter function.

clairehalloran: We discussed this a bit yesterday, but I'm confirming that `adjustText` uses a pretty different algorithm than the LP in `reeds.plots.optimize_label_positions`. Per the docstring:

> First "explodes" all texts to move them apart. 
    Then in each iteration pushes all texts away from each other, and any specified
    points or objects. At the same time slowly tries to pull the texts closer to their
    original locations that they label (this reduces chances that a text ends up super
    far away). In the end adds arrows connecting the texts to the respective points.

I'll work on making this package optional as we discussed.

PR #171, opened 2026-08-05
Plots: Add a capacity-credit-like plot for stress periods

patrickbrown4:
## Summary

Adds a capacity-credit-like plot, `reedsplots.plot_stress_cf()`, for use with stress periods. Credit to @jennielu92 for the original design.

### Implementation notes 

The new function takes the same `metric` argument used in `plot_stress_mix()` and `get_cap_rep_stress_mix()`. By default, we make 4 versions in `single_case_plots.py`, but the user can make others by running `reedsplots.plot_stress_cf()` directly:
- stress_top10_price
- stress_top10_netload
- stress_top10_load
- stress_bottom10_vregen

The premade "stress_top10_price" figure is included when running `compare_cases.py`.

### Additional changes 

- `input_plots.py`: Add `label` option to `plot_profile()`
- `compare_cases.py`:
  - Add runtime to the hodgepodge plot
  - Add a missing `try/except` around a use of `plots.optimize_label_positions()`
  - Tweak the membership of the non-default `maptechs` for faster processing (has to be explicitly requested by the user; not used by default)
- `plots.py`: Increase default figure resolution (only affects interactive plots)

Here are examples of the new plot for USA_defaults:

image

image

image

### Details to double-check

No


postprocessing/compare_cases.py
```diff
@@ -160,14 +160,11 @@
...
     maptechs = {
         'Battery/PSH': ['Battery', 'Pumped-Hydro', 'Pumped-Hydro-Flex'],
-        'Biopower': ['Biopower', 'BECCS'],
         'Coal': ['Coal'],
-        'Demand Response': ['DR Shed','EVMC_Shape','EVMC_Storage'],
         'Fossil+CCS': ['Coal-CCS-Flex', 'Coal-CCS', 'Coal-CCS_Upgrade','Gas-CCS',
             'Gas-CC-CCS', 'Gas-CC-CCS-Flex', 'Gas-CC-CCS_Upgrade'],
-        'Geothermal': ['Geothermal'],
         'H2 Turbine': ['H2-CC_Upgrade', 'H2-CC', 'H2-CT_Upgrade', 'H2-CT', 'Hydrogen'],
-        'Natural Gas': ['Gas-CC','Gas-CT','Oil/Gas Steam'],
```
wesleyjcole, line 170: Why pull OGS out?  OGS are primarily (though not exclusively) natural gas boilers.

postprocessing/compare_cases.py
```diff
@@ -160,14 +160,11 @@
 if args.map:
     maptechs = {
         'Battery/PSH': ['Battery', 'Pumped-Hydro', 'Pumped-Hydro-Flex'],
-        'Biopower': ['Biopower', 'BECCS'],
         'Coal': ['Coal'],
-        'Demand Response': ['DR Shed','EVMC_Shape','EVMC_Storage'],
         'Fossil+CCS': ['Coal-CCS-Flex', 'Coal-CCS', 'Coal-CCS_Upgrade','Gas-CCS',
             'Gas-CC-CCS', 'Gas-CC-CCS-Flex', 'Gas-CC-CCS_Upgrade'],
-        'Geothermal': ['Geothermal'],
```
wesleyjcole, line 168: I think it's best to leave geothermal from a tech neutrality perspective.

reeds/reedsplots.py
```diff
@@ -5302,6 +5302,14 @@ def separate_charge_discharge(df):
     df.loc[df.i.isin(storage_techs), 'i'] += '|discharge'
 
 
+def check_metric(metric):
```
wesleyjcole, line 5305: Thanks for adding this.

wesleyjcole (commented):
What do you think about changing the top X price metric to be a weighted price metric?  That way any stress period with a positive price would be counted, and in the case that less than 10 stress periods have a positive price, it would weight them accordingly rather than include zero-priced hours in the calculation?  That seems like it might give the best estimate of a marginal capacity credit.

postprocessing/compare_cases.py
patrickbrown4: This is a non-default option (used by passing the `--map/-z` command line argument) to reduce the number of diff maps and speed up the script when doing diagnostics on lots of runs. My intent was to keep it to technologies that tend to vary more between runs. But it's tangential to the rest of the PR so I'll just revert.

patrickbrown4:
> What do you think about changing the top X price metric to be a weighted price metric? That way any stress period with a positive price would be counted, and in the case that less than 10 stress periods have a positive price, it would weight them accordingly rather than include zero-priced hours in the calculation? That seems like it might give the best estimate of a marginal capacity credit.

Thanks, some thoughts:
- Because investment-driving prices can spike so high in individual hours, I worry that the results would be more dominated by single hours (even when specifying X high-price hours) and thus noisier from run to run
  - (Since we include opex prices in stress periods, I've seen few instances of zero prices in current default runs, but they might come up in other runs)
- I'm mostly trying to reproduce the way we've calculated capacity credit in the past, which is to look at the top X hours together. Weighting by load/netload/price would be a bigger rewrite of the existing stress metrics and feels more experimental to me. No problem trying it but my preference would be to add them as additional options (rather than replacing the existing options) and to do it in a separate PR since it affects multiple plots and would require more testing.

If you think "top 10 price hours" is misleading, my inclination would be one of the following:
1. Switch to "top 10 net load" or "top 10 load" for the version included in `compare_cases.py`
2. Don't include any of them in `compare_cases.py`; instead, force people to look at all the metrics (in the individual run folders) and decide which, if any, provides value for their application
    1. And/or include more values of X (1, 5, 20) in the versions saved to the run folder

wesleyjcole:
> > What do you think about changing the top X price metric to be a weighted price metric? That way any stress period with a positive price would be counted, and in the case that less than 10 stress periods have a positive price, it would weight them accordingly rather than include zero-priced hours in the calculation? That seems like it might give the best estimate of a marginal capacity credit.
> 
> Thanks, some thoughts:
> 
> * Because investment-driving prices can spike so high in individual hours, I worry that the results would be more dominated by single hours (even when specifying X high-price hours) and thus noisier from run to run
>   
>   * (Since we include opex prices in stress periods, I've seen few instances of zero prices in current default runs, but they might come up in other runs)
> * I'm mostly trying to reproduce the way we've calculated capacity credit in the past, which is to look at the top X hours together. Weighting by load/netload/price would be a bigger rewrite of the existing stress metrics and feels more experimental to me. No problem trying it but my preference would be to add them as additional options (rather than replacing the existing options) and to do it in a separate PR since it affects multiple plots and would require more testing.
> 
> If you think "top 10 price hours" is misleading, my inclination would be one of the following:
> 
> 1. Switch to "top 10 net load" or "top 10 load" for the version included in `compare_cases.py`
> 2. Don't include any of them in `compare_cases.py`; instead, force people to look at all the metrics (in the individual run folders) and decide which, if any, provides value for their application
>    
>    1. And/or include more values of X (1, 5, 20) in the versions saved to the run folder

Thanks for pointing out that we are unlikely to have zero price hours.  I just checked one of my runs and see that's the case, so I think we can ignore that for right now.

For simplicity, I'm on board with the top 10 hour approach, and I think price is the best indicator of capacity credit.  That seems like a good starting point that we can adjust some in the future.

I've been doing some weighted average approaches for estimating the value of storage value, and I think the weighted values are pretty good, but that can wait for a separate pull request and can be evaluated indepently.  Here's an example of stress period prices from a recent run for the CA_LA region in 2050.  The % number in each plot shows the weight that day would carry for the year (e.g., 32% means 32% of the weighted price happens on that day).  It can be reasonably well distributed, and have more than 10 hours.

CA_LA_2050

Here's another region (AL in 2050) where there are really only 7 hours with meaningful prices, so the extra three hours will likely degrade the estimate.

AL_2050

wesleyjcole (approved):

Issue #154, opened 2026-07-22
Try applying stress-period PRM via operating reserves instead of load


patrickbrown4:
### Summary

We currently implement the PRM by scaling up the stress-period load: https://github.com/ReEDS-Model/ReEDS/blob/915def187fd6c9f24a53eb7af98d797407611f3c/reeds/core/solve/2_temporal_params.gms#L774-L775

That's ok if the stress periods have negligible weight (we currently give them a weight of 6 hours / 8766 hours per year, or 0.07% of the total): https://github.com/ReEDS-Model/ReEDS/blob/915def187fd6c9f24a53eb7af98d797407611f3c/reeds/input_processing/hourly_writetimeseries.py#L623-L625

But if we increase their weight, we'd increase the overcounting of the PRM as actual load.

Reasons to consider increasing the stress period weight:
- Better capture frequency of peak- and near-peak conditions, which might allow us to get rid of the min-CF constraint
- Model energy constraints for LDES (H2) during stress periods, as on [pb/7year](https://github.com/ReEDS-Model/ReEDS/tree/pb/7year)
- Potentially help incorporate peaker operations into state policy considerations

### Proposed Changes

It seems like most of the functionality we need for an hourly PRM is already handled by the existing `OPRES` constraints (which are turned off by default). A rough outline of repurposing these for the PRM would be:
- [ ] Add `prm` as a new `ortype`
- [ ] Move the PRM from a load multiplier into the `orperc` parameter
- [ ] Probably separate control of the PRM functionality from the other operating reserve switches
- [ ] Double check all the transmission flow constraints, since I've made changes to these over the years without thinking about operating reserves
- [ ] Test to make sure the runtime impact isn't too bad. If it is, figure out how to simplify.
  - [ ] Can we aggregate `OPRES(ortype,i,v,r,h,t)` over `v` or `i`?
    - [ ] If that breaks other operating reserve functionality, can we make a similar `PRM(r,h,t)` variable?

PR #172, opened 2026-08-05
Switch to 5-year steps by default

patrickbrown4:
## Summary

This PR changes the default year step from 3 years to 5 years.

Reasons in favor:
- Reduces solve time for the USA_defaults case from 13 hours to ≤8 hours
  - Alternatively, we could instead increase the time resolution (more representative days to help capture near-peak conditions; finer time chunks for better storage representation; perhaps #154).
- We used 5-year steps in NTP; in that case, stakeholders often asked for results for the divisible-by-5 years, so modeling them directly makes that easier
- Minimal impact on the 2035 and 2050 results
- Also reduces peak GAMS memory use, but probably doesn't affect PRAS memory use (though we do run PRAS fewer times).

Reasons against:
- For scenarios where policies change substantially from year to year, can gloss over shorter-timescale effects
  - But since we don't have inter-year foresight anyway, it's arguable that longer timesteps could avoid some myopic decisions

### Implementation notes 

- I kept the old default (3-year steps from 2020–2050) for the Mid_Case scenario in `cases_test.csv` so we can make sure it stays working.

- `yearset`: Changed default to `2010..2050..5` (5-year steps)

Here's a compare report for USA_defaults: [results-3-yearsteps,5-yearsteps.pptx](https://github.com/user-attachments/files/30765130/results-3-yearsteps.5-yearsteps.pptx)

image

image

I think the changes to NPV are because we forward-fill non-modeled years instead of linearly interpolating. If we want to regularly compare NPVs for cases with different year steps, we might consider changing that. But by default it shouldn't cause a problem as long as cases with the same year steps are compared (which seems safest to me anyway).

### 20260827

Updated compare report: [results-y3,y5.pptx](https://github.com/user-attachments/files/31520636/results-y3.y5.pptx)

Changes are still small (aside from runtime, which is down to 6.3 hours):
image

### Details to double-check

- [x] Documentation updated if necessary (double checked and docs do not specify the default year step)

- [ ] No substantive impact on runtime for full-US reference case (13h -> 8h)
- [ ] No substantive impact on folder size for full-US reference case (4.9 GB -> 3.4 GB)

Nope


cases_test.csv
```diff
@@ -1,8 +1,8 @@
 ,Default Value,Pacific,USA_defaults,Mid_Case,USA_decarb,github_Pacific,github_Everything,github_MA_county_CC,Pacific_CC,Pacific_weks,Pacific_full_year,Interday_storage,Pacific_2020,Pacific_rep15,WY_county,WECC_county,PJM_county_CC,NYVT_mixed,OR_water,MonteCarlo_Random,MonteCarlo_LHS,Everything,Simple,USA_fast,USA_faster,MultiMetricRA,Pacific_DR,Pacific_MGA,Pacific_MGA_RV,Pacific_LoadSite95,MARICTNYNJPAOH_Offshore,R2P
 ignore,1,0,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
 GSw_Region,cendiv/Pacific,,country/USA,country/USA,country/USA,,st/ID.WY.NE.IA.IL,st/MA,,,,,,,st/WY,interconnect/western,transreg/PJM,st/NY.VT,st/OR,st/NE.NY.PA,st/NE.NY.PA,st/ID.WY.NE.IA.IL,st/KS,country/USA,country/USA,st/NY.NJ,,,,,st/MA.RI.CT.NY.NJ.PA.OH,
-endyear,2032,,2050,2050,2050,2029,2060,2026,,,,,,,,,,,2035,2030,2030,2060,2035,2050,2050,2050,,,,,,
-yearset,,,,,,,2010..2060..10,,,,,,,,,,,,,2010..2050..5,2010..2050..5,2010..2060..10,,2010..2050..5,2010_2025_2050,2010..2050..5,,,,,,
+endyear,2035,,2050,2050,2050,2030,2060,2030,,,,,,,,,,,2035,2030,2030,2060,2035,2050,2050,2050,,,,,,
+yearset,,,,2010_2015_2020..2050..3,,,2010..2060..10,,,,,,,,,,,,,,,2010..2060..10,,,2010_2025_2050,,,,,,,
```
bsergi: We were already planning to switch to 5-year steps for this year's Standard Scenarios. I thinking keeping this Mid-Case at 3-year for now make sense but if we do update this with the 2026 Mid-Case settings we might want to add a separate case somewhere that tests the 3-year steps.

cases_examples.csv
bsergi, line 1: there are several `endyear` values that are still 2032 in this file

.github/workflows/python-app.yaml
```diff
@@ -257,11 +257,11 @@ jobs:
       matrix:
         include:
           - scenario: github_Pacific
-            solve-year: 2029
+            solve-year: 2030
```
bsergi, line 260: I would think you need to make a similar change to the solve-year values in the `run-ReEDS` block further up, but your checks passed so I'm not certain if it matters.

bsergi (approved):
A few small things to check but otherwise no major concerns from my end. I think the runtime gains and ease of comparison to other studies are worth the change.

For Standard Scenarios we were also planning to update `this_year` to 2026. Combining that with this update would make the first endogenous year 2030. Not sure that's an issue but just flagging.

cases_test.csv
patrickbrown4: Ok good to know, in that case I'm happy moving it to a different test case (this one is already pretty long anyway) - I'll put it in `Simple`

cases_examples.csv
patrickbrown4: Good catch, thanks

patrickbrown4:
> For Standard Scenarios we were also planning to update `this_year` to 2026. Combining that with this update would make the first endogenous year 2030. Not sure that's an issue but just flagging.

@bsergi Good point. `this_year` is only used to convert `years_until_endogenous.csv` to `firstyear` (and it seems like we should go ahead and update that now), but `GSw_StartMarkets` is already 2026. Do you think we should add 2026 as a solve year? So `yearset = 2010..2025..5_2026_2030..2050..5` (or refactor it to always include `GSw_StartMarkets` in the parsed year set)? I can try it and see how big a difference it makes.

bsergi:
> @bsergi Good point. `this_year` is only used to convert `years_until_endogenous.csv` to `firstyear` (and it seems like we should go ahead and update that now), but `GSw_StartMarkets` is already 2026. Do you think we should add 2026 as a solve year? So `yearset = 2010..2025..5_2026_2030..2050..5` (or refactor it to always include `GSw_StartMarkets` in the parsed year set)? I can try it and see how big a difference it makes.

For Standard Scenarios we were leaning toward not including 2026 and just starting with 2030. If you do test I'd be interested in the outcome, but I suspect this isn't too big of an issue.

patrickbrown4:
> For Standard Scenarios we were leaning toward not including 2026 and just starting with 2030. If you do test I'd be interested in the outcome, but I suspect this isn't too big of an issue.

Here are the results when 2026 or 2027 are included: [results-2025-2030,2026,2027.pptx](https://github.com/user-attachments/files/31560750/results-2025-2030.2026.2027.pptx). Including them adds some near-term squiggles but doesn't change the long-term behavior. And in my mind it opens up more questions about near-term effects that the model is not great at handling. So I'm also happy leaving it out by default. But folks can of course use whatever yearset they want for their studies.

PR #176, opened 2026-08-10
Adjust battery lifetime to 30 years (with 15-year refurbishment)

### LINKED ISSUE #119: Update Battery Plant Lifetime to 30 years (with 15-year replacement)
### Summary
Batteries have an assumed 15-year lifetime, and are assumed to be fully retired at the end of 15 years.  However, many aspects of a battery system have lifetimes greater than 15 years, and many batteries are coupled with generation facilities that will have lifetimes greater than 15 years (nearly half of the storage [in the queue](https://emp.lbl.gov/queues) is in hybrids).  We think it would be more appropriate to model batteries as a 30-year asset with some refurbishment cost in year 15 rather than as a 15-year asset.  PV+battery hybrids in the Annual Technology Baseline (ATB) already use this approach.

### Proposed Changes
Vignesh Ramasamy did some bottom-up calculations to estimate that a refurbished battery would cost 71-83% as much as a new build (where the range depends on if 80% or 100% of the battery packs are replaced in year 15).  He sent me his bottom-up modeling spreadsheet we can use to get specific values as needed.

We would want the refurbishment cost to take into account the cost of storage in a future year, so I think the best approach would be:

[ ] Take a close look at Vignesh's work and any battery replacement literature to choose a refurbishment cost and add that refurbishment cost number (e.g., 75% of a new build cost) to `scalars.csv`
[ ] Decide whether to add the refurbishment cost as an upfront capex cost or as a FOM cost (ATB uses the FOM approach for hybrids)
[ ] Perform the refurbishment cost calculation in `plantcostprep.py`. It will need to read in some financial parameters to do discounting properly, but I think it still fits better in `plantcostprep.py` than in `calc_financial_inputs.py`.
[ ] Ensure that the refurbishment cost flows into the objective function properly
[ ] Change the economic and physical lifetime of batteries to 30 years

### Additional context
Depending on whether the FOM or capex approach is taken, it might be worth thinking about how this refurbishment cost shows up the system cost and retail rate calculations.

wesleyjcole:
## Summary
This pull request changes the default battery lifetime to 30 years, with a full refurbishment of the energy portion of the system still occuring at 15 years.  We do this because storage is increasingly paired with longer-lifetime assests (nearly half of the storage [in the queue](https://emp.lbl.gov/queues) is a hybrid), and to better represent the longer lifetime of many aspects of the battery system (e.g., permits, interconnection agreement, site prep, etc.).

## Technical details 
This PR changes both the physical lifetime and the book lifetime of batteries to 30 years.  It calculates a discounted energy refurbishment cost for year 15, and adds that to the energy cost of the storage.  That discounted cost uses the projected energy costs for 15 years out when doing the calculation.  The battery refurbishment year and the amount of the energy portion of the battery that needs refurbished are both added to `scalars.csv` to make them easy to adjust if needed.

The overall result is that the battery energy capital cost is 30-45% higher, but `cost_cap_fin_mult` for the entire battery system is 34-42% lower.  This reduces the overall lifetime cost of the battery.  For example, for the AL region in 2032, the overall battery cost is 12% lower with a 30-year lifetime (and 15-year refurbishment) than with a 15 year lifetime.

### Issues resolved 
This PR closes #119.

Vignesh Ramasamy did some bottom-up calculations to estimate that a refurbished battery would cost 71-83% as much as a new build (where the range depends on if 80% or 100% of the battery packs are replaced in year 15). The bottom-up model he used was the same one used for the [2025 battery costs](https://docs.nlr.gov/docs/fy25osti/93281.pdf).  By assuming full replacement of the energy portion of the battery, we see similary costs levels (ATB 2024 moderate case plotted here):

image

The longer battery lifetime results in more solar and battery capacity, which trades off with gas and wind.

image

System costs are reduced by 0.9%.

Full compare report is here: [results-Main,BattAdjust.pptx](https://github.com/user-attachments/files/30914040/results-Main.BattAdjust.pptx)

And the inputs.gdx diff is here: [diffile.zip](https://github.com/user-attachments/files/30914050/diffile.zip)

I manually spot-checked the output calculations from Python to verify they are correct.

### Details to double-check

Yes.  I used Claude to make the code changes implemented here, with editing and review throughout.  I re-wrote the comments (they were too verbose), and made the documentation updates without LLM use.


reeds/input_processing/plantcostprep.py
```diff
@@ -219,6 +219,41 @@ def deflate_func(data,case):
 battery = pd.read_csv(os.path.join(inputs_case,'plantchar_battery.csv'))
 battery = deflate_func(battery, sw.plantchar_battery)
 
+### Battery Mid-Life Refurbishment Cost
+# battery_li operates for 30 years, but its energy capacity (e.g., battery packs)
```
patrickbrown4: I think it's best to avoid hardcoding assumptions, even in comments
```suggestion
# Battery plant lifetime is specified in `inputs/plant_characteristics/maxage.csv`,
# but its energy capacity (e.g., battery packs)
```

reeds/input_processing/plantcostprep.py
```diff
@@ -219,6 +219,41 @@ def deflate_func(data,case):
...
+# must be replaced after batt_years_until_refurb years. The power capacity lasts
+# the full lifetime and is not refurbished. The energy refurbishment is represented
+# as a discounted adder to the overnight energy capital cost, valued at the cost
+# of new energy capacity in the refurbishment year.
+
+years_until_refurb = int(scalars['batt_years_until_refurb'])
+
+# real discount rate by build year
+financials_sys = pd.read_csv(os.path.join(inputs_case,'financials_sys_full.csv'))
```
patrickbrown4, line 233: Should we use the system discount rate or tech-specific discount rate (via `financing_risk_mult` or `finance_diff_real`) here?

It feels like a tech-specific discount would be more appropriate, but we apply `financing_risk_mult` downstream in `2_financials.gms` and don't want to double-count, so it's not immediately obvious (to me at least) given the different horizons involved.

reeds/input_processing/plantcostprep.py
```diff
@@ -219,6 +219,41 @@ def deflate_func(data,case):
...
+
+# cost of new energy capacity in the refurbishment year, holding the
+# last year of the cost projection constant beyond the end of the input data
+cost_by_year = battery.set_index('t')['capcost_energy']
+future_cost_batt = (
+    cost_by_year
+    .reindex(range(
+        cost_by_year.index.min(),
+        cost_by_year.index.max() + years_until_refurb + 1))
+    .ffill()
```
patrickbrown4, line 246: The forecast method beyond 2050 is specified in the [`plantcharout.csv`](https://github.com/ReEDS-Model/ReEDS/blob/3a1437a0b521c48780b9957f09e067acb1e89812/inputs/userinput/futurefiles.csv#L343) entry of `futurefiles.csv`. It's set to "constant", so they're consistent, but maybe worth adding a comment here or to the block above pointing to the potential interaction with that file. (I don't think it's worth making it dynamic at this point since we'd like to refactor `forecast.py` anyway.)

reeds/input_processing/plantcostprep.py
```diff
@@ -219,6 +219,41 @@ def deflate_func(data,case):
...
+    .loc[cost_by_year.index + years_until_refurb]
+    .set_axis(cost_by_year.index)
+)
+discount = d_real.reindex(cost_by_year.index) ** years_until_refurb
+
+# Add the discounted refurbishment cost to the overnight energy capital cost
+battery = battery.set_index('t')
+battery['capcost_energy'] = (
+    battery['capcost_energy']
+    + scalars['batt_refurb_perc_energy'] * future_cost_batt / discount
```
patrickbrown4: (nitpick) Seems marginally clearer to me for `discount` to be the discount factor instead of its reciprocal (like [here](https://github.com/ReEDS-Model/ReEDS/blob/3a1437a0b521c48780b9957f09e067acb1e89812/postprocessing/compare_cases.py#L515-L519)) but not a big deal

```suggestion
discount = 1 / d_real.reindex(cost_by_year.index) ** years_until_refurb

# Add the discounted refurbishment cost to the overnight energy capital cost
battery = battery.set_index('t')
battery['capcost_energy'] = (
    battery['capcost_energy']
    + scalars['batt_refurb_perc_energy'] * future_cost_batt * discount
```

patrickbrown4 (approved):
Thanks Wesley. A couple small comments, and I'd be interested in your thoughts on the most appropriate discount rate to use (systemwide or tech-specific), but otherwise looks good.

reeds/input_processing/plantcostprep.py
wesleyjcole: Good point.  I've added a comment.

reeds/input_processing/plantcostprep.py
wesleyjcole: Good question.  I think we want to use the system discount rate because if we were to solve this problem using an intertemporal setup, we would discount the costs in the future years by the system discount rate to caculate an overall present value.  That seems like the same thing we are doing here.  It also seems like using it here, and then having it multiplied by the `financing_risk_mult` in the objective function (via `cost_cap_fin_mult`) would be double-counting.  I'm not 100% sure on this one, but it at least feels reasonable to me to go this route.  Fortunately the difference between the two is small, so we wouldn't get a wildly different answer if we switched.

PR #177, opened 2026-08-11
Allow RPS/CES requirements to be enforced explicitly during stress periods

wesleyjcole:
## Summary
This pull request adds a new switch (`GSw_StateRPS_Stress`) that allows users to enforce the RPS/CES requirements explicitly during the stress periods.  Currently, the RPS/CES requirements are enforced only during the representative periods.  By turning on the switch, the requirements can be enforced during all stress periods (with equal weighting) or during the peak load stress periods.  Turning the switch on represents that a state would plan to be compliant with the RPS/CES requirement for planned stress conditions, not just for representative conditions.

The switch is off by default, so there is no change to default model behavior.

### Implementation notes 
The RPS equations now include a `htype` set, where `htype` is "rep" or "stress" to indicate whether representative or stress periods are being used for the equation.  Representative periods are weighted using `hours(h)` like before, and stress periods are weighted using "1" for any stress period included and "0" for all other stress periods.

A new linking set `st_szn_stresspeak(st,allszn)` is included to specify which stress period is a state's peak day.

Outputs are updated so that when the new switch is turned on, the marginal costs show up in the downstream cost reporting.

`GSw_StateRPS_Stress` is new: 0=off (default, no change), 1=weight all modeled stress hours equally, 2=weight only each state's own peak-load stress period(s).

### Known incompatibilities 
The new switch only works with the stress period representation, and the peak load version of the switch only works when peak load stress days are selected.  This PR includes checks for both of those conditions so an error message will indicate if a user selected an invalid combination.

Enforcing the RPS/CES requirements in the stress periods results in more firm low-emission capacity and less gas capacity:

image

As expected, the states with high CES policies experience the largest changes and build more in-state resources.  You can see that in the compare report deck by flipping through slides 18-20 (maps of the generation and transmission build out).

Comparison report: [results-Main,CESStress1,CESStress2.pptx](https://github.com/user-attachments/files/30953031/results-Main.CESStress1.CESStress2.pptx)

I confirmed that there are no changes with main when the `GSw_StateRPS_Stress` is turned off: [results-Main,CESBaseline.pptx](https://github.com/user-attachments/files/31046577/results-Main.CESBaseline.pptx)

### Details to double-check

Yes, I did the code design, and then implemented that design using Claude.  I edited most of Claude's comments (still haven't figured out how to get the LLM to write comments I like).  The model documentation changes did not use a LLM.


reeds/core/setup/b_inputs.gms
```diff
@@ -2366,14 +2366,14 @@ hydro_capcredit_delta(i,t)$hydro_d(i) =
 *         --- RPS data ---
 *====================================
 
-set RPSCat_i(RPSCat,i,st)     "mapping between rps category and technologies for each state",
-    RecMap(i,RPSCat,st,ast,t) "Mapping set for technologies to RPS categories and indicates if credits can be sent from st to ast",
+set RPSCat_i(RPSCat,i,st)       "mapping between rps category and technologies for each state",
+    RecMap(i,RPSCat,st,ast,htype,t) "Mapping set for technologies to RPS categories and indicates if credits can be sent from st to ast",
```
aschleif, line 2386: Why did the `htype` index have to be added to `RecMap`?

wesleyjcole: It doesn't have to be added.  You'll get the same answer even if it isn't.  Having it allows you to specify different trading requirements between the stress and representative periods, but that isn't a requirement for the method.

Yunzhi-Chen:
Thanks for the PR! I'm wondering whether this is also compatible with user-defined stress periods (i.e., when `GSw_PRM_StressModel` starts with `user`)?

wesleyjcole:
> Thanks for the PR! I'm wondering whether this is also compatible with user-defined stress periods (i.e., when `GSw_PRM_StressModel` starts with `user`)?

No, it isn't compatible, so I'm glad you flagged it.  I'll add a check so that it won't be allowed when the `user` option is selected.

wesleyjcole:
> > Thanks for the PR! I'm wondering whether this is also compatible with user-defined stress periods (i.e., when `GSw_PRM_StressModel` starts with `user`)?
> 
> No, it isn't compatible, so I'm glad you flagged it. I'll add a check so that it won't be allowed when the `user` option is selected.

Correction: it is compatible with `GSw_StateRPS_Stress = 1`, but not with `GSw_StateRPS_Stress = 2` because the `GSw_StateRPS_Stress = 2` option requires peak day identifiers, which are not present in user-defined stress periods.

Issue #158, opened 2026-07-24
follow up PR: update doc to reflect all the change in PR#12


atpham88:
A reminder to myself to update documentation, comments, and clean up after https://github.com/ReEDS-Model/ReEDS/pull/12 is merged in.

Items:
- Clean up comments and documentation mentioning 2010 start year
- Hourlize further clean up
- Update/test option for post-2010 start year
- Resolve other issues related to post-2010 start year (such as wind CF)
- Clean up/revert writesupplycurves.py

PR #183, opened 2026-08-18
2025 Supply Curves Update

bsergi:
## Summary

Updates the offshore wind and enhanced geothermal (EGS) supply curves to their 2025 versions and addresses bugs in hourlize. 

Remaining to-do items:
- [ ] Update offshore wind profiles on Zenodo. 
- [ ] Update documentation reference to 2025 reV supply curve report when available.
- [ ] Copy supply curve data from HPC to nrelnas01 and Yampa

## Technical details 

- Updated the wind-ofs and EGS supply curves to their 2025 versions as provided by the reV team. 
  - Both upv and wind-ons were already using 2025 but were re-run with hourlize to confirm, which resulted in slight changes to the class assignment of some supply curve points but no changes to the total capacity 
- Updated the `rev_paths.csv` to include the latest paths for all technologies.
- Fixed hourlize to work with changes introduced in #12 and did some general maintenance
  - Added `copy_rev_folders.py` with a function to automate copying the reV data using rsync when an hourlize run is launched.
  - Removed the cases.json files; these were unnecessary once we stopped having separate supply curves by spatial resolution. The cases to run is now determined from the `rev_paths.csv` file, with the option to subset via arguments to `run_hourlize.py`
- Updated documentation figures 
  - Added a new script for reproducing these figures to `docs/source/plotting_scripts`; this script relies on the existing `input_plots.map_supplycurves` function with a few small tweaks. 
  - The reV documentation report for the 2025 supply curves should be released soon, at which point we can update the references in the documentation and on Zenodo. 

### Issues resolved 

Addresses hourlize portion of #158 .

Total capacity supply curve capacity before and after (see updated figures in the docs for the maps):

sc_totals_all (1)

All three access cases successfully solve. Changes in capacity vary depending on the case but are generally < 40 GW.

Reference:
image

Open:
image

Limited:
image

Compare cases reports:
- Reference: [results-ref_main,ref_update.pptx](https://github.com/user-attachments/files/31184099/results-ref_main.ref_update.pptx)
- Open: [results-open_main,open_update.pptx](https://github.com/user-attachments/files/31184161/results-open_main.open_update.pptx)
- Limited: [results-lim_main,lim_update.pptx](https://github.com/user-attachments/files/31184164/results-lim_main.lim_update.pptx)

### Details to double-check

  - [ ] Units are specified
  - [ ] If new parameters are added to `d_objective.gms`, they are included in `objective_function_params.yaml` for completeness checking
- [ ] Code formatting standardized 
- [ ] Reusable functions used where possible instead of copy/pasted code

I used copilot to help draft initial versions of the `copy_rev_folders.py` and `supply_curve_plots.py` scripts, which I subsequently modified and tested on my own.


docs/source/model_documentation.md
shashwatsharma24, line 1: We should probably recreate Fig29. for EGS technical potential since the supply curve is updated? Or since reV supply curves for geohydro are forthcoming so we could update it for both technologies together later.

docs/source/plotting_scripts/supply_curve_plots.py
shashwatsharma24, line 1: This is very handy!

hourlize/run_hourlize.py
shashwatsharma24, line 1: Since the parser arguments have been modified we'll need to update README.md - specifically [setup for reV supply curves](https://github.com/ReEDS-Model/ReEDS/tree/main/hourlize#setup-for-rev-supply-curves), [running hourlize](https://github.com/ReEDS-Model/ReEDS/tree/main/hourlize#running-hourlize) sections.

docs/source/model_documentation.md
bsergi: Good point. I took a first pass at recreating the EGS portion and got the following:

supplycurve-egs

The map looks pretty different from what's in the docs, and the total supply curve capacity here (~13 TW) is lower than the ~17 GW of near-field + deep EGS listed in Table 6. This might be because the original map shows the full technical potential whereas this one shows what goes into ReEDS after picking the best resources at each depth. Do you know if that's the case or if something else if going on here?

As for hydrothermal I think we might consider dropping that figure for now since it isn't supported by the model and then adding it back in once we get the updated reV data.

hourlize/run_hourlize.py
bsergi: Updated. I also did a broader cleanup of a lot of the stale config settings. Take a look and let me know what you think here.

docs/source/model_documentation.md
```diff
@@ -884,46 +884,45 @@ The hydrothermal potential included in the base supply curve comprises only iden
...
+|           Class 3  |                        275–300 |              9.3 |                1.3 |           \- |
+|           Class 4  |                        250–275 |              0.7 |                8.3 |           \- |
+|           Class 5  |                        225–250 |              1.1 |                 74 |            1 |
+|           Class 6  |                        200–225 |              2.4 |                320 |          169 |
+|           Class 7  |                        175–200 |              0.2 |                709 |         3509 |
+|           Class 8  |                        150–175 |              2.6 |                996 |         8012 |
+|           Class 9  |                        125–150 |              1.1 |               1268 |          686 |
+|           Class 10 |                          \<125 |              4.7 |                 \- |            8 |
+|              Total |                                |             23.9 |               3377 |        12947 |
 ```
```
patrickbrown4, line 914: I think it'd be ok to remove this table since 1) we don't show similar tables for wind/solar and 2) it's hard to remember to keep it up to date, but your call since you've already updated it.
(Ideally it might be nice to have a jupyter notebook with questions like "What's the technical potential of PV/wind/geothermal" and little code blocks that pull it from the repo and display it, both to illustrate where to find things and to automate the table creation, but that'd be a different PR obviously. (And it might still be hard to remember to run it.))

hourlize/README.md
```diff
@@ -34,55 +31,32 @@ Hourlize is run by a call to `run_hourlize.py`, which assembles information on t
...
+* **Base config** (`config_base.json`): general hourlize settings (shared) as well as specific settings for the resource and load processes
 
-Before running resource hourlize the reV runs should be copied from their original location to the shared folder. Currently reV data used in ReEDS (along with hourlize inputs and outputs) are stored in three places:
+The `run_hourlize.py` process will generate a final config (`config.json`) from the relevant base and tech configs for each run. In the case of duplicated entries across configs hourlize the tech config value takes precendence.
 
-* on the nrelnas01 device at `\\nrelnas01\ReEDS\Supply_Curve_Data`
-* on Eagle at `/shared-projects/reeds/Supply_Curve_Data`
-* on Kestrel at `/kfs2/shared-projects/reeds/Supply_Curve_Data`
+The `srun_template.sh` file is used to govern HPC submission settings. Update with your allocation, email, and any other slurm specifications before submitting jobs. There is also a command line argument to via `run_hourlize.py` for running jobs using the debug partition.
```
patrickbrown4, line 45: It seems like we could drop this copy and just use `reeds/hpc/srun_template.sh` for everything (and avoid having two files with the same name), but your call.

hourlize/README.md
```diff
@@ -211,52 +160,48 @@ This section provides some descriptions and typical values for the settings in t
...
+| inputfiles | Config entries for files to copy into the hourlize input folder | ["reg_map_file", "class_path"] |
+| offshore_meshed | Whether to use meshed offshore transmission connections when matching supply curves to counties | false |
+| process_profiles | Whether to process and export hourly capacity-factor profiles | true |
+| profile_id_col | Unique identifier for supply-curve sites and profiles | 'sc_point_gid' |
+| reg_map_file | Path to the ReEDS region mapping file | '{reeds_path}/inputs/zones/z90/county2zone.csv' |
+| rev_paths_file | Path to the file defining reV supply-curve and profile locations | '{reeds_path}/inputs/supply_curve/rev_paths.csv' |
+| scale_factor | Factor used to scale capacity-factor values when writing hourly profiles | 1e4 |
+| state_abbrev | Path to the state-abbreviation file | '{hourlize_path}/inputs/resource/state_abbrev.csv' |
+| subsetvars | Columns in `rev_paths_file` used to select the appropriate reV path | ['tech', 'access_case'] |
+| subtract_exog | Legacy flag passed to supply-curve output processing; currently has no effect | false |
```
patrickbrown4, line 183: I'd just remove this row (and the argument)

hourlize/resource.py
```diff
@@ -801,6 +826,9 @@ def copy_outputs(
         config = json.load(f, object_pairs_hook=OrderedDict)
     cf = SimpleNamespace(**config)
 
+    #%% copy reV folders
+    copy_rev_folders.main(cf.rev_paths_file, [cf.tech], overwrite=True)
```
patrickbrown4, line 830: I might be misinterpreting but does this mean that every time someone runs hourlize (even just for testing), it will overwrite shared files that R2X depends on? That seems risky to me. Could we make it controlled by a switch that is off by default?

hourlize/run_hourlize.py
```diff
@@ -540,14 +713,40 @@ def setup_load(args):
...
-        print("Unsupported method for hourlize")
-    print(f"Hourlize setup for {args.mode}.py complete\n")
+        remotepath, args.local = get_remote_path(args.local)
+        print(f"\nSetting up hourlize calls to {args.mode}.py")
+        if args.mode == "load":
+            setup_load(args)
+        elif args.mode == "resource":
+            setup_resource(args)
+        else:
+            print("Unsupported method for hourlize")
```
patrickbrown4, line 751: I would print the mode and raise an error
```suggestion
            raise ValueError(f"Unsupported method for hourlize: {args.mode}")
```

patrickbrown4 (commented):
Thanks Brian and sorry for the late followup.

1. I updated `input_diff_plots.py` for the new CF and supply-curve formats; now it works for everything.
    1. Here are the results for reference access: [20260908 - supply curves.pptx](https://nrel.sharepoint.com/:p:/s/ReEDS/IQAme1OPjaJwTpsliCEGt8gjARf9LJNaiRv_MQBB4Du6BxQ?e=BTLWdE). The main thing to know is that reference-access siting for offshore wind removes a bunch of sites. Unclear if it's on the basis of depth or distance from shore (my guess is depth since it's not a uniform distance); we should confirm with the reV team to make sure we understand the reasoning.
        1. Image
1. I tried running the reference case but forgot it won't work until the new offshore wind profiles are uploaded to Zenodo. Let me know if it's helpful to talk through any of that. Once it's done I'll retry the run and then approve.
    1. Let's rerun `input_diff_plots.py` after the offshore wind profiles are uploaded just to double check.
1. I'm pretty sure most of the differences in the compare report come from different representative periods, but it'd be nice to double check if not too much trouble (if helpful, I have the rep periods for a recent main-branch run at `/projects/reedsweto/pbrown/github/ReEDS/inputs/temporal/period_szn_user_v20260911_mainK0_USA_defaults.csv`; can copy these and then set `GSw_HourlyClusterAlgorithm = user_v20260911_mainK0_USA_defaults`).

hourlize/run_hourlize.py
bsergi: This else block is actually dead code since the argument parser will check for supported mode arguments, so I'll remove it.

docs/source/model_documentation.md
bsergi: I'd be in favor deleting. @shashwatsharma24 are you ok with that?

hourlize/README.md
bsergi: Missed that one in the cleanup--thanks for flagging.

hourlize/README.md
bsergi: The reason we have two at the moment is that hourlize runs can fit on the short queue, so this file is configured for that. We could read in the main one and overwrite with a different time setting--do you think that would be worth doing?

hourlize/resource.py
bsergi: This is actually a different step--the overwrite here refers to copying the original reV folders to our Supply_Curve_Data folder whereas the one you are talking about is copying the hourlize outputs back to that folder (the latter is set by copy_to_shared which I set off by default). 

That said, I didn't really like how this step was handled so I've changed it to prompt the user if the original files have already been copied to ask whether to use the existing ones or copy new ones. The copy functionality uses rsync, so it will also only copy differences. I think this approach is better but let me know if you had other feedback here.

bsergi (commented):
Thanks Patrick. I've made a few updates to address your comments. I've also uploaded the new offshore wind profiles to Zenodo. Everything worked for me, so feel free to give it a try when you're ready. 

Good call on the representative periods--I reran the reference case with a standard set of days and the difference went down substantially. Let me know if you think we should do the same check for open and limited.

As far as the site difference, it's a bit unclear where the previous estimates we have in ReEDS came from. In discussing with Gabe he confirmed that the total estimates in this PR are correct, so I'm inclined to believe this one more than the previous data.

hourlize/README.md
patrickbrown4: Ok thanks - nah I don't think it's worth extra processing, so fine with me to leave it

patrickbrown4 (approved):
Thanks Brian, I tried running locally and it worked, and I checked the difference in site-level CF for reference-access wind-ofs and there was no substantive difference aside from the dropped sites.

Image

PR #189, opened 2026-08-19
Add price-weighted capacity credit output

wesleyjcole:
## Summary
This pull request builds on #171 to add a price-weighted variant of the capacity-credit-like plots for using stress periods.  We have an option that looks at the top 10 hours, but this instead looks at the all hours and weights them by their price.  For regions that have less than 10 hours with a meaningful price, this will mean the ~$0 periods won't get included.  Here's an example from Alabama for 2050, which only has 7 higher priced hours (the % values show how much of the price weight occurs on that day):

AL_2050

For other regions that have more than 10 hours, this allows all of those hours get captured.  For example, here is the CA_LA region for 2050:

CA_LA_2050

It also changes the function's default to this new method because I think it is the most accurate way to estimate a marginal capacity credit.

### Implementation notes 
I added a filter for infinity values in the capacity-credit division (gen / cap).  There is sometimes a region with zero capacity but a tiny nonzero residual in generation (~1e-16), which creates an infinity value. This is was present for any metric, and I don't think it impacted them, but filtering it out means it won't cause unintended issues.

I also changed the ymax value of the plots to be dynamic because the capacity credit can go above 100% for some technologies (we measure capacity credit using net summer capacity, and net winter capacity can be meaningfully higher than net summer).

The output for this new metric looks like this (drawn from the run performed for #176):

stress_price_weighted_padded

### Details to double-check

- ~~[ ] Included comparison reports for appropriate test cases~~
  
Yes — I implemented this with Claude (I used plan mode for design, I asked it to stop for my review after it completed each step, and I used to for tracing for the inf values).

### DIFF
--- a/postprocessing/single_case_plots.py
+++ b/postprocessing/single_case_plots.py
@@ -827,6 +827,7 @@ if not int(sw.GSw_PRM_CapCredit):
         'stress_top10_netload',
         'stress_top10_load',
         'stress_bottom10_vregen',
+        'stress_price_weighted',
     ]:
         savename = f"plot_stress_cf-{level}-{metric}.png"
         try:
--- a/reeds/reedsplots.py
+++ b/reeds/reedsplots.py
@@ -5532,7 +5532,7 @@ def stress_mix_label(case, metric):
     elif metric == 'stress_mean':
         xlabel = 'Stress mean gen'
     elif metric == 'stress_price_weighted':
-        xlabel = 'Stress price-weighted gen'
+        xlabel = 'Stress: price-weighted gen'
     elif ('top' in metric) or ('bottom' in metric):
         direction = ('top' if 'top' in metric else 'bottom')
         xlabel = "Stress: {} {} {} hours".format(
--- a/reeds/reedsplots.py
+++ b/reeds/reedsplots.py
@@ -5815,7 +5815,7 @@ def plot_stress_mix(
 def plot_stress_cf(
     case:str|Path,
     level='transreg',
-    metric='stress_top10_netload',
+    metric='stress_price_weighted',
     include_rep=True,
     figwidth=1.2,
     figheight=1.2,
--- a/reeds/reedsplots.py
+++ b/reeds/reedsplots.py
@@ -5531,6 +5531,8 @@ def stress_mix_label(case, metric):
         xlabel = f"Rep {metric.split('_')[1]} gen"
     elif metric == 'stress_mean':
         xlabel = 'Stress mean gen'
+    elif metric == 'stress_price_weighted':
+        xlabel = 'Stress price-weighted gen'
     elif ('top' in metric) or ('bottom' in metric):
         direction = ('top' if 'top' in metric else 'bottom')
         xlabel = "Stress: {} {} {} hours".format(
--- a/reeds/reedsplots.py
+++ b/reeds/reedsplots.py
@@ -5304,7 +5304,7 @@ def separate_charge_discharge(df):
 
 def check_metric(metric):
     allowed = (
-        r'(cap|rep_mean|stress_(mean|(max|min|top\d+|bottom\d+)_(gen|load|netload|price|vregen)))'
+        r'(cap|rep_mean|stress_(mean|price_weighted|(max|min|top\d+|bottom\d+)_(gen|load|netload|price|vregen)))'
     )
     if not re.match(allowed, metric):
         raise ValueError(f"metric={metric} must match {allowed}")
--- a/reeds/reedsplots.py
+++ b/reeds/reedsplots.py
@@ -5444,6 +5444,23 @@ def get_cap_rep_stress_mix(
                     .divide(gen_h_stress.groupby('t').h.unique().map(len), axis=0)
                 )
 
+            elif key == 'stress_price_weighted':
+                ## Price-weighted average generation across all stress hours:
+                ## sum_h(gen*price) / sum_h(price)
+                price_long = price_stress.stack('r').rename('price').reset_index()
+                price_sum = price_long.groupby(['t','r'], as_index=False).price.sum()
+                gen_price = gen_h_stress.merge(price_long, on=['t','r','h'], how='left')
+                gen_price['price'] = gen_price['price'].fillna(0)
+                gen_price['gen_x_price'] = gen_price.MW * gen_price.price
+                numer = gen_price.groupby(['t','i','r'], as_index=False).gen_x_price.sum()
+                df = numer.merge(price_sum, on=['t','r'], how='left')
+                df['MW'] = df.gen_x_price / df.price
+                df = df.set_index(['t','i','r']).MW.unstack('r')
+                ## Guard against exact-cancellation division by zero (positive/negative hourly
+                ## prices summing to ~0 while gen_x_price is nonzero); NaN (0/0) is also possible
+                ## and both should collapse to 0
+                df = df.replace([np.inf, -np.inf], np.nan).fillna(0)
+
             ## Generation during regional max hours
             elif key.startswith('stress'):
                 direction = ('top' if 'top' in key else 'bottom')
--- a/reeds/reedsplots.py
+++ b/reeds/reedsplots.py
@@ -5849,6 +5849,9 @@ def plot_stress_cf(
     techs = dfstress['cap'].index.tolist()
     capcredit = (dfstress[metric] / dfstress['cap']).reindex(techs) * 100
     repfraction = (dfstress['rep_mean'] / dfstress['cap']).reindex(techs) * 100
+    ## Replace and INF values with NaN (can occur due to small-number division)
+    capcredit = capcredit.replace([np.inf, -np.inf], np.nan)
+    repfraction = repfraction.replace([np.inf, -np.inf], np.nan)
     ### Set up plot
     dfmap = reeds.io.get_dfmap(case)
     regions = dfmap[level].bounds.minx.sort_values().index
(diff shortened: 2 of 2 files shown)


reeds/reedsplots.py
```diff
@@ -5304,7 +5304,7 @@ def separate_charge_discharge(df):
 
 def check_metric(metric):
     allowed = (
-        r'(cap|rep_mean|stress_(mean|(max|min|top\d+|bottom\d+)_(gen|load|netload|price|vregen)))'
+        r'(cap|rep_mean|stress_(mean|price_weighted|(max|min|top\d+|bottom\d+)_(gen|load|netload|price|vregen)))'
```
patrickbrown4, line 5307: How about changing this to `stress_weight_price`? Then it would match the format of the other inputs, which are of the form `stress_{how to select}_{which parameter to use to select}`. (Would need to update the calls elsewhere too.)

postprocessing/single_case_plots.py
```diff
@@ -827,6 +827,7 @@
         'stress_top10_netload',
         'stress_top10_load',
         'stress_bottom10_vregen',
+        'stress_price_weighted',
```
patrickbrown4: Can you also add it to the list of metrics passed to `plot_cap_rep_stress_mix()` above, so we can see how it compares to the others?

For consistency it'd be good to also change the `top5`/`bottom5` in that list to `top10/bottom10` (and also add `'stress_top10_price'`) to match the versions shown here (sorry, I missed that in #171).

reeds/reedsplots.py
```diff
@@ -5877,7 +5899,13 @@ def plot_stress_cf(
...
             _ax.set_xlabel(None)
     ## Formatting
-    _ax.set_ylim(0, 100)
+    finite_vals = np.concatenate([
+        capcredit.values[np.isfinite(capcredit.values)],
+        repfraction.values[np.isfinite(repfraction.values)],
+    ])
+    datamax = finite_vals.max() if finite_vals.size else 0
+    ymax = datamax * 1.02 if datamax > 100 else 100
+    _ax.set_ylim(0, ymax)
```
patrickbrown4, line 5908: Simpler alternative would be to default to the existing ymax if it's over 100:
```suggestion
    _ax.set_ylim(0, max(100, _ax.get_ylim()[1]))
```

reeds/reedsplots.py
```diff
@@ -5444,6 +5444,23 @@ def get_cap_rep_stress_mix(
...
+                gen_price['price'] = gen_price['price'].fillna(0)
+                gen_price['gen_x_price'] = gen_price.MW * gen_price.price
+                numer = gen_price.groupby(['t','i','r'], as_index=False).gen_x_price.sum()
+                df = numer.merge(price_sum, on=['t','r'], how='left')
+                df['MW'] = df.gen_x_price / df.price
+                df = df.set_index(['t','i','r']).MW.unstack('r')
+                ## Guard against exact-cancellation division by zero (positive/negative hourly
+                ## prices summing to ~0 while gen_x_price is nonzero); NaN (0/0) is also possible
+                ## and both should collapse to 0
+                df = df.replace([np.inf, -np.inf], np.nan).fillna(0)
```
patrickbrown4: Reshaping big dataframes with an h index is slow. Suggested alternative:
```suggestion
```

reeds/reedsplots.py
```diff
@@ -5796,7 +5815,7 @@ def plot_stress_mix(
 def plot_stress_cf(
     case:str|Path,
     level='transreg',
-    metric='stress_top10_netload',
+    metric='stress_price_weighted',
```
patrickbrown4, line 5818: The default doesn't really matter since we always specify, but could you say more about why you think this method is the most accurate way to estimate a marginal capacity credit? I understand that it reflects the hours that drive capacity revenues, but it seems the most different from how we've calculated capacity credit in the past.

My other concern is about stability; if it's based on fewer hours, it seems like it could jump around more between years/regions/runs and make it hard to tell trends from noise. For example, I'm not sure how "real" the low values for storage in Texas from 2032-2038 are:

Image

None of these appraoches are perfect because we're showing dispatch and not availability. So I guess it's more about whether we want a metric that's more similar to the way we calculate capacity credit (top 10 net load or top 10 load), or one that better captures the timing of the price spikes (price weighted).

reeds/reedsplots.py
```diff
@@ -5444,6 +5444,23 @@ def get_cap_rep_stress_mix(
                     .divide(gen_h_stress.groupby('t').h.unique().map(len), axis=0)
                 )
 
+            elif key == 'stress_price_weighted':
+                ## Price-weighted average generation across all stress hours:
+                ## sum_h(gen*price) / sum_h(price)
+                price_long = price_stress.stack('r').rename('price').reset_index()
+                price_sum = price_long.groupby(['t','r'], as_index=False).price.sum()
```
patrickbrown4, line 5451: Reshaping big dataframes with r/h indices can be slow (maybe not for the default case but for high-temporal/spatial-resolution cases). You can sum by `t` first and avoid the big reshape:

```suggestion
                price_sum = price_stress.groupby('t').sum().stack('r')
```

patrickbrown4: If we change the weighting for stress hours (discussed in #154), summing over hours directly will be wrong. Safer to first multiply by the hour weight in `hours_t(allh,t)`.

reeds/reedsplots.py
```diff
@@ -5444,6 +5444,23 @@ def get_cap_rep_stress_mix(
                     .divide(gen_h_stress.groupby('t').h.unique().map(len), axis=0)
                 )
 
+            elif key == 'stress_price_weighted':
```
patrickbrown4: How should negative hours be handled?

reeds/reedsplots.py
patrickbrown4, line 5402: For the suggested change below (avoid reshaping)
```suggestion
    price_stress = price_stress.groupby(['t','r','h']).Value.max()
```

patrickbrown4, line 5330: For the suggested change below. Import it here to localize the special-case dependence
```suggestion
    from reeds.core.terminus import report_calcs
    ### Check inputs
```

patrickbrown4, line 5354: For the suggested change below
```suggestion
    hours_t = (
        report_calcs.get_gams_results(case, 'hours_t')['hours_t'].reset_index()
        .astype({'t':int}).rename(columns={'allh':'h'}).set_index(['h','t']).squeeze(1)
    )
```

patrickbrown4, line 5473: Avoid reshaping
```suggestion
                    dfindex = price_stress.reorder_levels(['t','r','h'])
```

reeds/reedsplots.py
patrickbrown4, line 5462: If we change the weighting for stress hours (discussed in #154), summing over hours directly will be wrong. Safer to first multiply by the hour weight in `hours_t(allh,t)`.  Here's my suggestion:
```suggestion
                ## sum_h(gen * price * hours) / sum_h(price * hours)
                price_weighted = price_stress.multiply(hours_t, axis=0).dropna()
                df = (
                    (gen_h_stress.set_index(['t','i','r','h']).squeeze(1) * price_weighted)
                    .groupby(['t','i','r']).sum()
                    / price_weighted.groupby(['t','r']).sum()
                ).unstack('r').replace([np.inf, -np.inf], np.nan).fillna(0)
```
I'll note the other changes above/below

postprocessing/single_case_plots.py
patrickbrown4, line 785: Let's match the same options used below to facilitate comparison (not sure if the suggestion will apply correctly since it's outside the changed region but you get the idea)
```suggestion
        metrics = [
            'cap',
            'rep_mean',
            'stress_mean',
            'stress_top10_load',
            'stress_top10_netload',
            'stress_bottom10_vregen',
            'stress_top10_price',
            'stress_price_weighted',
        ]
```

patrickbrown4 (commented):
Thanks Wesley, sorry for the slow response.

reeds/reedsplots.py
patrickbrown4, line 5354: Sorry, the one other required change is to `get_gams_results()` in `report_calcs.py`; it should now be:
```python
def get_gams_results(case, key=None):
    print('Loading results.gdx')
    if isinstance(key, str):
        dictin = {key: gdxpds.to_dataframe(Path(case, 'outputs', 'results.gdx'), key)}
    else:
        dictin = gdxpds.to_dataframes(Path(case, 'outputs', 'results.gdx'))
    ## Set indices as multiindex
    valcols = ['Value','Level','Marginal','Lower','Upper','Scale']
    for key, df in dictin.items():
        indices = [i for i in df if i not in valcols]
        dictin[key] = df.set_index(indices).squeeze(1)
    print('Finished loading results.gdx')
    return dictin
```

reeds/reedsplots.py
wesleyjcole: I included the 2% because otherwise the line at the max value would get clipped, showing up very thin and making the plot look weird, so I'll keep that buffer value still.

wesleyjcole: Correction: `_ax.get_ylim()[1]` autoscales (didn't know that), so a buffer is already included.  So, yes, I'll adopt this one.

reeds/reedsplots.py
wesleyjcole: Good call.  I made the change, but still left it explicit in this list so that users won't think things like "stress_weight_load" would work.

reeds/reedsplots.py
wesleyjcole: Agreed, though I'm taking a slightly different approach.  I tried a benchmark with lots of hours and regions and `gen.set_index(['t','i','r','h']).squeeze(1) * price_weighted` was relatively slow.

postprocessing/single_case_plots.py
wesleyjcole: Done.

reeds/reedsplots.py
wesleyjcole: I put the import statement next to where it is used rather than here.  Is that reasonable convention, or is it better to have it here?

wesleyjcole: I put this in the if statement so that it is only loaded if it is used.

wesleyjcole: Done.

reeds/reedsplots.py
wesleyjcole: I'm updated the process and tested it againt a large dataframe (lots of r and t) and it worked quickly.

wesleyjcole:
On the default method, I think the price-weighted method is best because it is dynamic in capturing the timeslices when new capacity is actually needed.  If only one timeslice matters (has non-near-zero marginal cost), this with report it directly.  If 30 timeslices matter, it will report across all 30.  To me that is the best sense for how much new firm capacity will I get by adding technology X?  It will be the contribution of that same type of technology during the hours that matter.

PR #194, opened 2026-08-25
Copy MGA solver file if used

patrickbrown4:
## Summary

In #36, I forgot to copy the solver file specified by the `GSw_gopt_mga` switch to the case folder. So runs using MGA and `GSw_gopt_mga != GSw_gopt` were not finding the specified solver file and reverting to default solver settings, likely leading to slow performance with MGA. This PR copies the solver file specified by `GSw_gopt_mga` when using MGA.

I verified that the default MGA solver file (`cplex.op3`) now shows up in the case folder for the Pacific_MGA test case and not for a non-MGA test case.

The warning observed in `gamslog.txt` for the Pacific_MGA case on the main branch (`*** Error Cannot open parameter file "/Users/pbrown/github2/ReEDS/runs/v20260825_mainM0_Pacific_MGA/cplex.op3"`) goes away with this change.

Zero change for the Pacific test case: [results-v20260825_mainM0_Pacific,v20260826_mgaM0_Pacific.pptx](https://github.com/user-attachments/files/31432834/results-v20260825_mainM0_Pacific.v20260826_mgaM0_Pacific.pptx)

### Details to double-check

No

### DIFF
--- a/reeds/input_processing/copy_files.py
+++ b/reeds/input_processing/copy_files.py
@@ -954,6 +954,10 @@ def write_miscellaneous_files(
     case = Path(inputs_case).parent
     optfile = reeds.io.get_optfile(case)
     shutil.copy(Path(reeds_path, 'reeds', 'solver', optfile), case)
+    ## If using MGA, copy its solver file too
+    if float(sw.GSw_MGA_CostDelta) > 0:
+        optfile = reeds.io.get_optfile(case, GSw_gopt=sw.GSw_gopt_mga)
+        shutil.copy(Path(reeds_path, 'reeds', 'solver', optfile), case)
 
     ### Parsed switches
     pd.DataFrame(
--- a/reeds/io.py
+++ b/reeds/io.py
@@ -666,6 +666,10 @@ def get_switches_base(case=None, **kwargs):
             index_col=0,
             header=None,
         ).squeeze(1)
+    ### Overwrite values with keyword arguments if provided
+    for key, value in kwargs.items():
+        if key in sw.keys():
+            sw[key] = value
     return sw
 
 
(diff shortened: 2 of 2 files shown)


reeds/input_processing/copy_files.py
```diff
@@ -954,6 +954,10 @@ def write_miscellaneous_files(
     case = Path(inputs_case).parent
     optfile = reeds.io.get_optfile(case)
     shutil.copy(Path(reeds_path, 'reeds', 'solver', optfile), case)
+    ## If using MGA, copy its solver file too
+    if float(sw.GSw_MGA_CostDelta) > 0:
+        optfile = reeds.io.get_optfile(case, GSw_gopt=sw.GSw_gopt_mga)
```
atpham88, line 959: Very minor but I wonder if it's helpful to copy all the opt files instead? Because sometimes when an MGA run is stuck in certain year, switching to another opt setting mid-run when restarting might help, instead of having to completely rerun. But maybe that's not good practice...

patrickbrown4: The `GSw_gopt_mga` value is read from the `inputs_case/switches.csv` file and then passed to the `3_solve_oneyear.gms` call [here](https://github.com/ReEDS-Model/ReEDS/blob/5955870876d3a1dceff7014be77389ded9d6bc85/reeds/inputs.py#L358), so pointing it to a different solver file requires changing the switches mid-run, which seems unsafe to me. But it's fine to adjust the solver settings within the specified file (`{case}/cplex.op3` or `{case}/cplex.opt` depending on which part of the model) before restarting, or to modify the `{reeds_path}/reeds/solver/cplex.op3` file and then copy it into the run via the `--copy_solver_settings/-c` option of [restart_runs.py](https://github.com/ReEDS-Model/ReEDS/blob/5955870876d3a1dceff7014be77389ded9d6bc85/helpers/restart_runs.py#L18). Personally I would lean toward one of those options over changing `switches.cvs` manually during a run.

atpham88: Yeah in the past I updated switches.csv to change mga opt file when the run was stuck and sometimes it helped. But I agree I don't do it often and if in those rare cases I can just copy the file over. Thanks.

atpham88 (approved):
Thanks Patrick for fixing this.

Issue #185, opened 2026-08-18
Separate resource class from technology set `i`


patrickbrown4:
We currently bake the resource class for UPV, wind, and geothermal into the technology set `i` (e.g. `wind-ons_1` through `wind-ons_10`), so variables like `GEN(i,v,r,h,t)`, `OPRES(i,v,r,h,t)`, `CREDIT_H2PTC(i,v,r,h,t)`, `RECS(RPSCat,i,st,ast,t)`, etc. are resolved by resource class.

But we don't actually care about resource class for those variables; the only place resource class matters is in `eq_capacity_limit(i,v,r,h,t)`, where `GEN ≤ m_cf(i,v,r,h,t) * CAP(i,v,r,t)`. (Plus a few hybrid-specific equations.) So keeping the class in `i` expands the model size and increases degeneracy in the operations (since all these techs have zero marginal cost).

The hypothesis is that we might be able to do something similar to #40 for resource classes by separating out the class into its own `c` index, which only gets used in the few equations where we interact with `m_cf`. (The idea is similar to how we handle supply curves throughout the model with sets like `rscbin` and `tscbin`.) Then all the rest of the `i`-dependent variables and equations would be smaller, which hopefully would reduce solve time or enable higher time resolution. It would also help reduce duplication in the input data.

Some specific components (please update or rewrite as work progresses):

- [ ] Consider splitting offshore wind up into two technologies, `wind-ofs-fixed` and `wind-ofs-floating`; these are currently implicitly embedded in the class, but since they have different capex costs, it might help to break them up
- [ ] Reindex `m_cf` to `m_cf(i,c,v,r,h,t)`
  - [ ] Same for `cf_rsc(i,c,v,r,h,t)`
  - [ ] which comes from `cf_in(i,c,r,h)`
  - [ ] which is loaded from `cf_vre.csv`
  - [ ] which is written by `hourly_writetimeseries.py`
- [ ] Add `CAP_CLASS(i,c,v,r,t)`, where `CAP(i,v,r,t) = sum{c, CAP_CLASS(i,c,v,r,t)}`
- Option 1:
  - [ ] Add `GEN_CLASS(i,c,v,r,h,t)`, where `GEN(i,v,r,h,t) = sum{c, GEN_CLASS(i,c,v,r,h,t)}`
[...]

PR #196, opened 2026-08-25
Consolidate GAMS tech-range expansion and add resource-class map for wind and PV

wesleyjcole:
## Summary
This PR makes two changes to support #185:

1. It consolidates the various GAMS "x_1*x10" range expansion code into a single function.
2. It extends `inputs/techs/tech_resourceclass.csv` to cover upv, pvb1/2/3, wind-ons, and wind-ofs.

The long-term goal of #185 is to remove the class information from the `i` set to reduce the variables (particularly `GEN`) that are indexed by `i`, and both of these changes will help enable that.

### Implementation notes 
The GAMS range expansion happened in three different places in the code (`reeds_to_rev.py`, `recf.py`, and `techs.py`) with slightly different flavors.  This PR consolidates those and allows them to use the same functions in `techs.py`.

A national run shows exactly zero change: [results-Main,TechExpand.pptx](https://github.com/user-attachments/files/31439517/results-Main.TechExpand.pptx).

The gdxdiff is different due to the extra entries for tech-resource mapping.

### Details to double-check

I used Claude Code to survey how resource classes are represented across the model, generate the `tech_resourceclass.csv` rows, and perform the refactor and verification. I reviewed and tested each change individually.


reeds/techs.py
```diff
@@ -11,30 +11,46 @@
 
 
 ### Functions
-def expand_GAMS_tech_groups(df):
+def expand_star(techs):
```
bsergi, line 14: I like having this consolidated into 1 function

bsergi (approved):
Looks good to me

PR #203, opened 2026-08-28
Add class set to capacity factor parameters

wesleyjcole:
## Summary
This pull request adds a  new index `c` (resource class) to the capacity factor inputs (`cf_in`, `cf_rsc`, `m_cf`, `cf_adj_t`) as a step toward moving resource class out of the `i` (tech) set per issue #185. The `c` set is redundant with the class embedded in tech names, so the model itself is unchanged.

### Implementation notes 
I renamed the set `resourceclass` to `c`, and converted the parameter `inputs/techs/tech_resourceclass.csv` into a mapping set `i_class(i,c)`.  Any technology that does not have a class gets mapped to class "0". Every `i` maps to exactly one `c`.

The compare report shows exactly no changes: [results-Main,CFClass.pptx](https://github.com/user-attachments/files/31525332/results-Main.CFClass.pptx).

### Details to double-check

Yes, I used Claude Sonnet to implement the code changes and perform testing before starting a full US run.


reeds/core/solve/6_data_dump.gms
```diff
@@ -215,8 +215,8 @@ cost_cap_fin_mult_filt(i,r,t)$([storage_standalone(i)]) = cost_cap_fin_mult(i,r,
 
 cost_vom_filt(i,v,r)$cap_exist(i,v,r) = sum{t$tcur(t), cost_vom(i,v,r,t) } ;
 
-cf_adj_t_filt(i,v,t)$[cap_exist_iv(i,v)$trange(t)] = cf_adj_t(i,v,t) ;
-cf_adj_t_filt(i,v,"%next_year%") = cf_adj_t(i,v,"%next_year%")$(vre(i) or pvb(i)) ;
+cf_adj_t_filt(i,v,t)$[cap_exist_iv(i,v)$trange(t)] = sum{c$i_class(i,c), cf_adj_t(i,c,v,t) } ;
+cf_adj_t_filt(i,v,"%next_year%") = sum{c$i_class(i,c), cf_adj_t(i,c,v,"%next_year%") }$(vre(i) or pvb(i)) ;
```
cavraam: double-checking: We want to sum the cf_adj_t over all classes of a vintage, right?

reeds/core/setup/c_model.gms
```diff
@@ -1118,7 +1118,7 @@ eq_site_cf(x,h,t)
         $x_r(x,r)
         $valgen(i,v,r,t)],
 * Capacity factor of techs with endogenously-modeled spur lines
-        m_cf(i,v,r,h,t)
+        sum{c$i_class(i,c), m_cf(i,c,v,r,h,t) }
```
cavraam: Nit-pick: It seems like we do `sum_c m_cf(i,c,v,r,h,t)` across the model. 

- If we plan to move to the whole equation being defined for (i,c,v,r,h,t) down the road, then this is moving towards the right direction.

- If we plan to keep the same level of detail (i,v,r,h,t), you could consider importing m_cf**_in**(i,c,v,r,h,t) and collapsing the new parameter into the previous m_cf(i,v,r,h,t) = sum_c m_cf**_in**(i,c,v,r,h,t).

No strong feelings though, even if we're moving towards the second bullet, it's just slightly easier to follow the model code when avoiding redundant notation.

reeds/core/terminus/report.gms
```diff
@@ -654,7 +654,7 @@ gen_h(i,r,h,t)$[tmodel_new(t)$valgen_irt(i,r,t)] =
 * UPV capacity is already in MWac at this point (matching csp-ns),
 * so don't need to account for ILR.
 gen_h("csp-ns",r,h,t)$[cap_cspns(r,t)$tmodel_new(t)]
-    = cap_cspns(r,t) * m_cf("upv_5","new1",r,h,t) ;
+    = cap_cspns(r,t) * sum{c$i_class("upv_5",c), m_cf("upv_5",c,"new1",r,h,t) } ;
```
cavraam: Double-checking here:
- 'upv_5' is not (upv,5) under this PR, right?
- Is there a reason this should be '5'? E.g., should this be related to the last class, which we can grasp programmatically: i_class(i,c)$(ord(c) = card(c))

inputs/sets/README.md
```diff
@@ -92,6 +94,8 @@
 
 - `i.csv`: set of technologies
 
+- `i_class.csv`: map from technologies to their resource class (`c.csv`)
```
cavraam: Nit-pick: Is 'i_class' preferred over 'i_c'? No strong feelings at all (the current naming is also easy to follow); I just try to keep the names and mappings as consistent as possible...

cavraam (approved):
A few questions attached, mostly double-checking and for my understanding. Otherwise, this PR is ready to merge.

Thank you @wesleyjcole !

reeds/core/solve/6_data_dump.gms
wesleyjcole: That's right.

reeds/core/setup/c_model.gms
wesleyjcole: Thanks Charalampos.  I'm fine keeping it as is, as the next step will be to thread it into a new CAP_CLASS variable, and I think this structure will work well with that next step.

reeds/core/terminus/report.gms
wesleyjcole: The `i` set hasn't changed, so `i` is still `upv_5`.  `c` is now `5` for `upv_5`.  The reason it is 5 is because CSP was added into the supply curve as a 5.  It's fragile, and Patrick recommended removing it as part of the update to #185.  I plan to figure out how to deal with this one in the 5th PR for the #185 updates.

inputs/sets/README.md
wesleyjcole: Good call.  This is inconsistent with the other class pairings, so I'll change it to `i_c`.

reeds/core/setup/b_inputs.gms
```diff
@@ -3766,18 +3774,18 @@ $offdelim
 $onlisting
 / ;
 
-parameter cf_adj_t(i,v,t)        "--unitless-- capacity factor adjustment over time for RSC technologies" ;
+parameter cf_adj_t(i,c,v,t)    "--unitless-- capacity factor adjustment over time for RSC technologies" ;
```
patrickbrown4, line 3777: FYI I'm pretty sure this parameter is used by R2X (we write it for them [here](https://github.com/ReEDS-Model/ReEDS/blob/848c1c64a87ac0f8656593a36328ad2da58ce22a/reeds/core/terminus/report_params.csv#L247)); it won't change now since `i` is unchanged, but R2X will need to be adjusted once the class is moved out of `i`

PR #206, opened 2026-09-01
Fix double-counting of distribution losses for distributed PV

### LINKED ISSUE #202: `distloss` adjustement is applied twice to `distpv` profiles
### Summary
Because distributed pv (`distpv`) generation is at the point of consumption (end-use load), but ReEDS uses busbar load, we increase `distpv` profiles by the distribution loss rate (`distloss`).  That adjustment is applied twice in the model.  Once in `recf.py`:

```
    ### Scale up distpv by 1/(1-distloss)
    recf.loc[
        :, resources.loc[resources.tech.isin(vre_dist),'resource'].values
    ] /= (1 - distloss)
```

and again in `2_temporal_params.gms`:

```
* distpv capacity factor is divided by (1.0 - distloss) to provide a busbar equivalent capacity factor
m_cf(i,v,r,h,t)$[distpv(i)$valcap(i,v,r,t)] = m_cf(i,v,r,h,t) / (1.0 - distloss) ;
```

The double-counting is confirmed via `distpv` capacity factor values for p1:

`inputs/profiles_cf/cf_distpv_county.h5`: Max CF = 0.826 (ILR-adjusted maximum, 1/1.21 = 0.826)
`inputs/profiles_cf/cf_distpv_county.h5`: Max CF after aggregating counties to p1 = 0.7824
`inputs_case/recf.h5`: Max CF = 0.8237 ( = 0.7824 / (1 - 0.05), small float difference because of float16)
`m_cf` in GAMS: Max CF = .8671 ( = 0.8237 / (1 - 0.05))

There are a few lingering `cf_gather.py` references related to inputs here, so I think this double-application just appeared somewhere in our transition from `cf_gather.py` to our new configuration.

There are two ways to fix this:

1. Remove the `distloss` adjustment in `2_temporal_params.gms`
2. decrease the CF by `distloss` in `hourly_writetimeseries.py`

Option 1 is more straightforward, and then there wouldn't be three adjustments.  Option 2 would mean that the profiles in `inputs_case/rep` would be consistent with `load_allyear.csv` (which is in end-use terms), which could create inconsistencies or problems in other venues. (Removing the `distloss` adjustment in `recf.py` is not an option because that would create inconsistencies in PRAS)

There might be a third option where we rethink the entire `distloss` application and make it more straightforward throughout, but I haven't thought through that one yet.

I also think we could likely create a test that would check for things like this, so that it doesn't unintentionally slip through.

wesleyjcole:
## Summary
Per #202, the distribution loss adjustment (`distloss`) is applied twice to distributed PV (`distpv`), resulting in `distpv` profiles being 5% too high.  Tracing the `distloss` adjustments was challenging because it was applied, removed, and applied again (it looks like that was just the natural evolution of the development, not a design decision).  This PR adjusts the application of `distloss` to both distributed PV and load so that it is only applied once, which should make understanding what is end-use vs. busbar more straighforward.

### Implementation notes 

site | main | This PR
-- | -- | --
hourly_load.py:507 | /(1-dl) → busbar | kept — the one load conversion
recf.py:470 | distpv /(1-dl) → busbar | kept — the one distpv conversion
hourly_writetimeseries.py:208 | *(1-dl) down | deleted
2_temporal_params.gms:765 | /(1-dl) up | deleted → load_exog = load_allyear
2_temporal_params.gms:837, 852 | /(1-dl) up | deleted
2_temporal_params.gms:520 | distpv /(1-dl) up | deleted
compare_cases.py:633-639 | hardcoded distloss = 0.05, grosses peak_ccseason | deleted — file is already busbar, and the hardcoded scalar goes with it
c_model / report.gms end-use conversions | busbar→end-use at point of use | unchanged

`uncertainty_plots.py` includes a load plot, which now plots busbar load instead of end-use load. If that is intended to show end-use load, then we need a distloss adjustment there as well.

### Additional changes 
There were a few mentions to `cfgather.py`, which no longer exists.  Those are removed.

### Issues resolved 
Closes #202. 

As expected, the `distpv` generation is now lower by the `distloss` amount (representative day selection did not change with this change).  That generation is made up by gas, wind, and utility PV.

image

Capacity impacts are less than 5 GW:

image

I also did a gdxdiff of the `inputs.gdx` files and verified that the inputs (load, `distpv` CF values) were correct.

Compare report: [results-Main,distlossfix.pptx](https://github.com/user-attachments/files/31701307/results-Main.distlossfix.pptx)

### Details to double-check

  - ~~[ ] Dollar year recorded and converted to 2004$ for GAMS~~
  - ~~[ ] Timeseries are in Central Time~~
  - [x] Units are specified
  - ~~[ ] Preprocessing steps have been documented and committed to [ReEDS_Input_Processing](https://github.com/ReEDS-Model/ReEDS_Input_Processing)~~
  - ~~[ ] New large data files handled with .h5 instead of .csv~~
  - ~~[ ] If new parameters are added to `d_objective.gms`, they are included in `objective_function_params.yaml` for completeness checking~~

Yes, I used a LLM to brainstorm the fix and then to implement it.


bsergi (approved):
Looks good to me. Great to get the distributed PV issue fixed and helpful to have the treatment of the distribution losses on load simplified (I didn't realize load.h5 was busbar but load_allyears.csv wasn't).

PR #210, opened 2026-09-02
Add `CAP_CLASS` variable and give the resource-supply-curve variables a class index

wesleyjcole:
## Summary
This pull request adds the variable `CAP_CLASS(i,c,v,r,t)`, which rolls up to `CAP` through a new equation (`eq_cap_class_total`). It also allows the `m_cf` terms (which already have the `c` index added) to be multiplied against it.  Finally, the PR gives `CAP_RSC` and `INV_RSC` the same `c` index.  It supports #185, which (when completed) will reduce model size and degeneracy.

### Implementation notes 
`valcap_class(i,c,v,r,t)` is added as a new filtering set.  It is only populated for `cf_tech(i)` technologies.

This implementation grows the model size slightly, and the reduction won't happen until the classes are moved out of the `i` set.

### Issues resolved 
Partial progress on #185.

I ran Pacific test case and a combined `Pacific_CC` + interday-storage case (because the changes touched equations gated by those switches), each against a corresponding baseline on main.  The comparison showed the following impacts:

| | Pacific | Pacific_CC + interday |
| --- | --- | --- |
| Objective (`systemcost` total) | delta exactly 0 | delta exactly 0 |
| `cap`, `cap_ivrt`, `tran_out`, `error_check` | byte-identical | byte-identical |
| `gen_ann` | max rel diff 7.4e-13 | exact compensating swap, total delta 0 |
| Model size, 2032 | +586 eq / +586 var | +586 eq / +586 var |

Also, `CAP_CLASS.l == CAP.l` exactly.

The remaining differences are LP degeneracy, including intra-day hydro/PSH dispatch reshuffling, a `upv` to `wind-ons` curtailment shift of ±51,251.645796 MWh in one region-year (equal and opposite values to six decimals with total curtailment unchanged), and a `CAP_SDBIN` storage-bin reallocation.

I also ran a USA_defaults case, which showed almost zero changes in the solution, with the small changes to be expected because we have adjusted the LP.

image

Full compare report: [results-Main,CAP_CLASS.pptx](https://github.com/user-attachments/files/31745011/results-Main.CAP_CLASS.pptx)

Runtime increased in the USA_defaults test by about 20 minutes, which could be noise, but also could be driven by the increase in model size.

### Details to double-check

- ~~[ ] Reusable functions used where possible instead of copy/pasted code~~

Yes, I used Claude Opus 5 to implement the code changes and to perform the Pacific test case comparisons.


reeds/core/solve/4_post_solve_adjustments.gms
```diff
@@ -24,6 +24,7 @@ if(Sw_NewValCapShrink = 1,
                                 UPGRADES.l(ii,v,r,tt)})
                        ] = yes ;
     valcap(i,v,r,t)$valcap_remove(i,v,r,t,"%cur_year%") = no ;
+    valcap_class(i,c,v,r,t)$valcap_remove(i,v,r,t,"%cur_year%") = no ;
```
cavraam, line 27: I wonder if adding $i_c here makes sense down the road, when c is completely broken out of i?

reeds/core/terminus/report.gms
```diff
@@ -1240,12 +1241,12 @@ systemcost_techba("inv_investment_capacity_costs",i,r,t)$tmodel_new(t) =
               + sum{v$[valinv(i,v,r,t)$battery(i)],
                    INV_ENERGY.l(i,v,r,t) * (cost_cap_fin_mult_noITC(i,r,t) * cost_cap_energy(i,t) ) }
 *plus supply curve adjustment to capital cost (separated in outputs but part of m_rsc_dat(r,i,rscbin,"cost"))
-              + sum{(v,rscbin)$[m_rscfeas(r,i,rscbin)$valinv(i,v,r,t)$rsc_i(i)$[not sccapcosttech(i)]$(not spur_techs(i))],
-                   INV_RSC.l(i,v,r,rscbin,t) * m_rsc_dat(r,i,rscbin,"cost_cap") * rsc_fin_mult_noITC(i,r,t) }
+              + sum{(c,v,rscbin)$[m_rscfeas(r,i,rscbin)$valinv(i,v,r,t)$rsc_i(i)$[not sccapcosttech(i)]$(not spur_techs(i))$i_c(i,c)],
```
cavraam: - Nit-pick: [not sccapcosttech(i)] -> (not sccapcosttech(i))?
- Even greater nit-pick (also below): Would it be helpful to add the conditionals in the order the respective sets appear, so, $i_c first? I know this is not part of our coding practices, it just helps with memorizing batches of conditionals that appear repeatedly. Happens in a few

reeds/core/terminus/report.gms
```diff
@@ -1328,8 +1329,8 @@ systemcost_techba("inv_investment_water_access",i,r,t)$tmodel_new(t) =
 *cost of water access
               + (8760/1E6) * sum{ (v,w)$[i_w(i,w)$valinv(i,v,r,t)], sum{wst$i_wst(i,wst), m_watsc_dat(wst,"cost",r,t)} * water_rate(i,w) *
                         ( INV.l(i,v,r,t) + INV_REFURB.l(i,v,r,t)$[refurbtech(i)$Sw_Refurb] ) }
-              + sum{(rscbin,v)$[m_rscfeas(r,i,rscbin)$psh(i)], sum{wst$i_wst(i,wst), m_watsc_dat(wst,"cost",r,t) } *
-                        ( INV_RSC.l(i,v,r,rscbin,t) * water_req_psh(r,rscbin) ) }
+              + sum{(rscbin,c,v)$[m_rscfeas(r,i,rscbin)$psh(i)$i_c(i,c)], sum{wst$i_wst(i,wst), m_watsc_dat(wst,"cost",r,t) } *
```
cavraam: (cleanup suggestion, feel free to ignore) not part of this PR, but this might be easier to track if there was only one sum here:

sum{(wst,rscbin,c,v)$[i_wst(i,wst)$i_c(i,c)$m_rscfeas(r,i,rscbin)$psh(i)], 
        m_watsc_dat(wst,"cost",r,t)  * INV_RSC * water_req_psh}

cavraam (approved):
This PR is ready to merge:
- Results differences are negligible and are expected given the additional equations and variables.
- Comments include nit-picks (couldn't find anything else, really).
- I also checked that 'm_cf/m_cf_...' parameters are conditioned on cf_tech in d_temporal_params.gms.

reeds/core/terminus/report.gms
wesleyjcole: I'm going to hold off on this one until we get to the outputs portion of #185.  I think it will be best to address there.

reeds/core/solve/4_post_solve_adjustments.gms
wesleyjcole: I don't think we want a `$i_c` here because we want all class of a tech removed if `valcap_remove` indicates they should be removed.

reeds/core/terminus/report.gms
wesleyjcole: Good suggestions.  I'll apply both.

PR #219, opened 2026-09-11
Additional Class Expansion

wesleyjcole:
## Summary
(Note: this is getting merged into a dev branch and not into main)

This PR continues #185 to separate the resource class from the technology.  I thought I was ready to start collapsing the technology sets, but found several other places where class expansion was needed, or where it would eventually would be required.  This pull request continues the class expansion, but still ignores some capacity credit parameters (I want to do those in a separate PR because they will use the capacity credit switch which is off by default).

This PR is 6a in the list at https://github.com/ReEDS-Model/ReEDS/issues/185#issuecomment-5427027139.

### Implementation notes 
The primary changes were with exogenous capacity and prescribed builds, which now carry class.  The resources class addition was further propagated.  And the additional mapping of class meant that the `i2c` mapping in `writesupplycurves.py` could be removed.

### Issues resolved 
Partially addresses #185.

### Known incompatibilities 
R2X has not yet been fixed.

The Pacific test case showed that all outputs were exactly the same.  The USA_defaults test case was exactly identical until 2035, where the objective function differred by 1e-14.  That tiny difference propagated into the other tiny differences in 2040 and beyond.  The objective function stayed within 1e-4, and the model size was exactly the same in every solve year and iteration after the change.

USA_defaults capacity difference:

image

USA_defaults compare report: [results-Baseline,ClassExpand.pptx](https://github.com/user-attachments/files/32117620/results-Baseline.ClassExpand.pptx)

### Details to double-check

Yes, I used Claude Opus 5 to implement the code changes and to help me with verification of those changes (e.g., reading .lst files for model size, comparing outputs, etc.).

### DIFF
--- a/reeds/core/setup/b_inputs.gms
+++ b/reeds/core/setup/b_inputs.gms
@@ -1522,7 +1522,7 @@ $onempty
 parameter capacity_exog(i,v,r,allt)             "--MW-- exogenously specified capacity",
           capacity_exog_energy(i,v,r,allt)      "--MWh-- exogenously specified energy capacity",
           capacity_exog_rsc(i,c,v,r,rscbin,allt)  "--MW-- exogenous (pre-tfirst) capacity for wind-ons and upv",
-          m_capacity_exog(i,v,r,allt)           "--MW-- exogenous power capacity used in the model",
+          m_capacity_exog(i,c,v,r,allt)         "--MW-- exogenous power capacity used in the model",
           m_capacity_exog_energy(i,v,r,allt)    "--MWh-- exogenous energy capacity used in the model",
           geo_cap_exog(i,r)                     "--MW-- existing geothermal capacity"
 /
@@ -1637,9 +1637,11 @@ avail_retire_exog_rsc(i,v,r,t)$[refurbtech(i)$(capacity_exog(i,v,r,t-1) > capaci
 
 avail_retire_exog_rsc(i,v,r,t)$[not initv(v)] = 0 ;
 
-m_capacity_exog(i,v,r,t)$capacity_exog(i,v,r,t) = capacity_exog(i,v,r,t) ;
+m_capacity_exog(i,c,v,r,t)$[i_c(i,c)$capacity_exog(i,v,r,t)] = capacity_exog(i,v,r,t) ;
 m_capacity_exog_energy(i,v,r,t)$capacity_exog_energy(i,v,r,t) = capacity_exog_energy(i,v,r,t) ;
-m_capacity_exog(i,"init-1",r,t)$geo(i) = geo_cap_exog(i,r) ;
+m_capacity_exog(i,c,"init-1",r,t)$[i_c(i,c)$geo(i)] = geo_cap_exog(i,r) ;
+m_capacity_exog(i,c,v,r,t)$sum{rscbin, capacity_exog_rsc(i,c,v,r,rscbin,t) } =
+    sum{rscbin, capacity_exog_rsc(i,c,v,r,rscbin,t) } ;
 
 * We assign the ~1.3 GW of existing csp-ns to upv throughout the model, both in the
 * exogenous and the prescribed capacity, and convert it back to csp-ns when reporting.
@@ -1719,8 +1721,8 @@ $onlisting
 
 parameter inv_distpv(r,t) "--MW-- capacity of distpv that is build in year t (i.e., INV for distpv)" ;
 
-inv_distpv(r,t) = sum{(i,v)$distpv(i),
-                      m_capacity_exog(i,v,r,t) - sum{tt$tprev(t,tt), m_capacity_exog(i,v,r,tt) }
+inv_distpv(r,t) = sum{(i,c,v)$[distpv(i)$i_c(i,c)],
+                      m_capacity_exog(i,c,v,r,t) - sum{tt$tprev(t,tt), m_capacity_exog(i,c,v,r,tt) }
                      } ;
 
 set valcap(i,v,r,t)            "i, v, r, and t combinations that are allowed for capacity",
@@ -1765,9 +1767,9 @@ m_required_prescriptions(i,v,r,t)$tmodel_new(t)
 
 m_required_prescriptions(i,v,r,t)$[tmodel_new(t)
                                    $(sum{tt$[yeart(t)>=yeart(tt)], prescribedrsc(i,v,r,tt) }
-                                     or m_capacity_exog(i,v,r,t) )$rsc_i(i)]
+                                     or sum{c, m_capacity_exog(i,c,v,r,t) } )$rsc_i(i)]
         = sum{(tt)$[(yeart(t) >= yeart(tt))], prescribedrsc(i,v,r,tt) }
-        + m_capacity_exog(i,v,r,t)
+        + sum{c, m_capacity_exog(i,c,v,r,t) }
 ;
 
 m_required_prescriptions_energy(i,v,r,t)$tmodel_new(t)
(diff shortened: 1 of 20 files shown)


reeds/core/setup/b_inputs.gms
```diff
@@ -1637,9 +1637,11 @@ avail_retire_exog_rsc(i,v,r,t)$[refurbtech(i)$(capacity_exog(i,v,r,t-1) > capaci
 
 avail_retire_exog_rsc(i,v,r,t)$[not initv(v)] = 0 ;
 
-m_capacity_exog(i,v,r,t)$capacity_exog(i,v,r,t) = capacity_exog(i,v,r,t) ;
+m_capacity_exog(i,c,v,r,t)$[i_c(i,c)$capacity_exog(i,v,r,t)] = capacity_exog(i,v,r,t) ;
 m_capacity_exog_energy(i,v,r,t)$capacity_exog_energy(i,v,r,t) = capacity_exog_energy(i,v,r,t) ;
-m_capacity_exog(i,"init-1",r,t)$geo(i) = geo_cap_exog(i,r) ;
+m_capacity_exog(i,c,"init-1",r,t)$[i_c(i,c)$geo(i)] = geo_cap_exog(i,r) ;
```
cavraam, line 1642: (feel free to ignore; general question not related to this PR) Why does init-1 of geothermal capacity require special treatment? Is there potentially an upstream action item to either move this upstream in input processing and/or reconcile with other technologies?

reeds/core/setup/b_inputs.gms
```diff
@@ -1752,22 +1754,22 @@ m_rscfeas(r,i,c,rscbin)$[i_c(i,c)$csp(i)$(not ban(i))$sum{ii$[(not ban(ii))$tg_r
...
-parameter m_required_prescriptions(i,v,r,t)        "--MW-- required power prescriptions by year (cumulative)" ;
+parameter m_required_prescriptions(i,c,v,r,t)      "--MW-- required power prescriptions by year (cumulative)" ;
 
 parameter m_required_prescriptions_energy(i,v,r,t) "--MWh-- required energy prescriptions by year (cumulative)" ;
 
 *following does not include wind
 *conditional here is due to no prescribed retirements for RSC tech
 *distpv is an rsc tech but is handled different via binned_capacity as explained above
-m_required_prescriptions(i,v,r,t)$tmodel_new(t)
+m_required_prescriptions(i,c,v,r,t)$[tmodel_new(t)$i_c(i,c)]
```
cavraam, line 1764: Double-checking: are we assigning previous years' 'prescribednonrsc' to all classes c, or does $i_c map to only a single class?

reeds/core/solve/2_temporal_params.gms
```diff
@@ -526,7 +526,7 @@ m_cf(i,c,v,r,h,t)$[(m_cf(i,c,v,r,h,t)<0.01)$valcap(i,v,r,t)] = 0 ;
 m_cf(i,c,v,r,h,t)$[cf_tech(i)$valcap(i,v,r,t)$m_cf(i,c,v,r,h,t)] = round(m_cf(i,c,v,r,h,t),3) ;
 
 * Remove capacity when there is no corresponding capacity factor
-m_capacity_exog(i,v,r,t)$[initv(v)$cf_tech(i)$(not sum{(h,c)$i_c(i,c), m_cf(i,c,v,r,h,t) })] = 0 ;
+m_capacity_exog(i,c,v,r,t)$[initv(v)$cf_tech(i)$(not sum{h, m_cf(i,c,v,r,h,t) })] = 0 ;
```
cavraam, line 529: m_cf carries $i_c, so we're good here

reeds/core/solve/3_solve_oneyear.gms
```diff
@@ -78,11 +78,11 @@ loop(i$rsc_i(i),
 * to avoid forcing recently upgraded capacity into retirement
 if(Sw_Upgrades = 1,
 
-    m_capacity_exog(i,v,r,t)$[valcap(i,v,r,t)$sameas(t,"%cur_year%")
+    m_capacity_exog(i,c,v,r,t)$[i_c(i,c)$valcap(i,v,r,t)$sameas(t,"%cur_year%")
```
cavraam, line 81: Might be an action item later on, but adding here just in case: Would it make sense to reconcile down the road `i_c(i,c)$valcap` with `valcap_class` and not have to think about whether a technology has a capacity factor ($valcap_class) or not (i_c(i,c)$valcap)?

reeds/core/terminus/report.gms
```diff
@@ -782,7 +782,10 @@ opres_trade(ortype,r,rr,t)$[opres_routes(r,rr,t)$tmodel_new(t)] =
 *=========================
 
 gen_new_uncurt(i,r,h,t)$[(vre(i) or storage_hybrid(i)$(not csp(i)))$valcap_irt(i,r,t)] =
-      sum{(v,c)$[i_c(i,c)$valinv(i,v,r,t)], (INV.l(i,v,r,t) + INV_REFURB.l(i,c,v,r,t)) * m_cf(i,c,v,r,h,t) * hours(h) }
+      sum{(v,c)$[i_c(i,c)$valinv(i,v,r,t)],
```
cavraam, line 785: great catch!

reeds/resource_adequacy/capacity_credit.py
```diff
@@ -118,7 +118,8 @@ def reeds_cc(t, tnext, casedir):
     #%% Load some inputs
     inputs_case = os.path.join(casedir, 'inputs_case')
     hierarchy = reeds.io.get_hierarchy(casedir).reset_index()
-    resources = pd.read_csv(os.path.join(inputs_case, 'resources.csv'))
+    resources = pd.read_csv(
+        os.path.join(inputs_case, 'resources.csv'), dtype={'c': str})
```
cavraam, line 122: Having to define the type here makes me think whether it would be easier to define classes as c0--c100 v/s 0--100? No strong feelings either way; it might save us some minor edits like this one

reeds/resource_adequacy/diagnostic_plots.py
```diff
@@ -811,7 +811,7 @@ def plot_pras_load_units(sw, dfs):
     ## Net demand
     vre_gen = dfs['vre_gen'].copy()
     vre_gen.columns = pd.MultiIndex.from_tuples(
-        vre_gen.columns.map(lambda x: tuple(x.split('|'))),
+        vre_gen.columns.map(lambda x: (x.split('|')[0], x.split('|')[-1])),
```
cavraam: Not an expert in python architecture, but I was (potentially naively?) thinking that we might want to avoid hard-coding [0] and [1] here (... and in reporting + a few other parts downstream). I iterated with AI to see if there is an obvious, more robust way to deal with this part, but we weren't able to find a better alternative...

reeds/core/setup/b_inputs.gms
cavraam, line 1: Starting line 1952 and through the end we omit $i_c from conditionals, which I believe is correct given that these lines deal with sets and parameters at the (i,v,r,t) level at this point

cavraam (approved):
The PR looks great and differences within 1e-4 in the objective function are reasonable in my mind. Minor comments attached as sanity checks.

reeds/core/setup/b_inputs.gms
wesleyjcole: Because there is no binning of exogenous geothermal capacity, all of it is placed into init-1.

reeds/core/setup/b_inputs.gms
wesleyjcole: Yes.  All prescribednonrsc are class-less technologies, so will just be mapped to class 0.

reeds/core/solve/3_solve_oneyear.gms
wesleyjcole: That's a great question.  Would it be better for `valcap_class` to be defined for everything, or only for technologies that have a class?  E.g., should `valcap_class` for `gas-cc` be no?  Given how similar the name is to `valcap`, it should probably be comprehensive to avoid confusion.  I'll go ahead and adjust that in a downstream change.

reeds/resource_adequacy/capacity_credit.py
wesleyjcole: There are several places where GAMS uses a class as a number, so I think keeping it as a number would be helpful.

reeds/resource_adequacy/diagnostic_plots.py
wesleyjcole: I looked into this some more, and the change can be reverted because PRAS collapses the classes.  Both versions work, but one less change is valuable.

reeds/core/setup/b_inputs.gms
wesleyjcole: Yes, that's right.

PR #222, opened 2026-09-14
Fix `cases_small.csv` and merge it with `cases_test.csv`

wesleyjcole:
## Summary
Per #218, ReEDS crashes when running with `cases_small.csv`, triggered by disabling offshore wind (GSw_OfsWind=0).  That happens because an offshore wind file is expected in two separate place, but is produced in neither when the switch is off.  This fix write empty files when the switch is off so that the model can complete successfully.

This PR also merges `cases_small.csv` into `cases_test.csv` so that the small test case will be monitored with the rest of our tests.

### Known incompatibilities
@muneezus said that `cases_small.csv` does not work with the GAMS community license, so we should probably look into that at some point.  My test case in the verification section had less than 5,000 rows and columns, but I did not test it against a community license.

I verified that `cases_small.csv` works after the fix--it did not work before, so there is no diff to provide.

`cases_small.csv` output report: [report.zip](https://github.com/user-attachments/files/32132828/report.zip)

A Pacific test run with default settings shows exactly no change:  [report.zip](https://github.com/user-attachments/files/32199309/report.zip).

### Details to double-check

Yes. Claude Code (Opus 5) was used to diagnose both failures from the `gamslog.txt` files of two failed `cases_small` runs, trace each to its root cause, and propose and apply the fixes. It also verified each fix by re-running the individual failing preprocessing step against the corresponding run directory and inspecting the resulting `inputs_case/` files.

### DIFF
--- a/reeds/input_processing/recf.py
+++ b/reeds/input_processing/recf.py
@@ -362,7 +362,11 @@ def main(reeds_path, inputs_case):
     lookup['i'] = lookup.ragg.map(lambda x: x.rsplit('|',1)[0])
 
     ### Offshore Wind
-    if int(sw['GSw_OfsWind']) != 0:
+    # If offshore wind is turned off, create an empty dataframe with the same
+    # index as onshore wind
+    if int(sw['GSw_OfsWind']) == 0:
+        df_windofs = pd.DataFrame(index=df_windons.index)
+    else:
         df_windofs = calculate_class_region_cf_hourly(
             inputs_case,
             'wind-ofs',
--- a/reeds/input_processing/writesupplycurves.py
+++ b/reeds/input_processing/writesupplycurves.py
@@ -293,11 +293,22 @@ def main(
     alloutcap_list.append(windcap)
 
     if write:
-        ## Exogenous wind capacity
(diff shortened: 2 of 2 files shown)


patrickbrown4:
Can we move the `cases_small.csv` case into `cases_test.csv` and then delete `cases_small.csv`? We run the full `cases_test.csv` for big PRs, so that would help us keep this case working.

wesleyjcole:
> Can we move the `cases_small.csv` case into `cases_test.csv` and then delete `cases_small.csv`? We run the full `cases_test.csv` for big PRs, so that would help us keep this case working.

Good idea.  Yes, I'll do that.

kennedy-mindermann (approved):
One suggestion: I think `ND_small` would fit better right after `USA_decarb` and before the `github_*` scenarios. That keeps it with the core cases and makes it easier for new users to find.

wesleyjcole:
> One suggestion: I think `ND_small` would fit better right after `USA_decarb` and before the `github_*` scenarios. That keeps it with the core cases and makes it easier for new users to find.

I put it at the end because it has so many switches that are not used by any other case, so it keeps all of those switches at the bottom of the file.  Any preference for which of these cosmetic outcomes is better (single-use switches at the bottom of the file or ND_small next to more of the core cases)?

PR #228, opened 2026-09-16
Collapse `wind-ons_1..10` to `wind-ons`

wesleyjcole:
## Summary
(Note: this PR is to be merged into wjc/dev/separate_class and not into main)

This PR (finally!) collapses the wind-ons_1*10 techs into wind-ons.  This is the first PR in #185 that shrinks `i`, and is 6d in the PR list [here](https://github.com/ReEDS-Model/ReEDS/issues/185#issuecomment-5427027139).

Not that this PR builds on `wjc/collapse_class`, which should be merged first.  For a comparison against that branch, see this [diff](https://github.com/ReEDS-Model/ReEDS/compare/wjc/collapse_class...wjc/collapse_class_windons).

### Implementation notes 
`wind-ons_1` through `wind-ons_10` are no longer valid techs in ReEDS, which required a collapse of the technology name in the relevant input files.

This PR also adds a new set `i_multiclass(i)` to capture technologies that map to more than one class.

I also added a `valinv_class(i,c,v,r,t)` because `valcap_class` is not tight enough for the class-level investment decisions.

Reported parameters indexed by `i` but not `c` (e.g. `gen_ivrt`, `gen_h`, `cap_firm`) no longer have reported values by class for the wind techs, because those have been collapsed.

### Additional changes
I added a filter to the output curtailment calculation of 1e-6 because there is floating point noise in there that bloats the `curt_h` output.

- `GSw_OnsWind6to10=0` is now a class ban (`bannew_class`) on what was the wind techs. Classes 6-10 can still be built in regions and vintages where they have prescribed capacity even when the switch is truned on.

### Issues resolved 
Partially addresses #185.

The USA_defaults case shows basiclaly no change in capacity:

image

And only a shifting of generation between zero marginal cost resources:

image

Model size is meaningfully reduced:

image

Runtime is likewise reduce by over an hour (16%):

image

Total system cost is different by 1e-6, and `z_rep` matches every year within 7e-7.

Reported system cost is slightly higher because before this change, initial wind capacity was all mapped to wind-ons_1 for cost data, with a fallback to the national average when that mapping didn't exist.  Since many regions didn't have wind-ons_1, the national average was used for most existing wind.  The new method now maps directly to each class, which leads to a higher cost for the existing wind plants and carries forward into the reported system cost.

image

Full compare report is here: [results-ClassCollapse,Wind-onsCollapse.pptx](https://github.com/user-attachments/files/32299683/results-ClassCollapse.Wind-onsCollapse.pptx).

### Details to double-check

Yes. I used Claude Code (Claude Opus 5) to sweep the code for class-suffixed names and one-class-per-tech assumptions, and then to implement the code changes made in this PR.  I also used it to help me compare the USA_defaults runs before and after the change.


inputs/degradation/degradation_annual_default.csv
```diff
@@ -1,6 +1,6 @@
 distpv,0.007
 upv_1*upv_10,0.007
-wind-ons_1*wind-ons_10,0.0027
+wind-ons,0.0027
```
cavraam, line 3: ** chef's kiss ** very satisfying to remove the duplicate data here and in the .csv's below

inputs/sets/i_c.csv
cavraam, line 1: Do we need to add this manually, or is it created from the underlying capacity data?

reeds/core/setup/b_inputs.gms
```diff
@@ -513,12 +513,10 @@ if(Sw_OfsWind = 0,
   ban(i)$i_subsets(i,'ofswind') = yes ;
 ) ;
 
+set bannew_class(i,c) "resource classes banned from creating new capacity, except where the class has prescribed capacity" ;
```
cavraam, line 516: Curious why we do this? Also, would it make sense to make this a little more flexible, i.e., user-defined?

reeds/core/setup/b_inputs.gms
```diff
@@ -4957,6 +4953,26 @@ valinv(i,v,r,t)$[not valcap(i,v,r,t)] = no ;
 
 * Valid capacity by resource class
 valcap_class(i,c,v,r,t)$[i_c(i,c)$valcap(i,v,r,t)] = yes ;
+* For technologies with several classes, a class needs existing capacity (init vintages),
```
cavraam, line 4956: no strong feelings here; wondering if it makes sense to define i_multiclass closer to here (where it's being used) v/s all the way up (where it can get calculated). No preference either way.

reeds/core/setup/c_model.gms
```diff
@@ -699,8 +699,8 @@ eq_cap_new_noret(i,c,v,r,t)$[valcap_class(i,c,v,r,t)$tmodel(t)$newv(v)$(not upgr
     sum{tt$[inv_cond(i,v,r,t,tt)$(tmodel(tt) or tfix(tt))$valcap(i,v,r,tt)],
               degrade(i,tt,t) * (INV(i,v,r,tt)$(not cf_tech(i))
                                  + sum{rscbin$m_rscfeas(r,i,c,rscbin),
-                                       INV_RSC(i,c,v,r,rscbin,tt) }$cf_tech(i)
-                                 + INV_REFURB(i,c,v,r,tt)$[refurbtech(i)$Sw_Refurb])
+                                       INV_RSC(i,c,v,r,rscbin,tt) }$[cf_tech(i)$valinv_class(i,c,v,r,tt)]
```
cavraam, line 702: good catch with valinv_class(i,c,v,r,t)

reeds/core/setup/c_model.gms
```diff
@@ -2838,7 +2838,7 @@ eq_RPS_OFSWind(st,t)$[tmodel(t)$stfeas(st)$offshore_cap_req(st,t)$Sw_StateRPS
 
 * investments over time
     + sum{(i,v,r,tt)$[r_st(r,st)$ofswind(i)$inv_cond(i,v,r,t,tt)$(tmodel(tt) or tfix(tt))],
-          INV(i,v,r,tt) + sum{c$i_c(i,c), INV_REFURB(i,c,v,r,tt) }$[refurbtech(i)$Sw_Refurb] }
+          INV(i,v,r,tt) + sum{c$valinv_class(i,c,v,r,tt), INV_REFURB(i,c,v,r,tt) }$[refurbtech(i)$Sw_Refurb] }
```
cavraam, line 2841: I really like how neat these look with valinv_class

reeds/core/terminus/report.gms
```diff
@@ -797,6 +797,7 @@ curt_h(r,h,t)$tmodel_new(t) =
     - sum{(ortype,i,v)$[Sw_OpRes$opres_h(h)$reserve_frac(i,ortype)$valgen(i,v,r,t)$vre(i)],
           OPRES.l(ortype,i,v,r,h,t) }
 ;
+curt_h(r,h,t)$[abs(curt_h(r,h,t)) < 1e-6] = 0 ;
```
cavraam, line 801: might want to add a comment above to highlight that this is happening? I can't imagine this making a difference in any analysis, but good to have one more flag in the model

reeds/input_processing/recf.py
```diff
@@ -295,7 +295,15 @@ def check_missing_class_resource(existing_techs, resources):
         raise ValueError(err)
     else:
         print('All capacities and resources are matched.')
-            
+
+
+def label_class_region(tech, columns, collapsed):
```
cavraam, line 300: not super savvy in the way pre-processing scripts are organized, but this function seems like a good fit for io.py or reeds/tech.py I think...

cavraam (approved):
This PR is **very** exciting! 
- The change in the objective function is negligible and explainable by the wind-ons cost assignment.
- It is wonderful to see runtimes go down. I would be curious to test the runtime reduction for a wider range of scenarios (e.g., zero-carbon, HOG/LOG) when we're closer to merging into main.
- No major concerns; minor comments attached.

reeds/core/setup/b_inputs.gms
wesleyjcole: It's used for making model instances that are small enough to work with the GAMS community license.  Patrick [recommended ](https://github.com/ReEDS-Model/ReEDS/issues/185#issuecomment-5427434775) we make this user-defined, and my plan is to do that after we finish collapsing the classes as I think that will make it easier.

inputs/sets/i_c.csv
wesleyjcole: We create this ourselves and then use it to map capacity data.

reeds/core/setup/b_inputs.gms
wesleyjcole: I prefer defining it up higher so that if it is used in earlier code it won't need to be moved up.

reeds/core/terminus/report.gms
wesleyjcole: Good call.  Adding a comment.

reeds/input_processing/recf.py
wesleyjcole: Nothing else is using this currently, though I think this is part of what will need evaluated when things are done.  If we are able to collapse all techs with classes many of these will go away and then it won't matter.

PR #229, opened 2026-09-17
Add new CT fips to county2zone.csv

grantbuster:
## Technical details 
It seems like counties in CT have updated their FIPS codes. These new rows in the county mapping table will allow you to join to the counties shape file from the census bureau. 

More details: 
https://developer.ap.org/ap-elections-api/docs/CT_FIPS_Codes_forPlanningRegions.htm

I verified that these FIPS codes do not appear elsewhere in the mapping table. 

### Details to double-check

- [x] Charge code provided to reviewers: GMLC 12765 07.01.03

No. 

@louisaserpe edits to county mappings, you may be poc?


louisaserpe (commented):
I believe these are the legacy FIPS codes, are you wanting to include them here in addition to the new 9 planning region FIPS codes? 

Can you point me to the shapefile that the old FIPS code are compatible with?

patrickbrown4 (changes requested):
I'm not sure which are current, but I forgot the FIPS codes over time! Some notes for when they change (more specifically, when we choose to update to reflect a change):
- [ ] Every `inpus/zones/{GSw_ZoneSet}/county2zone.csv` file needs to be updated, not just `inputs/zones/z90`
- [ ] The relevant entries in all the `inpus/zones/{GSw_ZoneSet}/zonehash.csv` files also need to be updated (the zonehashes are explained in https://github.nrel.gov/ReEDS/ReEDS-2.0/pull/2003)
- [ ] The relevant entries in `inputs/transmission/itl_NARIS.csv` (existing transmission capabilities) need to be updated. It's probably easiest just to rerun the affected interfaces for all the supported `GSw_ZoneSet`s and add them to the table rather than trying to update in place.

jvcarag:
Apologies for the confusion here: I had accidentally shared the 2018 US county-level shapefile with @grantbuster by linking to [this line in reeds/spatial.py](https://github.com/ReEDS-Model/ReEDS/blob/d3cc885ed47b06334826b330b55ae744ee160496/reeds/spatial.py#L61), instead of the line directly below it that points to the 2022 shapefile with the updated CT FIPS codes. 

Closing this PR since the updated CT FIPS codes are already defined in the model, and the ones added by this PR are for the legacy CT regions.

